PR #5, opened 2015-01-27
Initial draft of a node binding via swig 3.x

dnwe:
As an alternative to the emscripten-based javascript bindings, swig 3.x now has native support for generating node wrappers, so it makes sense to offer them as a further alternative entry point (alongside the emscripten one) that is closer to the existing swig bindings for the other languages.


dnwe:
Pushed as d564ddcfbfa3901cafe823ead283d091db1157b7

PR #6, opened 2015-02-03
PROTON-814: only return error rc on a failed poll

dnwe:
In posix pn_selector_select, if a poll() call returns result == -1 and
sets selector->error from the errno, any subsequent pn_selector_select
calls on the same selector will return have a return code of the
previously cached errno even if they succeed.


dnwe:
@rhs please can you take a look at this pull request and confirm that you're happy with the change?

PR #8, opened 2015-02-16
PROTON-811: safely handle TransportException

dnwe:
Previously TransportException could be thrown in the TransportImpl
beforePosition initializer that was using tail() directly without a
separate catch to ignore it. It seems cleaner to add a position()
accessor to the TransportInput interface that handles closed streams
in the same way as capacity(), without throwing the RuntimeException


dnwe:
@rhs this is a fix to the patch you previously merged in from @prestona for PROTON-811

After a ``@close`` was sent from the remote end, we could end up throwing a TransportException before the input was processed and the close would never be ACK'ed.

rhs:
Looks good to me.

dnwe:
@rhs thanks

PR #11, opened 2015-03-10
NO-JIRA: Measure size of encoded data.

alanconway:
Introduce pn_data_encode2 which allows you to both encode data and, if you don't have enough
space, find out how much space you need. You can also use this to simply find out how
much space is required to  the data.

This supports two patterns of use:

1. Check how much space you need, allocate the space and encode.

2. Try to encode into a buffer you already have. If that doesn't work, re-allocate
   to the correct size and encode.

Eliminates the need for guessing and buffer doubling strategies, which are
inefficient given that we know exactly what is in the data.

The original pn_data_encode is unmodified for compatibility.

### DIFF
--- a/proton-c/include/proton/codec.h
+++ b/proton-c/include/proton/codec.h
@@ -512,6 +512,21 @@ PN_EXTERN int pn_data_format(pn_data_t *data, char *bytes, size_t *size);
 PN_EXTERN ssize_t pn_data_encode(pn_data_t *data, char *bytes, size_t size);
 
 /**
+ * Writes the contents of a data object to the given buffer as an AMQP data
+ * stream. If the entire contents cannot be written, returns the number of bytes
+ * that would be needed to write it all.
+ *
+ * @param data the data object to encode
+ * @param bytes the buffer for encoded data (0 if you just want to calculate the size needed)
+ * @param size the size of the buffer (0 if you just want to calculate the size needed)
+ *
+ * @param size_t returns the size needed to encode all the data. If the returned
+ *     size <= size parameter then all the data was encoded successfully. If not,
+ *     a buffer of the returned size is required to encode it all.
+ */
+PN_EXTERN size_t pn_data_encode2(pn_data_t *data, char *bytes, size_t size);
+
+/**
  * Decodes a single value from the contents of the AMQP data stream
  * into the current data object. Note that if the pn_data_t object is
  * pointing to a current node, the decoded value will overwrite the
--- a/proton-c/src/codec/codec.c
+++ b/proton-c/src/codec/codec.c
@@ -1424,6 +1424,11 @@ ssize_t pn_data_encode(pn_data_t *data, char *bytes, size_t size)
   return pn_encoder_encode(data->encoder, data, bytes, size);
 }
 
+size_t pn_data_encode2(pn_data_t *data, char *bytes, size_t size)
+{
+  return pn_encoder_encode2(data->encoder, data, bytes, size);
+}
+
 ssize_t pn_data_decode(pn_data_t *data, const char *bytes, size_t size)
 {
   return pn_decoder_decode(data->decoder, bytes, size, data);
--- a/proton-c/src/codec/encoder.c
+++ b/proton-c/src/codec/encoder.c
@@ -30,6 +30,7 @@
 #include "data.h"
 
 struct pn_encoder_t {
+  bool counting;   /* Count the total size of the data even if there is not enough buffer */
   char *output;
   size_t size;
   char *position;
@@ -39,6 +40,7 @@ struct pn_encoder_t {
 static void pn_encoder_initialize(void *obj)
 {
   pn_encoder_t *encoder = (pn_encoder_t *) obj;
+  encoder->counting = false;
   encoder->output = NULL;
   encoder->size = 0;
   encoder->position = NULL;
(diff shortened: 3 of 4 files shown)


astitcher:
Thinking about it a bit more, I think I particularly object to the new public API.
What you really want at an API level here is simply a function to tell you how many bytes you need to allocate, not a strange combined encode, but if you can't tell how many bytes I would need.

FWIW, I think that there are some simple strategies that can get you the needed size out of a pn_data_t with a little extra being held in that structure, withotu having to do the full encode.

alanconway:
On Wed, 2015-03-11 at 08:00 -0700, Andrew Stitcher wrote:
> Thinking about it a bit more, I think I particularly object to the new
> public API.
> What you really want at an API level here is simply a function to tell
> you how many bytes you need to allocate, not a strange combined
> encode, but if you can't tell how many bytes I would need.

A simpler pn_data_encoded_size() API would be easy to do but potentially
requires scanning the data structure twice to encode...

> 
> FWIW, I think that there are some simple strategies that can get you
> the needed size out of a pn_data_t with a little extra being held in
> that structure, withotu having to do the full encode.

I'd love to hear some details. Are you suggesting we accumulate the
encoded size every time you modify the data?

The motivation for all this is to avoid the "guess and double"
allocation dance that the bindings currently do, e.g. in python:

    size = 1024
    while True:
      cd, enc = pn_data_encode(self._data, size)
      if cd == PN_OVERFLOW:
        size *= 2

alanconway:
On Wed, 2015-03-11 at 14:26 -0400, Alan Conway wrote:
> On Wed, 2015-03-11 at 08:00 -0700, Andrew Stitcher wrote:
> > Thinking about it a bit more, I think I particularly object to the new
> > public API.
> > What you really want at an API level here is simply a function to tell
> > you how many bytes you need to allocate, not a strange combined
> > encode, but if you can't tell how many bytes I would need.

Not that strange BTW, this is what snprintf does.

astitcher:
> I'd love to hear some details. Are you suggesting we accumulate the encoded size every time you
> modify the data?

Not especially, but possibly - Iobserve that when you need to resize your buffer you always end up doing at least 2 passes of the data. So in this case it isn't really a big deal to ask for the size after the first attempt failed. That leads me to caching the required size in the pn_data_t whenever we do encode even if (especially if) it fails.

So the flow would look like (in python)

    size = 1024
    cd, enc = pn_data_encode(self._data, size)
    if cd == PN_OVERFLOW:
      size = pn_data_encoded_size(self._data)
      cd, enc = pn_data_encode(self._data, size)
      assert(cd!=PN_OVERFLOW)

So no dance any more (at least only the two step, not round -and-round). And if this is not fast enough we can cache away the size during pn_data_encode() for use in pn_data_encoded_size(), remembering to invalidate it if the struct is further changed (not actually likely).

Just in case it isn't clear this is better than calculating the size first because there is no wasted work assuming a cached size (at least compared to the existing and your proposed scheme), whereas always asking for the size and calculating it is always extra work if 80% of the time the buffer would have been big enough anyway.

This isn't so far from your proposal, I just think the API doesn't suck ;-)

alanconway:
On Thu, 2015-03-12 at 00:52 -0700, Andrew Stitcher wrote:
>         I'd love to hear some details. Are you suggesting we
>         accumulate the encoded size every time you
>         modify the data?
>         
> 
> Not especially, but possibly - Iobserve that when you need to resize
> your buffer you always end up doing at least 2 passes of the data. So
> in this case it isn't really a big deal to ask for the size after the
> first attempt failed. That leads me to caching the required size in
> the pn_data_t whenever we do encode even if (especially if) it fails.
> 
> So the flow would look like (in python)
> 
> size = 1024
> cd, enc = pn_data_encode(self._data, size)
> if cd == PN_OVERFLOW:
>   size = pn_data_encoded_size(self._data)
>   cd, enc = pn_data_encode(self._data, size)
>   assert(cd!=PN_OVERFLOW)
> 
> So no dance any more (at least only the two step, not round
> -and-round). And if this is not fast enough we can cache away the size
> during pn_data_encode() for use in pn_data_encoded_size(), remembering
> to invalidate it if the struct is further changed (not actually
> likely).
> 
> Just in case it isn't clear this is better than calculating the size
> first because there is no wasted work assuming a cached size (at least
> compared to the existing and your proposed scheme), whereas always
> asking for the size and calculating it is always extra work if 80% of
> the time the buffer would have been big enough anyway.
> 
> This isn't so far from your proposal, I just think the API doesn't
> suck ;-)

Agreed, same behavior nicer API. Will rework & re-post. Thanks!

alanconway:
Comitted the suggested API upstream. Doesn't do the suggested caching yet.

commit fac7c86c8bc818ea845d6426fd85095a189522d6
Author: Alan Conway <aconway@redhat.com>
Commit: Alan Conway <aconway@redhat.com>

    NO-JIRA: Measure size of encoded data.
    
    Introduce pn_data_size() to calculate the buffer space required to encode a data object.
    
    - encoder.c measures full size of data.
    - interop.py verifies computed size matches actual decode size.
    - update java/jython compatilbility layer.
    
    Code review at https://github.com/apache/qpid-proton/pull/11.
    
    Did not (yet) implement size caching as suggested in the review as invalidating
    the cache is a little involved. Leaving it for another day.

astitcher:
I think you new pn_data_size() possibly has the wrong side effects - it seems to lose the pn_data cursor position, so that merely querying the size will be hard to intermix with appending to the pn_data. It is true that this is not something the current code ever does does, but I think making the query function have side effects (except caching the size itself perhaps) is unexpected.

alanconway:
On Mon, 2015-03-16 at 12:31 -0700, Andrew Stitcher wrote:
> I think you new pn_data_size() possibly has the wrong side effects -
> it seems to lose the pn_data cursor position, so that merely querying
> the size will be hard to intermix with appending to the pn_data. It is
> true that this is not something the current code ever does does, but I
> think making the query function have side effects (except caching the
> size itself perhaps) is unexpected.
> 
> —
> Reply to this email directly or view it on GitHub.
> 
Fixed

commit 973bad033ba3a1b700ab80ab4edee209ab81f05a
Author: Alan Conway <aconway@redhat.com>
Commit: Alan Conway <aconway@redhat.com>

    NO-JIRA: Restore data position when measuring size of encoded data.

PR #13, opened 2015-03-16
PROTON-834: further UTF-8 encoder fixes

dnwe:
After commit c65e897 it turned out there were still some issues with
strings containing a codepoint >0xDBFF which was being incorrectly
treated as a surrogate pair in the calculateUTF8Length method.

Fixed this up and added some more test coverage.


gemmellr:
Looks good to me. We should request for inclusion in 0.9 if there is an RC3 to pick up aconways SSL change.

PR #14, opened 2015-03-30
NO-JIRA: export.sh creates pax-based tar

dnwe:
Having switched to using git-archive, we end up creating a tar that uses
the PAX-based scheme for supporting long pathnames. This causes issues
for a number of extraction tools (ant, 7-zip, etc.) which only support
the GNU tar format. We should stick with GNU tar, as used for previous
qpid-proton releases.

Also update the script to revert to using a suffix of .tar.gz (rather
than .tgz) for the source distribution.

### DIFF
--- a/bin/export.sh
+++ b/bin/export.sh
@@ -20,6 +20,8 @@
 #
 
 # export.sh - Create a release archive.
+set -e
+trap "cleanup" 0 1 2 3 9 11 13 15
 
 ME=$(basename ${0})
 SRC=$(dirname $(dirname $(readlink -f $0)))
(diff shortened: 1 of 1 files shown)


rhs:
+1

dnwe:
@rhs as we've not announced 0.9 on the website yet, it might be worth re-creating the 0.9 release tar using the updated export.sh and re-uploading it?

gemmellr:
As the archive has already been distributed via the mirror system, I think updating the current named files would likely be considered bad form at this point. I'm not sure how archive.apache.org (which saves the contents even after they are deleted from the mirror system) would react to that either. A quick point release instead perhaps?

dnwe:
*technically* we wouldn't have a name clash because of the .tgz --> .tar.gz renaming, so people would only notice if they browsed the mirror filesystems and saw both a .tgz and a .tar.gz

But, you're right, it might be worth pushing it out as qpid-proton-0.9.1.tar.gz

gemmellr:
Ah, I actually missed the note about changing the filename entirely :)

Given that it wouldnt clash as I thought and would actually restore the filename to its historical convention it might actually be helpful to post it for 0.9, though it could be confusing to some people seeing both in the same dir.

rhs:
FYI I've added the updated tarballs for 0.9

PR #17, opened 2015-04-09
PROTON-334: SASL Implementation for Proton-C using Cyrus SASL

astitcher:
This work Adds some new APIs to the transport and connection
objects to make a higher level abstraction for authentication.
This generally makes it much easier to use authentication.

It also vastly changes the Proton C API for SASL and deprecates
nearly all of the previous interface that allowed reading and
writing individual SASL frames.

### DIFF
--- a/proton-c/CMakeLists.txt
+++ b/proton-c/CMakeLists.txt
@@ -44,8 +44,16 @@ endif()
 
 # Set the default SSL/TLS implementation
 find_package(OpenSSL)
-
 find_package (PythonInterp REQUIRED)
+find_package(SWIG)
+# FindSwig.cmake "forgets" make its outputs advanced like a good citizen
+mark_as_advanced(SWIG_DIR SWIG_EXECUTABLE SWIG_VERSION)
+
+# See if Cyrus SASL is available
+find_library(CYRUS_SASL_LIBRARY sasl2)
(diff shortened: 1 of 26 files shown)


astitcher:
See the wiki for more information and context:
https://cwiki.apache.org/confluence/x/B5cWAw

dnwe:
@astitcher minor issue, unrelated to the feature itself, but the CMake build doesn't currently report which sasl implementation is being used. I just see:
```
-- Looking for sasl_checkpass in sasl2
-- Looking for sasl_checkpass in sasl2 - found
-- Looking for include file sasl/sasl.h
-- Looking for include file sasl/sasl.h - found
-- Found SASL: 1  
```
Could you perhaps add something along the lines of:
```
if (FOUND_SASL_LIB)
  message(STATUS "Found SASL: ${FOUND_SASL_LIB}")
else()
  message(STATUS "Found SASL: BUILT-IN")
endif(FOUND_SASL_LIB)
```

astitcher:
@dnwe I think I copied that CMake code directly from qpid. I'll look at doing something like you suggested - the CMake output does look clunky.

rhs:
Looks good to me, assuming all the tests pass and what not, I'm +1.

PR #19, opened 2015-04-14
Mbed minor changes and sasl mech

dcristoloveanu:
This are 2 minor changes in order to be able to:
- Compile on an Arm CC compiler and more specifically compile in mbed.org's environment. These are basically some minor header include changes.
- Addition of an API to inject which client SASL mechanism to use (needed in order to be able to talk to Azure Event Hubs while using CBS).

Any feedback is most appreciated.

Thanks,
/Dan


rhs:
@astitcher: do you want to wait until PROTON-334 lands to pull this in given the conflict you mention?

dcristoloveanu:
I had a look at PROTON-334 and I'll remove the SASL API from my pull request, since it seems PROTON-334 should allow the case we want and like you said the API I would be adding here would become obsolete.

dcristoloveanu:
I’ve removed the SASL client mechanism setter as I think PROTON-334 should do when it comes online. And if it does not we can always bring this back in another pull request. One small thing at a time.

Thanks for the feedback.

/Dan

astitcher:
@dcristoloveanu I've landed PROTON-334 now, so you should check that trunk now does what you want.

dcristoloveanu:
I'll take a look at PROTON-334 and see if we need anything extra. If so, I'll create another pull request.
Could we merge in this pull request, as I removed the new SASL API from it?

Thanks,
/Dan

dcristoloveanu:
I think this can be merged in now, unless there's anything else that you think needs addressing.

Thanks!
/Dan

PR #20, opened 2015-04-20
PROTON-848-and-849: stop leaking the transport state

gemmellr:
Stops leaking the transport state by removing the maps storing the TransportSession or TransportLink objects, instead using the references set on the associated Session and Link objects at the time they would have been entered into the maps. Not much to that bit.

The bulk of the annoyance came from some tests failing after those changes due to alteration of the events being emitted when calling free+unbind on the connection+transport. Much debugging suggested this was possibly as result of existing differences in behaviour (mainly that things were not showing FINAL events until the unbind step in proton-j, unlike proton-c) which just happened to become more visible following my modifications. After looking at some of the differences between proton-c and proton-j's session+link reference handling I made some alterations to have it act more like proton-c, which fixed the earlier test failures, however I'd like a second opinion from someone who actually understands what this stuff is meant to do :)

### DIFF
--- a/proton-j/src/main/java/org/apache/qpid/proton/engine/impl/TransportImpl.java
+++ b/proton-j/src/main/java/org/apache/qpid/proton/engine/impl/TransportImpl.java
@@ -96,10 +96,6 @@ public class TransportImpl extends EndpointImpl
     private TransportInput _inputProcessor;
     private TransportOutput _outputProcessor;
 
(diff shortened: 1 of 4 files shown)


gemmellr:
@rhs can you take a look please? Mainly at the channel/handle process updates, the actual leak prevention wasnt that interesting in the end.

rhs:
Looks good to me.

PR #21, opened 2015-04-20
PROTON-853: stop erroneous attach being sent

gemmellr:
The changes for PROTON 154 commit 7d3063e7c488c97b9bad61e862d54b2b11dbc3d5 in 0.9 lead to situations where the transport will decide to send a new Attach frame out for a link that has been closed/detached.

The original change aimed to ensure that a link with the same name as one which had just been closed could be attached on the same session, by clearing the 'attach sent' and 'detach received' flags of the TransportLink. The reason that seemed necessary is that the old link object (and in turn TransportLink object) would be returned due to some caching within SessionImpl, unless you ensure to first call free() on the old link. Gordon noticed the same effect from the server side in proton-c via PROTON 850.

These changes stop the erroneous attach being sent by first reverting the change from PROTON 154, and looking to stop the earlier situation arising by returning a new Link object rather than a cached one if the states are closed (i.e similar to Gordons from PROTON 850), meaning a new TransportLink gets created rather than the old one being reused and there then being no need to play with the boolean states. It keeps a list of any 'old links' that would have been forogtten to ensure they get free'd when the session is.

The earlier test was ineffective and passed with or without the earlier changes. This was because it actually created new connections and sessions when it was aiming to be reuse the same session. I updated the test to do that, and it now catches both PROTON 154 and PROTON 850 (in proton-c too).

### DIFF
--- a/proton-j/src/main/java/org/apache/qpid/proton/engine/impl/SessionImpl.java
+++ b/proton-j/src/main/java/org/apache/qpid/proton/engine/impl/SessionImpl.java
@@ -21,6 +21,7 @@
 package org.apache.qpid.proton.engine.impl;
 
 import java.util.*;
+
 import org.apache.qpid.proton.engine.EndpointState;
 import org.apache.qpid.proton.engine.ProtonJSession;
 import org.apache.qpid.proton.engine.Session;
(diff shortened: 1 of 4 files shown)


gemmellr:
@rhs can you take a look? @dnwe can you also take a look to confirm if the changes work for you, e.g. handles any tests you have elsewhere based on needing the earlier changes?

dnwe:
@gemmellr sure, just pulling in this PR now and will run some scenarios through it

gemmellr:
Just a note to say, the new test will currently fail against proton-c just now, but Gordons fix from PROTON-850 resolves that. I mentioned this to him and he will push it in at some point.

rhs:
+1 from me

Initially I thought this wouldn't account for detach properly, but looking at the code I managed to convince myself that it will. If you have time, however, it would be nice to augment/add a test of the detach scenario.

gemmellr:
Great. I'll add a test for the detach case as well in the morning.

Could you also look at PR 20 so I dont have to rebase my main repo to commit these? :P

dnwe:
+1 from me, not seen any issues in our FV after applying these commits + PROTON-850

PR #23, opened 2015-04-27
Fixes for Mac OS X

dnwe:
1. default to building universal binaries so the various language bindings can link against the system libraries out of the box

2. small fix to SelectableMessengerTest as it was regularly failing on Mac builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,13 @@ if (MSVC)
 endif (MSVC)
 option(BUILD_WITH_CXX "Compile Proton using C++" ${DEFAULT_BUILD_WITH_CXX})
 
+if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+  # Default to universal binary on Mac OS X unless user has overriden
+  if (NOT DEFINED CMAKE_OSX_ARCHITECTURES OR "${CMAKE_OSX_ARCHITECTURES}" STREQUAL "")
+    set(CMAKE_OSX_ARCHITECTURES "i386;x86_64")
+  endif ()
+endif ()
+
 if (BUILD_WITH_CXX)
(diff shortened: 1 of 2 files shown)


dnwe:
@rhs I believe you originally wrote the SelectableMessengerTest, could you take a look at change and confirm if this is an acceptable change in the test behaviour? It wasn't clear to me if you were intending to validate that a single call to Pump#pump would always result in at least one message being flowed into incoming?

dnwe:
fixes-for-mac: build passing

rhs:
I'm a little puzzled as to why the test change would be necessary. Pump has been written to keep looping until there is no I/O left to perform for any of its messengers. Given that the receiving messenger has issued credit, and the sending messenger has posted messages for transfer, I think this is probably somehow indicative of a real bug somewhere (possibly in some of the test framework code). How easy is it to reproduce this on a mac?

dnwe:
@rhs readily reproducible, [output from 20 runs](https://gist.github.com/dnwe/76c504f2a3067b32b904)
and [python trace bz2](https://dl.dropboxusercontent.com/u/951543/trace.log.bz2) from a single run of the four tests

rhs:
Hmm, I'm not sure how useful the trace is since it is once where all the tests passed. Any chance you could grab one where the tests fail?

Thinking a bit more, I think the test makes the assumption that the underlying loopback interface is synchronous, i.e. as soon as a write succeeds, any corresponding reads will be ready. I've added info to the assert that will hopefully test this theory. Can you do a git pull and reproduce the failure again and post the traceback?

rhs:
For the record, I tried making the test assertion more specific, but it is difficult to come up with an assertion that accommodates the asynchronous behavior and isn't extremely brittle and specific. Given that I'm +1 on the change.

PR #24, opened 2015-04-28
Fix none_sasl.c to use pn_strdup instead of strdup

dcristoloveanu:
(PROTON-870) none_sasl.c should use pn_strdup.

none_sasl.c uses strdup, but strdup is not part of C99 and subsequently does not exist on some platforms. Instead pn_strdup should be used.

Thanks,
/Dan


astitcher:
Oops

PR #25, opened 2015-04-29
Initial transport frame size 256 bytes

dcristoloveanu:
In order to reduce RAM footprint on rather simple scenarios on small devices the initial transport frame size could be lowered to 256, since the buffer is anyhow automatically enlarged if needed.

Per earlier suggestions I created a config.h where parameters like this can go (the next candidate is the initial size for pn_data). If the need arises these parameters should be easily changeable in this case to match the platform need.

### DIFF
--- a/proton-c/CMakeLists.txt
+++ b/proton-c/CMakeLists.txt
@@ -342,6 +342,8 @@ set (qpid-proton-core
   src/messenger/transform.c
   src/selectable.c
 
+  src/config.h
+
   ${CMAKE_CURRENT_BINARY_DIR}/src/encodings.h
   ${CMAKE_CURRENT_BINARY_DIR}/src/protocol.h
   )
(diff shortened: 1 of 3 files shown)


dcristoloveanu:
Friendly ping ...

Thanks,
/Dan

dnwe:
@dcristoloveanu couple of buddycheck comments

1. the include guard `CONFIG_H` is a bit generic, perhaps change to `_PROTON_SRC_CONFIG_H`
2. would 512 be a better default as it would match [MIX-MAX-FRAME-SIZE](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#definition-MIN-MAX-FRAME-SIZE)

dcristoloveanu:
@dnwe

1.Sure, updated 
2.Changed to 512.

Thanks!

dnwe:
@dcristoloveanu you don't appear to have pushed the updated commits for this PR - could you also `git pull --rebase` from master and then squash into a single commit with the updates

dcristoloveanu:
@dnwe Sorry about that, I updated the branch to have only on commit.

Thanks,
/Dan

PR #26, opened 2015-04-29
proton-c: minor code fixes

dnwe:
Prevent pn_data_vfill from peeking at memory outside the intended bounds and fix a few minor -Wextra warnings about incomplete struct initialisation


dnwe:
build passing

https://travis-ci.org/dnwe/qpid-proton/builds/60601702

rhs:
Looks good to me. On the struct initialization stuff, it might be worth seeing if we can make gcc turn those into errors if it causes problems on other compilers.

PR #27, opened 2015-04-30
Add some NULL checks and fix some realloc leaks

dcristoloveanu:
Along the lines of PROTON-727, added NULL checks for some malloc/realloc calls that were either ignoring the result or leaking in the case of realloc.
 
Thanks,
/Dan

### DIFF
--- a/proton-c/src/buffer.c
+++ b/proton-c/src/buffer.c
@@ -40,10 +40,21 @@ struct pn_buffer_t {
 pn_buffer_t *pn_buffer(size_t capacity)
 {
   pn_buffer_t *buf = (pn_buffer_t *) malloc(sizeof(pn_buffer_t));
-  buf->capacity = capacity;
-  buf->start = 0;
-  buf->size = 0;
-  buf->bytes = capacity ? (char *) malloc(capacity) : NULL;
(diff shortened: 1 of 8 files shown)


dcristoloveanu:
Friendly ping ...

Thanks,
/Dan

dnwe:
@dcristoloveanu I'd be happy to merge these changes in for you as it is certainly better for the library not to segfault if a malloc fails.

However, I do wonder if there are numerous places where these methods are currently called and expected to always be non-NULL. For example, `pn_buffer` allocations are used liberally and rarely checked for validity before being used, so unless I'm mistaken, it seems like we will just push the segfault up the call chain?

dcristoloveanu:
I couldn't agree more. Actually all calls should have their result checked. After all malloc/realloc are not more important than pb_buffer and all the constructors or other APIs.

Unfortunately doing all of it in one sweep is not really sustainable, so I thought it would be a good idea to start somewhere. I'd love to send more PRs like this to finish dealing with malloc/realloc and then move higher to ensuring that each unit checks returns of the functions it calls.

dnwe:
OK I'm happy. Merged :+1:

PR #28, opened 2015-05-06
NO-JIRA: improvements to export.sh

dnwe:
+ allow specific tag to be specified as cmdline arg
+ confirm the tag exists before proceeding
+ use the version.txt from that tag as the tar prefix folder
+ re-run the version.sh script before packaging to ensure the correct
  version is set
+ add .gitattributes to exclude .gitignore files from the exported tar

Closes #28

### DIFF
--- a/bin/export.sh
+++ b/bin/export.sh
@@ -28,7 +28,8 @@ SRC=$(dirname $(dirname $(readlink -f $0)))
 
 usage()
 {
-    echo "Usage: ${ME} [DIR]"
+    echo
+    echo "Usage: ${ME} [DIR] [TAG]"
     exit 1
 }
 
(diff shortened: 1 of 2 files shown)


gemmellr:
A little more info:most of the changes are to ensure the dir within the tar matches the version, in case the tag name does not (e.g a proposed final RC being voted on).

Looks good to me.

bin/export.sh
```diff
@@ -54,15 +65,18 @@ WORKDIR=$(mktemp -d)
...
-    TAG=$(git describe --tags --always)
     MTIME=$(date -d @`git log -1 --pretty=format:%ct tags/${TAG}` '+%Y-%m-%d %H:%M:%S')
     ARCHIVE=$DIR/qpid-proton-${TAG}.tar.gz
+    VERSION=$(git show tags/${TAG}:version.txt)
+    PREFIX=qpid-proton-${VERSION}
     [ -d ${WORKDIR} ] || mkdir -p ${WORKDIR}
-    git archive --format=tar --prefix=qpid-proton-${TAG}/ tags/${TAG} \
+    git archive --format=tar --prefix=${PREFIX}/ tags/${TAG} \
         | tar -x -C ${WORKDIR}
+    ${SRC}/bin/version.sh ${WORKDIR}/${PREFIX} ${VERSION}
```
rhs: This doesn't make sense to me. This script is supposed to simply export exactly the contents of a tag in the git repo. This tag is generally a signed release, and running scripts over the content of the release will invalidate that process entirely.

rhs:
Fixing the prefix in the tarball makes sense to me, however some of the other changes do not. The release process is supposed to proceed from a signed tag, which means that if the version for that signed tag is wrong, then something prior in the process has gone wrong.

In general having export.sh modify the contents of the exported source tree in any way is dangerous since it will lead to divergence between the contents of the git tag for a given release and the contents of the exported tarballs for that release, and that is obviously really bad.

dnwe:
@rhs the only reason for the re-run of the version.sh was for re-creating the .tar.gz for older proton releases <= 0.8 which were tagged in the repository without the version tags having been properly set

i.e. if you ran export.sh for 0.8 you actually get 1.0-SNAPSHOT in the pom.xml

dnwe:
Removed by 81c2110

PR #29, opened 2015-05-06
Reactor

prestona:
This is an initial implementation of a proton-j version of the reactor.  It comes with samples, unit tests and interoperation tests.   It should be good enough to "kick the tyres" but requires further test cases, code cleanup and Javadoc.  For the most part, the code is additive to the proton-j codebase.


rhs:
Given my recent rant about new development on master, let's try pulling this in on a branch first. I've create the proton-j-reactor branch. If you can resubmit against that I'll pull it in and raise the issue on the 0.10 release thread of whether we want to include it in the release.

prestona:
Closing.  Will submit again against proton-j-reactor

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/EventImpl.java
```diff
@@ -157,12 +165,51 @@ public void dispatch(Handler handler)
...
+            break;
+        case TIMER_TASK:
+            handler.onTimerTask(this);
+            break;
         default:
             handler.onUnhandled(this);
             break;
         }
+
+        Iterator<Handler> children = handler.children();
```
gemmellr: Commenting here to avoid spamming the semi-unrelated JIRA mentioned linked with the other PR that got merged.

I have only skimmed this PR, since I haven't got much of experience of the reactor code in any of the other languages, so I'm not sure what many things are actually meant to do.

It felt a little strange at times that some of the Connection etc objects now have some very reactor specific methods even though you might not be using the reactor with them (i.e all existing use), but I can certainly live with that if thats how it works elsewhere, and they presumably dont hurt anything if you dont use them.

I did spot this 1 specific bit though, where I think it would be nice to optimise for what anyone using this currently will be doing. Existing handlers wont have any children, and that would currently mean that every time you dispatch an event you will create a pointless iterator, so it would be nice to avoid that in the case of no children.

PR #34, opened 2015-05-25
PROTON-881: Graceful handling of runtime exceptions thrown within han…

prestona:
Adds support for more gracefully handling runtime exceptions thrown from within a handler.

When a runtime (unchecked) exceptions is thrown within a handler, the reactor now catches it and throws a HandlerException (which is a checked exception).  The HandlerException references the Handler that threw the original exception and also chains the original exception as its "cause".

One facet of this PR that I'm not 100% happy with is the changes to EventImpl.  Really EventImpl.dispatch() should throw HandlerException - but this would be a pretty disruptive change, so instead I've added a 'lastHandler' attribute to the class which remembers the last handler to be dispatched.  This means that if a handler throws an exception inside the dispatch() method the calling class can discover which handler was responsible (even if dispatch() has recursed  into one, or more, level of child handler).

If there's a better way to do this (I'm sure it'll be really obvious in hindsight...), please let me know and I'll resubmit the PR with the changes.

### DIFF
--- a/examples/java/reactor/src/main/java/org/apache/qpid/proton/example/reactor/Cat.java
+++ b/examples/java/reactor/src/main/java/org/apache/qpid/proton/example/reactor/Cat.java
@@ -30,6 +30,7 @@ import org.apache.qpid.proton.Proton;
 import org.apache.qpid.proton.engine.BaseHandler;
 import org.apache.qpid.proton.engine.Event;
 import org.apache.qpid.proton.reactor.Reactor;
+import org.apache.qpid.proton.reactor.HandlerException;
 import org.apache.qpid.proton.reactor.Selectable;
 
 public class Cat extends BaseHandler {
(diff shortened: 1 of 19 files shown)


rhs:
Hi Preston, sorry to take so long to look at this.

My impulse here would be to avoid using a checked exception for HandlerException. I generally follow the advice that you should only use checked exceptions for error conditions that a programmer can anticipate and recover from, and I'm not sure that's the case here. It's nice to have the HandlerException for debugging and clarity so it's obvious when the handler is at fault rather than the reactor implementation itself, but I'm not sure how someone could catch and recover if it is thrown. Making it unchecked would also allow you to avoid the EventImpl changes you mention in your comment.

prestona:
No problem.  I'll re-work and submit another PR based around HandlerException being unchecked.

PR #35, opened 2015-06-03
C++ binding for proton.

alanconway:
Cliff Jansen has been working on an event-driven C++ binding for proton, similar to the python binding.
The work is not complete but I think it is almost ready to move to trunk for easier access.  Please see proton-c/bindings/cpp/README.md for  a TODO list. I'll be working on the TODO list this week and will move the binding to trunk next week if there are no objecctions. Any other comments are most welcome.


CMakeLists.txt
```diff
@@ -20,16 +20,15 @@ cmake_minimum_required (VERSION 2.6)
 
 project (Proton C)
 
+# Enable C++ now for examples and bindings subdirectories, but make it optional.
+enable_language(CXX OPTIONAL)
+
```
astitcher: The code block below already conditionally enables C++, the blocks should be merged

examples/cpp/example_test.py
```diff
@@ -0,0 +1,105 @@
+#
```
astitcher: Is this file being added by mistake? (It seems to be a python example not C++)

astitcher:
There seem to be a couple of extraneous go example files that have snuck into this PR

proton-c/bindings/CMakeLists.txt
```diff
@@ -109,6 +109,11 @@ if (EMSCRIPTEN_FOUND)
   set (DEFAULT_JAVASCRIPT ON)
 endif (EMSCRIPTEN_FOUND)
 
+# Prerequisites for C++
```
astitcher: I don't understand what this block is doing

proton-c/bindings/cpp/README.md
```diff
@@ -0,0 +1,24 @@
+# C++ binding for proton.
+
+This is a C++ wrapper for the proton reactor API.
```
astitcher: It's not a "wrapper for the proton reactor" API it's a "C++ binding for the Proton API"

astitcher:
Some overall comments (from a not yet thorough) read through:

* I really don't like using set/get in a C++ API. It's not idiomatic C++ API style; it's not necessary; and it doesn't read as well (IMO) -

So instead of `Proton::transport::setXXXX('blah')`/`Proton::transport::getXXXX()`, can we please use `Proton::transport::XXXX('blah')`/`Proton::transport::XXXX()`.

We use idiomatic properties in Python we should use them in C++ too.

* I'm not sure that we really need to use the Pimpl pattern for this binding. The real need for a Pimpl is to insulate this API from changes to code that it depends on - in this case once the C API is stable then its own API stability requirements should ensure that nothing gets broken in the C++ API too. I may not have thought this through hard enough though [is that a sentence with too many 'ough's or what ;-) ]

This will save the indirection and extra resource use that Pimpl brings, which would force 2 allocations for each of the structures.

This strikes me as important because the C++ API can be as efficient as the C API if we do this work well, and it will be much easier to write and read code in good C++ than C. So I think that no one should need to do any application work in C even for applications that need the full efficiency of low level control of the library.

examples/cpp/example_test.py
alanconway: No, it is a test script that runs the C++ examples and verifies they work.

CMakeLists.txt
alanconway: enable_language(CXX OPTIONAL) enables cmake support for C++ if it is available. I had to move it up from proton-c because CMake requires you do this above any subdirs that will use it (now examples and proton-c). The next block just decides whether or not to use C++ for building proton. The C++ binding is default enabled if C++ is available (checks CXX_COMPILER), and can explicitly enabled/disabled with BUILD_CPP like all other bindings.

proton-c/bindings/CMakeLists.txt
alanconway: There is a foreach at the end of the file that sets up all the user-settable binding options. For each binding you must set a DEFAULT_FOO to be on or off depending if the requirements to build the binding are available, then the foreach sets up the BUILD_FOO option. CMAKE_CXX_COMPILER is set if a C++ compiler is available.

proton-c/bindings/cpp/README.md
alanconway: Agreed, fixed.

alanconway:
On Fri, 2015-06-05 at 11:13 -0700, Andrew Stitcher wrote:
> Some overall comments (from a not yet thorough) read through:
> I really don't like using set/get in a C++ API. It's not idiomatic 
> C++ API style; it's not necessary; and it doesn't read as well (IMO) 
> So instead of
> Proton::transport::setXXXX('blah')/Proton::transport::getXXXX(), can
> we please use
> Proton::transport::XXXX('blah')/Proton::transport::XXXX().
> We use idiomatic properties in Python we should use them in C++ too.

100% agree, will fix. Those get/sets in Qpid have annoyed me for years.

> I'm not sure that we really need to use the Pimpl pattern for this 
> binding. The real need for a Pimpl is to insulate this API from 
> changes to code that it depends on - in this case once the C API is 
> stable then its own API stability requirements should ensure that 
> nothing gets broken in the C++ API too. I may not have thought this 
> through hard enough though [is that a sentence with too many 'ough's 
> or what ;-) ]
> This will save the indirection and extra resource use that Pimpl
> brings, which would force 2 allocations for each of the structures.

Agreed. Cliff also raised this question. He added some extra state to
those classes to make things more user-friendly, but I think we should
push such "helper" functionality to separate classes and keep a layer
that is simply a C++ wrapper of a C pointer.

> This strikes me as important because the C++ API can be as efficient 
> as the C API if we do this work well, and it will be much easier to 
> write and read code in good C++ than C. So I think that no one should 
> need to do any application work in C even for applications that need 
> the full efficiency of low level control of the library.

Yup, agreed.

proton-c/bindings/cpp/include/proton/cpp/Acceptor.h
```diff
@@ -0,0 +1,50 @@
...
+
+namespace proton {
+namespace reactor {
+
+class Acceptor : public ProtonHandle<pn_acceptor_t>
+{
+  public:
+    PROTON_CPP_EXTERN Acceptor();
+    PROTON_CPP_EXTERN Acceptor(pn_acceptor_t *);
+    PROTON_CPP_EXTERN Acceptor(const Acceptor&);
```
astitcher: With C++11 onwards need to consider move construction and move assignment too (this is a general point), so if class is movable (and if it is copyable it probably is moveable too) need to add:
`Class(Class&&)`/`Class& operator=(Class&&)`

Actually though it would be good C++11 style to declare all 5 special functions for every class even if they are default just to show that we really had thought about it:
```C++
class C {
#ifdef __cplusplus >= 201193L
  C(const C&) = default;               // Copy constructor
  C(C&&) = default;                    // Move constructor
  C& operator=(const C&) & = default;  // Copy assignment operator
  C& operator=(C&&) & = default;       // Move assignment operator
#endif
  virtual ~C() { }                     // Destructor
...
};
```

alanconway:
On Fri, 2015-06-05 at 10:37 -0700, Andrew Stitcher wrote:
> There seem to be a couple of extraneous go example files that have 
> snuck into this PR
> —
Ooops, removed.

proton-c/bindings/cpp/include/proton/cpp/Acceptor.h
alanconway: C++ is unbelievable. You have a default so you don't need to constantly re-type repetitive completely uninformative code, and now we have a special syntax so you can type something to show that you know you don't have to type it.

CMakeLists.txt
astitcher: You are correct. I think that we could now move the block below somewhere else - it was only here to get the project command in the top level file.

proton-c/bindings/cpp/include/proton/cpp/Acceptor.h
astitcher: Not quite - that syntax tells the compiler to definitely generate the default implementation. It won't always otherwise, subject to irritating ules that might change as they are a little contraversial

proton-c/bindings/CMakeLists.txt
astitcher: [Of course I should have known that, as I wrote that foreach logic!]

astitcher:
Another general point (I can't exactly see where to attach it in the changes):

As well as using a class hierarchy to get callbacks, we should think about using functions, as C++11 has a nice lambda syntax and std::function too.

Well when I say nice lambda syntax I mean succinct and reasonably readable syntax. `[](){}` anyone?

astitcher:
There are a number of places in the API where the API has something like:
`void setXXX(const char* bytes, size_t size);` this isn't very C++ like - in fact it adds very little value over the original API whilst still adding an extra layer wrapping it. If we are going to add a wrapper like this we should be making it more C++ like either using `void XXXX(const std::string&);` or `void XXXX(const std::vector<char>&);` (one or other or both depending on the context).

This is equally true for getters as setters, although for these we might need to create some wrapping types to avoid copying if the value returned is a pointer into an existing value.

alanconway:
On Fri, 2015-06-05 at 13:05 -0700, Andrew Stitcher wrote:
> There are a number of places in the API where the API has something 
> like:
> void setXXX(const char* bytes, size_t size); this isn't very C++ like 
> - in fact it adds very little value over the original API whilst 
> still adding an extra layer wrapping it. If we are going to add a 
> wrapper like this we should be making it more C++ like either using 
> void XXXX(const std::string&); or void XXXX(const 
> std::vector<char>&); (one or other or both depending on the context).
> This is equally true for getters as setters, although for these we 
> might need to create some wrapping types to avoid copying if the 
> value returned is a pointer into an existing value.

Agreed. I would be inclined to do std::string in the C++ API and
provide access to the underlying pn_foo_t* for edge cases.

gemmellr:
Can we close this old PR, doesnt seem like its being used anymore?

alanconway:
Out of date, latest is on master branch.

PR #37, opened 2015-06-15
PROTON-881: Graceful handling of runtime exceptions thrown within han…

prestona:
…dlers

When a runtime (unchecked) exceptions is thrown within a handler, the reactor now catches it and rethrows it as a HandlerException (which is also an unchecked exception).  The HandlerException references the handler that threw the original exception and also chains the original exception as its "cause".

Note: this PR makes a (slight) change in behaviour to the Event.dispatch() method.  Currently if an application throws an unchecked exception from within the dispatch() method, this exception will peculate up the stack until it is caught.  With this PR: any unchecked exceptions thrown by a handler will be trapped inside the dispatch() method and re-thrown linked to a HandlerException.  I /think/ that this is unlikely to break (m)any applications - as relying on exact runtime behaviour when throwing unchecked exceptions from callbacks is a rather risky thing to do...

However: if this is a problem I can move the code for trapping unchecked exceptions out of the EventImpl class.  The disadvantage to doing this is that it will no longer be possible to identify the specific handler that threw the exception.


prestona:
(Not sure what is happening with the Travis CI build - as I can't reproduce the test case failure if I pull the reactor branch and build locally)

prestona:
This was merged by @rhs as cd42324

PR #40, opened 2015-07-01
PROTON-881: Make connect a non-blocking operation

prestona:
Previously connecting was a blocking operation which meant that if the server
being connected to took a long time to accept the connection this blocked all
other work in an instance of the Reactor.

This commit makes connect a non-blocking operation, allowing the reactor to
continue processing other work while the connection is established (or not).

Unfortunately, I've not found a satisfactory way to test this behavior in the
test suite - because Java never blocks during connect if it is using the
loopback adapter.  Instead, to test the non-blocking connect code path, I had
to configure firewall rules to drop all packets sent to a particular port.


rhs:
FYI, I merged this onto trunk after landing the reactor branch.

prestona:
Closing as merged.

PR #41, opened 2015-07-03
PROTON-931: proton-j expose detached state of Link

dnwe:
The Link detached state is required to assess when a
LINK_REMOTE_DETACHED event is received as to whether or not it was as a
result of a local detach or the remote end detaching or closing the
link.


gemmellr:
Looks good to me.

PR #45, opened 2015-07-09
-Fix 2 Code Analysis warnings in iocp.c and a realloc leak

dcristoloveanu:
-Fix one realloc leak and the fact that the realloc result was not checked. This rippled through codec.c as more functions needed to have error checking. Also, more fixes regarding return values being checked need to be done throughout message.c, transport.c, etc., but the amount of changes is significant, so it needs to be done in several chunks.

### DIFF
--- a/proton-c/include/proton/error.h
+++ b/proton-c/include/proton/error.h
@@ -40,6 +40,7 @@ typedef struct pn_error_t pn_error_t;
 #define PN_TIMEOUT (-7)
 #define PN_INTR (-8)
 #define PN_INPROGRESS (-9)
+#define PN_OUT_OF_MEMORY (-10)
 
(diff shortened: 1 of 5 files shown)


astitcher:
I've committed these changes now, I did fix them up a bit:

Specifically:
- I changed the logic in pn_data_add() to use early return too - I must have missed that in my review.
- I coalesced a couple of declarations with first use.
- I removed a now unecessary code block
- I removed the error checking from uses of pn_data_enter/exit as these don't return an error code at all. Rather the returned boolean has meaning as defined in the documentation. In the use here it can never return false.

dcristoloveanu:
Thanks, appreciate your help with the review!
/Dan

PR #47, opened 2015-07-17
NO-JIRA: Change travis configuration to use container based builds

astitcher:
This is specifically for comment from @dnwe -  I'm proposing to change the travis build config to use containers, but I've stopped using the existng jenkins build script.

This does make the travis CI builds start almost immediately for me, so is a big win in my experience.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,35 @@
----
-language: python
-install:
-- sudo apt-get update -qq
-- sudo apt-get install -y -qq bash cmake libssl-dev maven ruby ruby-dev
-    python python-dev python3 python3-dev php5 openjdk-7-jdk swig uuid-dev
-    valgrind libsasl2-dev sasl2-bin
-- pip install tox
-- gem install rspec simplecov
-before_script:
-- export PATH=${HOME}/.local/bin:${PATH}
-- export PYTHON_PATHS=$(ls -d /opt/python/*)
-- echo PYTHON_PATHS=${PYTHON_PATHS}
-- for PYTHON_DIR in ${PYTHON_PATHS} ;
-    do export CMAKE_PREFIX_PATH=${CMAKE_PREFIX_PATH}:${PYTHON_DIR} ;
-  done
-- echo CMAKE_PREFIX_PATH=${CMAKE_PREFIX_PATH}
-script:
-- bin/jenkins-proton-c-build.sh
 os:
 - linux
+sudo: false
+language:
+- python
+python:
+- 2.7
+addons:
+  apt:
+    packages:
+    - cmake
+    - libssl-dev
+    - libsasl2-dev
+    - sasl2-bin
+    - swig
+    - python-dev
+    - valgrind
+    - ruby
+    - ruby-dev
+    - python3-dev
+    - php5
+    - openjdk-7-jdk
+install:
+- pip install --upgrade pip
+- pip install tox
+- gem install rspec simplecov
+- gem install minitest --version 4.7.0
+before_script:
+- export PATH=${HOME}/.local/bin:${PATH}
+script:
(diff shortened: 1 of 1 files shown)


astitcher:
It looks to me that the appveyor failure is due to some incidental issue with the service.

dnwe:
Changes look reasonable to me. Faster sudoless docker builds would be cool. 

You dropped the PYTHON_DIRS bit which was for tox to locate the various Python interps. With the new format do we still get coverage on all the pythons?

astitcher:
The tox thing is the bit I'm shakiest about - it seems to find 3 out of
the 4 pythons (not 2.6 I think) for some reason. I experimented with a
matrix test which also builds under python 3.4 and think that would be a
better way to cover multiple pythons using travis. However proton didn't
build with python 3.4 like that!


On Fri, Jul 17, 2015, at 06:29 PM, Dominic Evans wrote:
> Changes look reasonable to me. Faster sudoless docker builds would
> be cool.


> You dropped the PYTHON_DIRS bit which was for tox to locate the
> various Python interps. With the new format do we still get coverage
> on all the pythons?


> — Reply to this email directly or view it on GitHub[1].


>

Andrew



Links:

  1. https://github.com/apache/qpid-proton/pull/47#issuecomment-122435936

dnwe:
Ok. :+1. Ship it!

gemmellr:
Do the earlier comments mean that it no longer tests on 2.6 but did previously? Is that perhaps because the newer infrastructure doesn't include 2.6? or perhaps since the config is now specifying 2.7 explicitly? Possibly something worth mentioning more directly on the list.

Similarly, is Ken aware of the issue you saw with Python 3.4? Might also be worth pointing out.

dnwe:
@gemmellr the error message for py26 is quite odd. It seems like its failing on the new "compile proton from python if libqpid-proton isn't available to link against" that was added to the setup.py. I'm not sure why the libqpid-proton build via cmake isn't considered usable though.

Here's the output log

```
Processing ./.tox/dist/python-qpid-proton-0.10.0.zip
Building wheels for collected packages: python-qpid-proton
  Running setup.py bdist_wheel for python-qpid-proton
  Complete output from command /home/travis/build/apache/qpid-proton/proton-c/bindings/python/.tox/py26/bin/python2.6 -c "import setuptools;__file__='/tmp/pip-zTErui-build/setup.py';exec(compile(open(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" bdist_wheel -d /tmp/tmpTPU6sFpip-wheel-:
  running bdist_wheel
  running build
  running build_ext
  running configure
  building '_cproton' extension
  swigging cproton.i to cproton_wrap.c
  swig -python -threads -o cproton_wrap.c cproton.i
  cproton.i:394: Error: Unable to find 'proton/cproton.i'
  error: command 'swig' failed with exit status 1
  
  ----------------------------------------
  Failed building wheel for python-qpid-proton
Failed to build python-qpid-proton
```

gemmellr:
@dnwe Yeah, I'm a bit lost with these bits I'm afraid hence the sugestion of mentioning any issues more directly on the list so those with a clue are aware something needs attention, as they likely won't have noticed these comments.

astitcher:
@gemmellr  I'm going to try to characterise the python 3 build issue a bit better then raise a JIRA.

gemmellr:
@astitcher ok great

astitcher:
I've raised [PROTON-951](https://issues.apache.org/jira/browse/PROTON-951) and [PROTON-952](https://issues.apache.org/jira/browse/PROTON-952) - which are related bugs to this discussion. [PROTON-952](https://issues.apache.org/jira/browse/PROTON-952) may be related to the python 2.6 tox issue.

PR #48, opened 2015-07-29
PROTON-964: Proton-J extensible event types

bozzzzo:


proton-j/src/main/java/org/apache/qpid/proton/engine/BuiltinHandler.java
```diff
@@ -0,0 +1,71 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+package org.apache.qpid.proton.engine;
+
+public interface BuiltinHandler extends Handler {
```
gemmellr: 'Builtin' seems a bit off as a name. Perhaps 'Core' / 'CoreEvents' or something else along those lines?

proton-j/src/main/java/org/apache/qpid/proton/engine/BuiltinHandler.java
```diff
@@ -0,0 +1,71 @@
...
+    void onLinkFinal(Event e);
+
+    void onDelivery(Event e);
+    void onTransport(Event e);
+    void onTransportError(Event e);
+    void onTransportHeadClosed(Event e);
+    void onTransportTailClosed(Event e);
+    void onTransportClosed(Event e);
+
+    void onReactorInit(Event e);
```
gemmellr: I think it could be argued that the reactor-specific events might be better in their own extended handler, but they aren't currently so lets save that for later ;)

proton-j/src/main/java/org/apache/qpid/proton/engine/Event.java
```diff
@@ -80,15 +83,46 @@
...
         SELECTABLE_ERROR,
-        SELECTABLE_FINAL
+        SELECTABLE_FINAL,
+        
+        /**
+         * This value must never be used to generate an event, it's only used as
+         * a guard when casting custom EventTypes to builtin {@link Type} via
+         * {@link Event#getBuiltinType()}.
+         */
+        NOT_A_BUILTIN_TYPE;
```
gemmellr: NOT_A_BUILTIN_TYPE similarly seems a bit off as a name. Perhaps following in the theme of the work, something like EXTENDED, or following the earlier suggestion, NON_CORE, etc?

Also, for the javadoc link, the Event#getBuiltinType()  method doesnt exist.

proton-j/src/main/java/org/apache/qpid/proton/engine/Record.java
```diff
@@ -29,6 +29,10 @@
 public interface Record
 {
 
+    public interface Accessor<T> {
```
gemmellr: Can this be in its own file, e.g 'RecordAccessor'?

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/BuiltinEventTypeImpl.java
```diff
@@ -0,0 +1,144 @@
+package org.apache.qpid.proton.engine.impl;
+
+import org.apache.qpid.proton.engine.BuiltinHandler;
+import org.apache.qpid.proton.engine.Event;
+import org.apache.qpid.proton.engine.Handler;
+
+public class BuiltinEventTypeImpl {
```
gemmellr: This seems to be dispatcher-specific, might be clearer if we named it so?

Same as before regarding the 'Builtin'.

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/CollectorImpl.java
```diff
@@ -57,9 +58,13 @@ public void pop()
         }
     }
 
-    public EventImpl put(Event.Type type, Object context)
+    public EventImpl put(EventType type, Object context)
     {
-        if (tail != null && tail.getType() == type &&
+        if (type == null)
```
gemmellr: Missing braces on if statements.

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/CollectorImpl.java
```diff
@@ -57,9 +58,13 @@ public void pop()
...
+            throw new IllegalArgumentException("Type cannot be null");
+        if (!type.isValid())
+            throw new IllegalArgumentException("Cannot put events of type " + type);
+        if (tail != null && tail.getEventType() == type &&
```
gemmellr: Would an equals() check be more appropriate now that EventType isn't necessarily an Enum (though granted it is natural and we know it is for the core events)?

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/EventImpl.java
```diff
@@ -53,9 +54,9 @@
         this.type = null;
     }
 
-    void init(Event.Type type, Object context)
+    void init(EventType type2, Object context)
     {
-        this.type = type;
+        this.type = type2;
```
gemmellr: Is there a reason to rename this?

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/EventImpl.java
```diff
@@ -68,12 +69,19 @@ void clear()
...
     @Override
-    public Type getType()
+    public EventType getEventType()
     {
         return type;
     }
 
     @Override
+    public Type getType() {
+        if (type instanceof Type)
```
gemmellr: Braces on if statement.

It would be nice if we didnt need to do an instanceof check, though I can only see making all non-core events implement getType() fixing that... :/

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/EventImpl.java
```diff
@@ -217,6 +101,19 @@ public void dispatch(Handler handler)
             dispatch(children.next());
         }
     }
+    
+    @Override
+    public void redispatch(EventType as_type, Handler handler) throws HandlerException
```
gemmellr: A little unclear why this is needed, but ignoring that...it seems like using it inside a handler (as in the test) would lead to 'redispatch' to the current and child handlers before the original event is then subsequently given to any child handlers. Expected? Worth documenting?

gemmellr:
I've commented on the code, but one final random thought:

Could we use generics to enforce the correct type of handler is supplied for dispatch and avoid some instanceof checks? (e.g  BuiltinEventTypeImpl.java)

Its possibly more trouble than its worth, but it feels a little weird we dispatch to/with only 'Handler' now when it no longer has any methods on it of use in dispatching except onUnhandled().
Given the way all the children for a handler also get called that might not be particularly doable, unless the children are similarly forced to be of compatible types.

Of course, that whole process also suggests that onUnhandled is usually going to have to be a no-op anyway, or some way would be needed to filter only particular events to particular children.

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/BuiltinEventTypeImpl.java
```diff
@@ -0,0 +1,144 @@
+package org.apache.qpid.proton.engine.impl;
```
gemmellr: Missing a license header.

proton-j/src/test/java/org/apache/qpid/proton/engine/EventExtensibilityTest.java
```diff
@@ -0,0 +1,381 @@
+package org.apache.qpid.proton.engine;
```
gemmellr: Missing a license header.

proton-j/src/main/java/org/apache/qpid/proton/engine/BuiltinHandler.java
bozzzzo: Ack

proton-j/src/main/java/org/apache/qpid/proton/engine/BuiltinHandler.java
bozzzzo: Yeah, that's where I want(ed) to go, but it would make for a much more invasive change.

proton-j/src/main/java/org/apache/qpid/proton/engine/Event.java
bozzzzo: I'll go with something on the theme NON_CORE_EVENT?

The getBuiltinType() is from a previous iteration of the patch and should have been renamed to getType()

proton-j/src/main/java/org/apache/qpid/proton/engine/Record.java
bozzzzo: Ack

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/BuiltinEventTypeImpl.java
bozzzzo: This is out-of-line implementation of EventType interface for Event.Type#dispatch.

CoreEventDispatcher?

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/CollectorImpl.java
bozzzzo: ack

proton-j/src/test/java/org/apache/qpid/proton/engine/EventExtensibilityTest.java
```diff
@@ -0,0 +1,381 @@
...
+    // //////////////
+
+    public static class ExtraEventTypeImpl {
+        /**
+         * Implementation of dispatch method for this extension
+         * 
+         * @param type
+         *            The implementation should accept only it's own
+         *            {@link EventType} instances
+         * @param handler
```
gemmellr: Forgot to comment earlier. The javadoc params dont match the method signature.

(I know, only a test, but we have more than enough javadoc warnings/errors already :P)

proton-j/src/main/java/org/apache/qpid/proton/engine/Event.java
gemmellr: That would also work yep.

(I dislike needing to have it at all, but I can see why its there)

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/BuiltinEventTypeImpl.java
gemmellr: Yep that seems good.

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/CollectorImpl.java
bozzzzo: Frankly, I'm not even sure why this check is needed at all.
But extensibe events don't need to change collector semantics. the change here is
a mostly about robustness and a bit about the fact that not only core events are now possible.
If needed this should be a separate change with separate justification.

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/EventImpl.java
bozzzzo: uhuh, will fix

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/EventImpl.java
bozzzzo: Why this needs to be so is more obvious if you consider how Core events (maybe they should be Engine events) would look if reactor had the ReallyReallyCore events :)

Wdyt, replace Core with Engine in all above comments and leave Core for the deeper refactoring.

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/CollectorImpl.java
gemmellr: I believe it is coaslescing 'duplicate' events.

Your explanation of the changes you did make here reinforces why I think it should probably be changed to equals() now. Before your changes, it wasnt possible for the == check here to do the wrong thing. Following these changes in future it will be possible (even if not that likely) for this to do the wrong thing and it wont necessarily be the easiest thing to find then, whereas it is just now.

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/EventImpl.java
bozzzzo: This is for now the only way to actually generate non-core, events...

I'll extend the javadoc to make it explicit. This was the minimal change needed and right now behaves as you describe.

A typical example would be a message decoder handler that turns a DELIVERY event into a MESSAGE event with attached message in the event. Or a timer for periodic sampling of a Connection, and so on.

I'm in fact sitting on some more reactor/collector changes that were done later and for other use cases. One of them is the ability for a handler to have children process the event synchronously.
I can also fold the implementation of that change in here as it would then allow redispatch() to have child handlers be invoked first with the original event, before redispatching the new type.
And keep exposing the interface for that in a separate jira.

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/EventImpl.java
gemmellr: I'd be more inclined to drop the Reactor events from the Type given that there are currently no users of those events yet (its never been released), whereas the others have been part of the main API for using the engine for quite some time and changing them would definitely hurt a bunch of folks. That said, my comment wasn't really about anything like that, just noting a dislike of needing to do the instanceof but also recognising why :)

I'm not sure about using 'Egnine'. I did actually think of that name earlier and decided agaisnt suggesting it. I'd say the older events are 'more core' in some ways than the new reactor ones, with the reactor being built on top of the eingine...but I guess also has IO related events that dont. Having both 'engine' and 'reactor' events named Core right now seems nicer in some ways than having the reactor events named 'Engine' events. I dunno. Mainly I just disliked 'Builtin' :)

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/CollectorImpl.java
bozzzzo: Can you please qualify 'the wrong thing', I really want to understand this?

With this overall change, I'm only working around the `final` of the enum, not around other aspects of it's semantics. This change brings us about on-par with proton-c imho where EventType === int.

bozzzzo:
> Could we use generics to enforce the correct type of handler is supplied for dispatch and avoid some instanceof checks? (e.g BuiltinEventTypeImpl.java)

I'm not sure how this would work, because it's easy (and I think necessary) to mix and create a tangle of handlers of different descent in a tree (or even DAG) of handlers with Handler#add(Handler).

So an Event instance will get dispatched to many handlers, and necessarily only some will be of an assignable type. I'm not sure how to express that with generics.

The core can afford the luxury of declaring 'everybody extend me' but that is not viable for a library that sits on top of proton.

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/EventImpl.java
bozzzzo: > I'd be more inclined to drop the Reactor events from the Type given that there are currently no users of those events yet (its never been released)

If by drop you mean move reactor events to a future o.a.q.p.reactor.ReactorEvent.Type then
yes, that is exactly what I had in mind for a perfect world. It would be a code-massacre, tho. Deleting them outright would be a mistake, they are useful, and used ;)

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/CollectorImpl.java
gemmellr: 'wrong thing' as in fail to coalesce the 'duplicate' events like it is trying to, in the case an EventType instance ever finds its way in there that is not an Enum and isn't used via a single instance, which wouldnt have been possible before but is now, however unlikely.

There are other cases where the code does an == that I didn't say anything about because they are comparing against a specific enum value so their behaviour can't change, but it just seemed like in this case when the line was already being changed anyway it would be simple to avoid any future issues by making the update now.

The comment wasnt meant to be some big thing :)

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/EventImpl.java
gemmellr: Yes, 'drop from Type' as in move somewhere else, not get rid of entirely as they are obviously used.

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/EventImpl.java
gemmellr: Ok. Sounds like you might be using this in a slightly different way than I was thinking, which is likely because I haven't actually used any of the reactor (or handler) stuff at all yet :)

It would be a good idea for some other folks to look this over who actually have, like @rhs and @prestona.

gemmellr:
> I'm not sure how this would work, because it's easy (and I think necessary) to mix and create a tangle > of handlers of different descent in a tree (or even DAG) of handlers with Handler#add(Handler).
>
> So an Event instance will get dispatched to many handlers, and necessarily only some will be of an assignable type. I'm not sure how to express that with generics.

Yeah, the rest of my reply was getting at that. It just seemed a little icky that everyone who implements a handler should need the instanceof check and call unhandled othewise. Not saying I have an answer :)

> The core can afford the luxury of declaring 'everybody extend me' but that is not viable for a library that sits on top of proton.

Not sure I understood this bit. If I was writing a library on top of Proton I'd possibly want to declare my own 'everybody extend me' point if things outwith that library were expected to extend it (to try and insulate somewhat from changes like this in Proton).

bozzzzo:
@gemmellr thanks for the review! I think I addressed all the things you pointed out.

bozzzzo:
@gemmellr I think I have managed to eliminate almost all casts by delegating half of the dispatch to the handler -- the handler already knows it's type, no need for generics even.

proton-j/src/main/java/org/apache/qpid/proton/engine/EventType.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Entry point for external libraries to add event types. Event types should be
+ * <code>final static</code> fields. EventType instances are compared by
+ * reference.
+ * <p>
+ * Event types are best described by an <code>enum</code> that implements the
+ * {@link EventType} interface, see {@link Event.Type}.
+ * 
+ * @author bozzo
```
gemmellr: Can you drop all the @author tags, we dont tend to use them.

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/CollectorImpl.java
```diff
@@ -57,9 +58,16 @@ public void pop()
...
+    public EventImpl put(EventType type, Object context)
     {
-        if (tail != null && tail.getType() == type &&
+        if (type == null) {
+            throw new IllegalArgumentException("Type cannot be null");
+        }
+        if (!type.isValid()) {
+            throw new IllegalArgumentException("Cannot put events of type " + type);
+        }
+        // XXX: @gemmellr points out that type equality is wrong for event de-duplication. What about context equality?
```
gemmellr: May now be wrong rather than definitely is wrong was my point, its now ambiguous due to the change to an interface for the event type...though I do wish I had just never bothered making the comment :)

Context is probably more interesting, in that as it has been used it was to tie a particular proton object to a particular non-proton object for reference, so I can see there may be cases where you actually wouldnt want to de-dup for equivalent contexts and so using equals() instead of == could actually be wrong for the context.

bozzzzo: I can respin or drop the commit :)

gemmellr:
Hi Bozo, sorry for the delay getting back to you, I was a bit tunnel-visioned trying (and failing :/) to get 0.10 out the door this week.

I like the recent change giving the handler a method to handle events. I'm not sure it really reduces (so much as perhaps moves) the casting much, but it at least gives the Handler interface something to do other than onUnhandled and also gives the Handler the decision over how to handle an Event rather than the other way around, which seems nicer.

It would still be good for some other folks who actually wrote the event/reactor stuff to take a look (like @rhs and @prestona, <poke>) but master is open for 0.11 at this point....

proton-j/src/main/java/org/apache/qpid/proton/engine/impl/CollectorImpl.java
gemmellr: Drop it.

PR #52, opened 2015-12-17
Expose get/set on 'properties Map' of Link 'Attach' performative

SreeramGarlapati:
Hello Folks!

I (sreeram, from windows azure eventHubs team) am starting to build eventHubs Java SDK and taking a dependency on proton-j (which will be available on GitHub, hopefully, soon). EventHubs amqp client will need to set AttachProperties to enable Critical scenarios for us.

Currently, in the proton-j framework - all attach performative fields are exposed using the Link Interface. So, following the pattern I added new Get/Set methods for setting AttachProperties.

Please embrace this code change as soon as possible.

Thanks!
Sree

### DIFF
--- a/proton-j/src/main/java/org/apache/qpid/proton/engine/Link.java
+++ b/proton-j/src/main/java/org/apache/qpid/proton/engine/Link.java
@@ -21,6 +21,7 @@
 package org.apache.qpid.proton.engine;
 
 import java.util.EnumSet;
+import java.util.Map;
 
 import org.apache.qpid.proton.amqp.transport.ReceiverSettleMode;
 import org.apache.qpid.proton.amqp.transport.SenderSettleMode;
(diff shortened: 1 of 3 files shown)


SreeramGarlapati:
SSL tests were failing - for which the Current merge has nothing to do with.

SreeramGarlapati:
https://issues.apache.org/jira/browse/PROTON-1082 - The JIRA request...

astitcher:
I looked at this - I can't see any problems, but I don't have enough real expertise in the Java code to be comfortable saying it's ok and merging it.

gemmellr:
Thanks Andrew. I was planning to get to this once the final bits for the Proton 0.11.1 release are out the way.

I think it needs adjustment for naming consistency with the similar Connection level bits, and as only the 'local' half has been implemented here it wouldn't satisfy the needs of either servers or any clients looking to inspect the remote link properties, so I'd add that too. I'll look at doing that later today.

gemmellr:
Hi Sreeram,

I have committed changes to implement support for both setting the local properties to be sent to the peer and inspecting the remote properties conveyed from the peer.

The local bit is fairly close to your changes, other than updating the names for consistency the main change is that it adds type details to the Map and restricts the keys to be Symbol values, since symbols are the only type of key AMQP 1.0 spec permits in the case of the link properties.

Robbie

(Mentioning PROTON-1082 here so the bot updates it too, since it isn't in the PR title)

SreeramGarlapati:
Hello Robbie,

Thank you very much for the merging the Change & especially for the Quick turnaround during Holiday timeframe. I deeply appreciate it.

Looked through your changes - make perfect sense.

Sree

PR #53, opened 2016-01-12
[PROTON-1096]: expose seting MessageFormat on Transfer frames in proton-j

SreeramGarlapati:
This is a feature request from windows azure eventhubs team.

I (sreeram, from windows azure eventHubs team) am building eventHubs Java SDK and taking a dependency on proton-j (which will be available on GitHub, hopefully, soon). EventHubs amqp client will need to set Custom MessageFormat to enable a Critical scenario for Sending (Send Batch) messages to EventHubs from our JavaClient.

https://issues.apache.org/jira/browse/PROTON-1096

### DIFF
--- a/proton-j/src/main/java/org/apache/qpid/proton/engine/Delivery.java
+++ b/proton-j/src/main/java/org/apache/qpid/proton/engine/Delivery.java
@@ -42,6 +42,8 @@ public interface Delivery extends Extendable
     /** TODO is this required?? */
     public int getMessageFormat();
 
+    public void setMessageFormat(int messageFormat);
+
     /**
      * updates the state of the delivery
      *
--- a/proton-j/src/main/java/org/apache/qpid/proton/engine/impl/DeliveryImpl.java
+++ b/proton-j/src/main/java/org/apache/qpid/proton/engine/impl/DeliveryImpl.java
@@ -50,6 +50,7 @@ public class DeliveryImpl implements Delivery
     private boolean _remoteSettled;
     private DeliveryState _remoteDeliveryState;
     private DeliveryState _defaultDeliveryState = null;
+    private int _messageFormat = 0;
 
     /**
      * A bit-mask representing the outstanding work on this delivery received from the transport layer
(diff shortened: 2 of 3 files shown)


SreeramGarlapati:
@gemmellr : pls. do the needful...

gemmellr:
Hi Sreeram,

I have made some changes that should let you do what you want. They change the API a little by usling a long to make it more intuitive (the message-format field is an unsigned int) and also implement the incoming side so that [proton-j] recievers can inspect the value should it have been adjusted.

Robbie

As an aside, some hints/tips regarding the Apache<->GitHub integration:

If you create a JIRA before the pull request and include the JIRA key in the PR title then a bot will update the JIRA with notice of creation of the PR, and also seperately for any comments made on the PR.

If you also include the JIRA key in any commits, then if they are merged as-is their details will be added as a comment on JIRA when pushed to the ASF repo, helping keep the details together for later reference.

For example, see #6 and its associated JIRA PROTON 814 (missed out the dash to avoid it being updated).

(mentioning PROTON-1096 here to have the bot update it too)

SreeramGarlapati:
Awesome! Thanks a lot Robbie.
Details about JIRA integration is helpful. Thanks!

gemmellr:
Hi Sreeram,

As you might have already seen from the JIRA update or commit just now, I have made a further update here to go back to using an int to carry the 32bit value for the message format. My change to use a long had caused some compatibility issues that I don't think it is important enough a change to warrant, so I've put it back to the int. 

If you were just using an int (as your original PR would have needed), or perhaps explicitly casting a long to int, then you shouldnt have to change anything, but if you were now passing in a long directly after the earlier changes you will need to make some updates too, in which case apologies for any hassle caused.

Robbie

tourili:
Hi
I have a question about the proton-j. Is proton-j a valid option for developing **android client** to connect to **azure service bus**. Please if any one can confirm a way or another to acheive this.
I appreciate a lot

Thanks

gemmellr:
I have replied to your thread on the mailing list, please keep general discussion there.

tourili:
Thank you. Sorry for hijacking this thread.
I am answering you on the mailing list...

SreeramGarlapati:
Not a problem Robbie! Will merge this soon.
Thanks for the Update!

PR #57, opened 2016-01-14
PROTON-1088 - Added two functions pn_ssl_get_cert_fingerprint() and p…

ganeshmurthy:
…n_ssl_get_remote_subject_subfield() to obtain certificate fingerprint and certificate subject subfields respectively


astitcher:
Also note that the integration tests have failed - You should investigate why that is, and comment on it if it has nothing to do with your change.

PR #58, opened 2016-01-14
Fix nullPointerException in event impl in the Reactor.onSessionInit codePaths

SreeramGarlapati:
Eventhubs java client depends on Proton-j Reactor APIs. While running our stress scenarios - we encountered a NullPointerException in this codepath - onSessionInit command (it happens transiently though). Will fill in with full stack trace if I encounter this error again.

(tagging @gemmellr for helping with merge).

### DIFF
--- a/proton-j/src/main/java/org/apache/qpid/proton/engine/impl/EventImpl.java
+++ b/proton-j/src/main/java/org/apache/qpid/proton/engine/impl/EventImpl.java
@@ -239,10 +239,12 @@ class EventImpl implements Event
             return (Transport) context;
         } else if (context instanceof Connection) {
         	return ((Connection)context).getTransport();	
-        } else if (context instanceof Session) {
-        	return ((Session)context).getConnection().getTransport();
         } else {
-            return null;
+        	Session session = getSession();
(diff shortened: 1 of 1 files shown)


gemmellr:
Hi Sreeram,

Do you actually mean that there was an NPE inside EventImpl as the title suggests? Given usage of the getTransport() inside EventImpl method, it doesnt seem like that would be the case or that the change would affect it.

If on the other hand you mean that a (your?) onSessionInit() handler calls getTransport() on the session init event (not entirely clear why it would?) and then NPE's when using the result then I guess I can see that. Its not particularly clear why that would only happen transiently though unless there is actually another issue at play. I think seeing a stacktrace would be good before making any changes.

P.S. no need to tag me when creating each PR, the mails all go to the same place so it wont typically make me spot it or do anything with it any quicker ;)

gemmellr:
As it won't cause any harm I've pushed a change (essentially yours with the whitespace fixed) for this via PROTON-1105 to close out the PR, but as mentioned if you were seeing unpredictable behaviour around this then that probably points to some other issue.

PR #63, opened 2016-01-19
Event tidying

astitcher:
Pull request for comment by @alanconway  & @cliffjansen

*Brief summary of these changes:*

- Rename API classes so that `event` and `handler` are the API classes users see.
- Change the inheritance hierarchy so that the internal event and handler classes are not related to the API classes
- Miscellaneous tidying
- One potential issue: the order of calling the flow_handler specified by the [messaging_]handler has changed - it is now called after the users handler.
  - This should be fixed by making this handler set up by the container/connection_engine from and option (as should all the other `handler` constructor arguments go away.
 - `handler` shouldn't really make the `messaging_adapter` the container/connection_engine should so that the constructor arguments can be link (or whatever) options.

These changes passed the tests for me, so hopefully they work - I'm sometimes seeing Travis hang in the ruby tests - which doesn't seem like it is related to these changes.


alanconway:
Ship it! Neater and cleaner, I rebased my connection_engine branch on it, the only changes I had to make were simpifications to the dispatch loop, everything works. I'll wait for this to go in before I commit the engine stuff.

If you want to look at latest engine see:
https://github.com/alanconway/qpid-proton/tree/cpp-engine-examples

cliffjansen:
Ship it!

PR #65, opened 2016-02-05
Schema parsing should not be so greedy

hekonsek:
Hi,

Current address schema parser doesn't work if `://` string is included in a path. The reasons to include `://` in a path are as follows:
- it is not against AMQP address specification
- ActiveMQ AMQP module uses [1] `topic://`prefix to indicate that message should be dispatched to broker topic, not queue (and other broker implementations could do the same)

So to connect to ActiveMQ's topic named `topicname`, you could use the following AMQP address - `host:423/topic://topicname` . The problem with current `Address` implementation is that it will interpret `://` in `topic://` as a schema separator.

Since `://` is optional in address and the only two acceptable schemas are `amqp` and `ampqs`, I think we should narrow schema parsing to those two values.

What do you think? 

[1] http://activemq.apache.org/amqp.html

### DIFF
--- a/proton-j/src/main/java/org/apache/qpid/proton/messenger/impl/Address.java
+++ b/proton-j/src/main/java/org/apache/qpid/proton/messenger/impl/Address.java
@@ -58,10 +58,12 @@ public class Address
     {
         clear();
         int start = 0;
-        int schemeEnd = address.indexOf("://", start);
-        if (schemeEnd >= 0) {
-            _scheme = address.substring(start, schemeEnd);
-            start = schemeEnd + 3;
+        if (address.startsWith("amqp://")) {
+            _scheme = "amqp";
+            start = 7;
+        } else if (address.startsWith("amqps://")) {
+            _scheme = "amqps";
+            start = 8;
         }
 
         String uphp;
--- a/proton-j/src/test/java/org/apache/qpid/proton/messenger/impl/AddressTest.java
+++ b/proton-j/src/test/java/org/apache/qpid/proton/messenger/impl/AddressTest.java
@@ -38,6 +38,7 @@ public class AddressTest {
     {
         testParse("host", null, null, null, "host", null, null);
         testParse("host:423", null, null, null, "host", "423", null);
+        testParse("host:423/topic://topicname", null, null, null, "host", "423", "topic://topicname");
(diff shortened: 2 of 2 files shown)


alanconway:
The problem is that the addresses used by proton are modelled on URLs, and those characters are not safe in a URL path:

From RFC 1738 specification:

Thus, only alphanumerics, the special characters "$-_.+!*'(),", and reserved characters used for their reserved purposes may be used unencoded within a URL.

The charcters ":" and "/" are reserved. Now protons addresses are not really standardized anywhere, and we could easily hack on the proton C URL parser to do whatever we like. However many of the proton bindings use the language-native URL parser instead of the proton C one so breaking the URL rules could get into trouble elsewhere.

Note that proton's parser (and any language-binding-native URL parser) should respect the standard URL percent-encoding for special characters, so that might be the best way to go if you want to encode topic:// as part of a URL path.

gemmellr:
I hadn't got round to looking at this closely but was likely just going to apply it if nothing obvious came up, and probably wouldnt have considered the above.

As per the diff, doesnt seem like java Messenger is using the native URL bits to parse its URL-like strings, and I don't see anything in there handling special char encodings, though I guess it could be hiding elsehwere (I don't really know the Messenger bits at all). It does kinda looks like whats there at present might accept the topic:// in the path if a scheme is given before it in the string, but I haven't tried that.

The idea that the schemes used should be looked for seems sensible, though perhaps other values should be rejected if that were to be done, but then it wouldn't play nice with without encoding special chars or again having the actual scheme present at the start.

astitcher:
I think that @alanconway  has a good point.

Looking at the C implementation of the url decoder it doesn't seem to expand the % escapes for the path, but as far as I can see it should - it would be simple to make it do that.

I would say that the Java and C url parsers should be closely aligned (even if they can't be exactly the same code)

ssorj:
In previous discussion on APIs and bindings, we have said that we'd like to use the language-native URL facilities if feasible.

hekonsek:
Hi,

Actually as vertx-proton API seems to favored over a messenger API, then feel free to ignore this pull request. We will be using vertx-proton anyway.

Cheers!

astitcher:
@hekonsek could you close this pull request now, as it is no longer for consideration.

hekonsek:
Done :) .

PR #67, opened 2016-02-19
PROTON-1139 tx_recv_interactive.py example doesn't work

scholzj:
This pull request should fix the tx_recv_interactive.py example.


scholzj:
I have to say that I haven't tested it on Windows, but I guess the original example didn't worked there either. Is there some way how to make the example work on Windows as well? If not, this change is probably better than nothing.

grs:
Thanks Jakub!

PR #68, opened 2016-02-23
Fixing broken documentation links

robreeves:
Several links were changed from *.html to *.md.

### DIFF
--- a/docs/markdown/index.md
+++ b/docs/markdown/index.md
@@ -1,9 +1,9 @@
 
 Proton is a library for speaking AMQP, including:
 
-- The AMQP [Messenger API](messenger/index.html), a simple but powerful interface to send and receive
+- The AMQP [Messenger API](messenger/index.md), a simple but powerful interface to send and receive
   messages over AMQP.
-- The [AMQP Protocol Engine](engine/engine.html), a succinct encapsulation of the full
+- The [AMQP Protocol Engine](engine/engine.md), a succinct encapsulation of the full
   AMQP protocol machinery.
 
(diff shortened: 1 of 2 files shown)


ssorj:
This is a little vexing.  Your patch fixes those links for the documentation as rendered in GitHub, but it will break them elsewhere.  We use markdown as a convenient way to produce static html, and static html requires the links to end in .html.  I think this is more an issue with GitHub than it is the documentation in this case.

astitcher:
Is there no way to tell the process that renders the markdown to html to inspect the links and modify them as necessary (if they refer to another markdown document that is also being rendered into html)?

ssorj:
https://github.com/github/pages-gem/issues/69

This ticket discusses the very same issue.  It appears to be unresolved, but there are a few hacks there.  Nothing that appealed to me right away.

alanconway:
What Andrew said. Sphinx handles this properly, the source .rst files have links to other .rst files - works nicely for dispatch docs. The HTML generator converts local .rst links to .html in the generated html. We could modify our HTML generator to do the same. 

Rant: This is a symptom of the fundamental problem with markdown: it was designed as a hack to generate HTML but touted as a generic markup language. So in practice markdown documents are riddled with HTML specific assumptions, like using .html in links, which break as soon as you try to use them in a new context. It's a pity because MD is so close to being a useful markup language that it consistently fools people. RST is much better but less popular, which means it's not really better since standard markup is only standard if it's widely used :(

robreeves:
Closing PR because it does not seem like this is the appropriate fix.

PR #70, opened 2016-03-10
PROTON-1158 - Removed hardcoded port 12345 from MessengerSSLTests. Us…

ganeshmurthy:
…ed free_tcp_ports() instead


ganeshmurthy:
Made the change you suggested. Thanks.

PR #72, opened 2016-03-31
NO-JIRA: Fix break in Javascript bindings build

prestona:
Attempting to build Proton with Emscripten on the path fails with the following
message:
/home/prestona/git/qpid-proton/examples/c/messenger/recv-async.c:26:10: fatal
error: 'pncompat/misc_funcs.inc' file not found
#include "pncompat/misc_funcs.inc"
         ^
1 error generated.

This appears to have been introduced by commit
84e416e389fe74ec7a1cab9252af2342fc95b83f

### DIFF
--- a/proton-c/bindings/javascript/CMakeLists.txt
+++ b/proton-c/bindings/javascript/CMakeLists.txt
@@ -192,11 +192,11 @@ target_link_libraries(qpid-proton-bitcode)
 
 
 # Compile the send-async.c and recv-async.c examples into JavaScript
-include_directories(${PN_PATH}/../examples/c/include)
-add_executable(send-async.js ${PN_PATH}/../examples/c/messenger/send-async.c)
+include_directories(${Proton_SOURCE_DIR}/examples/c/include)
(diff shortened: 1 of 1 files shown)


proton-c/bindings/javascript/CMakeLists.txt
```diff
@@ -192,6 +192,7 @@ target_link_libraries(qpid-proton-bitcode)
 
 
 # Compile the send-async.c and recv-async.c examples into JavaScript
+include_directories(${PN_PATH}/../examples/c/include)
```
astitcher: It might be better to avoid relative paths and use

${Proton_SOURCE_DIR}/examples/c/include

My spidey sense tells me that this would be more robust in the face of changes, but I can't really substantiate that.
This would also apply to the relative paths in the lines below.

astitcher:
This looks fine modulo the inline comment.

Do you need someone else to merge it?

prestona:
Hi @astitcher - I've added a second commit which uses ${Proton_SOURCE_DIR}, as per your comment.  If the CI builds complete successfully then: yes I would appreciate it if someone could merge this on my behalf.

PR #73, opened 2016-04-05
PROTON-1171: SimpleSSLTransportWrapper doesn't flush all bytes from the Decode buffer to the next transport layer's buffer

SreeramGarlapati:
Actual Issue/scenario hit by Microsoft Azure EventHubs:
We have a pattern where customers sends messages in a burst to our Queue and stop sending and then wait for all of them to be received. Because of this issue - many messages were stuck in the SSL Decode Buffer and were only picked up when new messages arrive.


proton-j/src/test/java/org/apache/qpid/proton/engine/impl/ssl/SimpleSslTransportWrapperTest.java
```diff
@@ -117,16 +117,6 @@ public void testInputIncompletePacketInThreeParts()
     }
 
     @Test
-    public void testUnderlyingInputUsingSmallBuffer_receivesAllDecodedInput() throws Exception
-    {
-        _underlyingInput.setInputBufferSize(1);
```
SreeramGarlapati: I deleted this Test from this File and Moved it to a brand new file for 2 reasons:
1) This function has a bug - SetInputBufferSize - sets only the integer value. It doesn't change anything with actual Buffer Size.
2) Since, changing buffer size will change the behavior of SimpleSslTransportWrapper - this could impact other testcases in the same file. So, moved the Size related test to a separate file...

gemmellr:
I have made a change that should resolve this. Its a little different than your proposal to both resolve an issue with it (unlikely but possible infinite loop), and also change some other existing behaviours I found while digging into it. If you could give it a try with your bits and report back that would be great.

SreeramGarlapati:
This is huge. Thanks @gemmellr !
We validated this in our stress run - you are correct - with my fix - we were still hitting another issue - which I was still looking at. After I packed proton with your commit - we are completely unblocked from receive stuck issue.

Looking forward for 0.12.2.

SreeramGarlapati:
After a couple of hours of our Stress run - we are still experiencing some bytes stuck in this layer. Have sent you a mail on the proton forum. Thx!

PR #74, opened 2016-05-31
proton-j : Upgrade BouncyCastle

JemDay:
Changed to use a <provided> dependency.
Removed 'reflective' constructs.
Verified correct test operations with BC 1.48 and 1.54

### DIFF
--- a/examples/cpp/selected_recv.cpp
+++ b/examples/cpp/selected_recv.cpp
@@ -47,7 +47,7 @@ namespace {
             << proton::binary(selector_str)
             << proton::codec::finish();
         // In our case, the map has this one element
-        map[filter_key] = filter_value;
+        map.put(filter_key, filter_value);
         opts.filters(map);
     }
 }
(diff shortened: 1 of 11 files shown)


JemDay:
I'm not sure what the CI build is doing to cause it to fail but i can confirm that "mvn clean test" works for me from the command line .. see attached
[build-log.txt](https://github.com/apache/qpid-proton/files/291766/build-log.txt)

gemmellr:
The CI builds are failing while running the cmake build (using it is outlined in DEVELOPERS.md) rather than the maven build. The cmake build creates a proton-j jar by itself for use in certain tests, a process which these changes break through introduction of a compile time dependency on bouncycastle that it doesn't accommodate because it didn't previously need to. Either the update needs to be made fully-reflective as before, or the cmake build updated to accommodate the new compilation dependency.

Some more general comments:
- The main module and tests are specifying difference versions of bouncycastle to use.
- If you include the JIRA title in the PR title, it will be updated by a bot upon PR open/comments/close.
- Please reference the JIRA in commit msg, i.e "PROTON-XYZ: msg".
- Please remove merge commits from PRs.

(Mentioning PROTON-1224 so it also gets updated)

JemDay:
OK .. i'll take a look at setting up a cmake environment and cleaning the commits..

JemDay:
Closing PR .. will recreate

PR #75, opened 2016-06-10
PROTON-1224: Upgrade BouncyCastle

JemDay:
Support 1.48+ of BouncyCastle
Added unit-test and associated certificate and key test resources.


tests/pom.xml
```diff
@@ -99,7 +99,7 @@ mvn test \
     <dependency>
       <groupId>org.bouncycastle</groupId>
       <artifactId>bcpkix-jdk15on</artifactId>
-      <version>1.47</version>
+      <version>1.48</version>
```
gemmellr: Given the idea is to update it probably makes sense to go for the latest version rather than this 3+ year old version (same goes for the other instance).

proton-j/src/test/resources/org/apache/qpid/proton/engine/impl/ssl/cert.pem.txt
```diff
@@ -0,0 +1,27 @@
+-----BEGIN CERTIFICATE-----
```
gemmellr: Please provide instructions on exactly how to (re)create the SSL resources used for the tests, so they can be easily updated later as needed.

E.g like the existing proton SSL test resource details: https://github.com/apache/qpid-proton/blob/master/tests/python/proton_tests/ssl_db/README.txt
Or the qpid-jms SSL test resource details (which can be sourced and execute directly): https://github.com/apache/qpid-jms/blob/master/qpid-jms-client/src/test/resources/README.txt

gemmellr:
I still need to have a final look, but if you could address the couple niggles I commented on that would be good. As I mentioned before, if you could also please remove merge commits from PRs that would be good, you have added a second in this version. Feel free to rebase the PR and squash all the commits, you don't need to open a new PR.

tests/pom.xml
JemDay: will do..

proton-j/src/test/resources/org/apache/qpid/proton/engine/impl/ssl/cert.pem.txt
JemDay: will do

JemDay:
Robbie - thanks for reviewing and sorry for the slow response, i'll look to try and resolve these today.

As previously discussed i'm not an enormous fan of this reflective approach but the cmake build process for java floored me a bit w.r.t how to achieve the maven style dependency at compile time.

JemDay:
OK ... my lack of high-wire skills with GITis really showing through now ... i'm completely stumped as to how to 'not' include other commits when pushing back to my fork ..

JemDay:
i think this is royally screwed up now, the file-change list looks clean but using a rebase rather than a pull (to remove the merge commits) now appears to be having an even worse affect.

I'm open to suggestions - i can close this PR, trash my fork, and then re-create the change if that's going to be a simpler path forward but before that i would like to better understand the fork/pull/commit/push/PR cycle when there may be multiple commits to the fork before the PR is actually raised. Squashing after commits have been pushed seems to just open world of hurt

gemmellr:
you just need to ensure you have the upstream repo up to date, then rebase your fork against it, then [force] push to your fork, which should result in your commit(s) sitting at the head of the branch with the current upstream commits preceding them and no unnecessary merge commits. You can repeat that process as many times as you like.

You obviously want to avoid force pushes on branches you are using to directly collaborate with other people on, but typically you commit work on a specific branch for the PR and then since noone else is likely to be depending on it doing such a force push is of little consequence to anyone.

JemDay:
I would like to close this PR and open a new one with (hopefully) a much cleaner change delta ...

JemDay:
Closing for re-creation

PR #76, opened 2016-06-30
PROTON-1249: Safeguard type initialisations.

dnwe:
In #readValue() for ArrayType, BinaryType, ListType and MapType
decoding, if the 'count' specified is very large then it is likely to
trigger an OutOfMemoryException. As these can come from an external data
source, during the SASL init for example, there is a potential for a
denial of service. The fix is to throw an IllegalArgumentException if
the count value is larger than the amount of data available in the
received bytes.


gemmellr:
Changes look reasonable. Some tests might be nice ;)

dnwe:
Yeah. I wasn't 100% sure about putting unittests in as then it would make the attack obvious ¯\_(ツ)_/¯

gemmellr:
raising the JIRA kind of does that too...if concerned about that, next time private@ :)

gemmellr:
Also, 0.13.1 is being cut tomorrow, this should probably go in it.

PR #77, opened 2016-08-11
Adds support for transport layers not currently implemented in the library

andrew-buckley:


andrew-buckley:
The Travis build error seems to be unrelated

gemmellr:
Hi Andrew,

I've created an alternative version resolving some niggles with the changes here (exposing various impl types, wildcard imports, licence header content, no tests). Its similar but you will need to cast to TransportInternal to get at the added method, as while I have no real issue with making the implementation itself support this I don't think it should be made part of the public Transport interface currently, so I took an inbetween approach.

https://github.com/gemmellr/qpid-proton/commit/d7c82bf0f14f2468d3aa2e39109003e00e3a4909
(or https://github.com/gemmellr/qpid-proton/commit/d7c82bf0f14f2468d3aa2e39109003e00e3a4909.patch to get the patch)

Give it a try out and let me know how you get on, if all is well I'll merge it to the apache master.

Robbie

(mentioning PROTON-1283 here to make this comment hit the related JIRA, as it was opened afterwards and wasn't in the original PR title)

andrew-buckley:
Hi Robbie,

Your proposed implementation works great! I understand the decision not to make this part of the Transport implementation.

Thanks,
-Andrew

PR #79, opened 2016-08-25
PROTON-1224: support newer BouncyCastle versions

gemmellr:
Update based on PR #78


gemmellr:
@JemDay I've made some updates to your patch, if youd like to give things a look over and try out and let me know how you get on that would be good, then I can merge it (might squash things first, left them separate for now).

JemDay:
Robbie

Thanks, i've taken a quick look and nothing leaps out at me, i'll try pulling this down and checking locally. I will try and get to this today but it might stretch to tomorrow.

I'd also like to discuss off-line another enhancement we'd like to make whereby we can provide/inject an SSLContext directly - this is the way we can conform with our in-house security constraints and mechanisms.

Jem..

JemDay:
Robbie..

As-per JIRA comment we were able to verify functionality with successful connections to a dispatch-router network.

Looks good, thanks for your patience.

J--

PR #81, opened 2016-09-15
PROTON-1303: Replace go binding URL parser with version that doesn't …

astitcher:
This PR is specially for @alanconway to review.

Note that interpretation of some URLs has now changed:
URLs without a host are now illegal (as they cannot be parsed as URLs)
":1223" is now illegal.
The empty URL "" does however default to localhost:5672

### DIFF
--- a/proton-c/bindings/go/src/qpid.apache.org/amqp/url.go
+++ b/proton-c/bindings/go/src/qpid.apache.org/amqp/url.go
@@ -40,7 +40,7 @@ func splitHostPort(hostport string) (string, string, error) {
 	}
 	if hostport[0] == '[' {
 		// There must be a matching ']' as already validated
-		if l := strings.LastIndexByte(hostport, ']'); len(hostport) == l+1 {
+		if l := strings.LastIndex(hostport, "]"); len(hostport) == l+1 {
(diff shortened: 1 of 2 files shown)


astitcher:
@alanconway Did you note the change in meaning for a few URLs?

alanconway:
OOps no: I don't like that this is not allowed ":1234" becase it *is* allowed by the go URL parser and is correctly interpreted as "listen on all interfaces" or "connect to localhost" by the go runtimes. It is commonly used in go documentation, e.g .the standard way to start the godoc webserver is 
  godoc -http :6060

I'm starting to feel that there isn't much/any need for a special URL parser if we simply provide the Update function to fill in defaults. Then we don't have to do anything to mess with parsing/what is acceptable as a URL, we just provide AMQP info - which is our job

astitcher:
":1234" is not allowed by the go URL parser - that is what is rejecting it. It gets rejected before our code has anything to do with it.

[That is because it is invalid as a URL - you can't have a leading ":" without a scheme.

alanconway:
You're right, godoc takes a host:port  not a URL. Ship It.

PR #82, opened 2016-10-06
optimizations on Decoder/Encoder

clebertsuconic:
- Avoiding multiple instances. The encoder and decoder can be used as singletons.
- There's still an option to instantiate your own encoder / decoder
- stateless Decoder/Encoder (can always reuse the same instance now)
- Implementing WritableBuffer and ReadableBuffers fully, so you could use NettyBuffers integrated.

### DIFF
--- a/proton-j/src/main/java/org/apache/qpid/proton/TestDecoder.java
+++ b/proton-j/src/main/java/org/apache/qpid/proton/TestDecoder.java
@@ -18,7 +18,12 @@
  */
 package org.apache.qpid.proton;
 
-import org.apache.qpid.proton.TestDecoder;
+import java.nio.ByteBuffer;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
 import org.apache.qpid.proton.amqp.Binary;
 import org.apache.qpid.proton.amqp.Decimal128;
 import org.apache.qpid.proton.amqp.Decimal32;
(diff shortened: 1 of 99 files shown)


clebertsuconic:
This is the first thing I wanted to do on the encoders / decoders.

There's a next step where I can integrate my own Buffer on Transports (or maybe a refactored version of it)...


This is at least the first step. Tests are running fine... and the performance tests I made are ok.

tabish121:
Would suggest removing the author tags from the new files, they are discouraged by the ASF

clebertsuconic:
@tabish121 done.. thanks

clebertsuconic:
@gemmellr there's a failure in one of these CIs.. what test is that? it doesn't seem related.

Can you check please?

gemmellr:
Its the python-based test suite that is failing to run against proton-j.

It seems like Appveyor have added Maven to the build nodes in the last couple weeks, which they didn't previously do, and the cmake build is failing to run the tests against proton-j now, whereas before it likely wasn't trying to run them at all (as noone downloaded maven in the Appveyor build config). I cant say I've ever run them on Windows myself, and certainly not via the cmake build...something to look at.

I've not yet had time to look properly at the PR itself, in part given how huge it is. Its obviously a breaking change anyway so it won't be going in the imminent 0.15.0 release, I'll look after that is out the way.

clebertsuconic:
I will send this again on the new repo (in a week or so)

PR #83, opened 2016-10-11
[Go binding] Replaced c handler based flowcontroller with native go

astitcher:
- The go binding now has no dependency on the proton-c reactor code

@alanconway - could you take a look at this an tell me what you think.

### DIFF
--- a/proton-c/bindings/go/src/qpid.apache.org/proton/handlers.go
+++ b/proton-c/bindings/go/src/qpid.apache.org/proton/handlers.go
@@ -19,8 +19,6 @@ under the License.
 
 package proton
 
-// #include <proton/handlers.h>
-import "C"
 import "fmt"
 
 // EventHandler handles core proton events.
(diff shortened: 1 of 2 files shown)


proton-c/bindings/go/src/qpid.apache.org/proton/handlers.go
```diff
@@ -308,7 +316,7 @@ func (d *MessagingAdapter) HandleEvent(e Event) {
 			d,
 		}
 		if d.Prefetch > 0 {
-			d.flowcontroller = cHandler{C.pn_flowcontroller(C.int(d.Prefetch))}
+			d.flowcontroller = flowcontroller{ window:d.Prefetch, drained:0 }
```
alanconway: Go cases do not fall through by default so your first 3 cases do nothing. I think you mean:

switch e.Type() {
case ELinkLocalOpen, ELinkRemoteOpen, ELinkFlow, Edelivery: ...

Otherwise it is  a thing of beauty, should have done this ages ago. :)

astitcher:
@alanconway Thanks for pointing out the stupid Cism in my go code.

I've fixed it now.

I'd much porefer to check this iin with sone sort of test - any clue how I should go about that?

alanconway:
Add module tests to proton_test.go. As you can see right now it is empty which is not ideal (!) The actual tests are in electron/electron_tests.go and the examples are also auto-tested by ctest. We should really have unit tests for proton itself though so if you want to start that off, please do :)

PR #85, opened 2016-11-07
PROTON-1330: [python] bundle the C source in the python source distri…

kgiusti:


astitcher:
A couple of comments:

* It seems entirely arbitrary (to me anyway) to move generating encoding.h and protocol.h from the C build process to generating them when producing the C part of the python install. We know that we have python available at the time of build the C code (well duh!) so we could keep on generating at C build time when we do the python install.

This seems to me to minimise the divergence between the packaged C source for the python package and the real C source tree.

* Given the c code built by the python install is somewhat selective anyway - it makes sense (again to me anyway) only to package C source which is actually used in the library used by the python installation.

kgiusti:
re: encoding.h and protocol.h:

I didn't want to duplicate the generation of these files in the python CMakeLists.txt but as it stands now these files are generated during the build phase - not the cmake configuration pass.   I can either 1) require the C compilation is done prior to building the Python source distribution or 2) move the file generation to the cmake configuration phase.  

re: packaging C source

I'm not entirely sure I understand - are you suggesting some of the C files are not needed by the Python bindings?  I assume I can hunt around for these files, but isn't all Proton/C functionality accessible via the python bindings, since the binding is used by the python unit tests?

kgiusti:
I've got a different approach that may address the issues you've raised Andrew.

PR #88, opened 2016-11-15
PROTON-1330: [python] bundle the C source in the python source distribution

kgiusti:
This alters the previous approach:  All c source files are enumerated in the CMakeLists.txt in the proton-c directory.  The python CMakeLists.txt introduces a new target for the python source distribution, using the source lists from the parent CMakeLists.txt


proton-c/bindings/python/CMakeLists.txt
```diff
@@ -122,3 +129,55 @@ install(TARGETS ${SWIG_MODULE_cproton_REAL_NAME}
...
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/README.rst.in
+               ${py_dist_dir}/README.rst
+)
+
+file(COPY ${py_dist_files} DESTINATION ${py_dist_dir})
+
+file(MAKE_DIRECTORY ${py_dist_dir}/proton)
+file(COPY ${pysrc} DESTINATION ${py_dist_dir}/proton)
+
+#file(COPY ${py_dist_proton_c_files}
```
astitcher: Do you really want to add commented out lines?

kgiusti: Oopsie - no, not at all.

I'll fix that.

astitcher:
I like this approach.

I was expecting to see the file lists substituted in setup.py.in somewhere. I didn't see that - how are the CMake file lists used inside setup.py to generate the source bundle?

PR #89, opened 2016-11-16
[PROTON-1352] Trivial casting from UnsignedByte to SenderSettleMode and ReceiverSettleMode

ppatierno:
Refactored ReceiverSettleMode and SenderSettleMode for having trivial casting from/to UnsignedByte
Added message annotations to the String visualization of AMQP message

### DIFF
--- a/proton-j/src/main/java/org/apache/qpid/proton/amqp/transport/ReceiverSettleMode.java
+++ b/proton-j/src/main/java/org/apache/qpid/proton/amqp/transport/ReceiverSettleMode.java
@@ -27,11 +27,28 @@ import org.apache.qpid.proton.amqp.UnsignedByte;
 
 public enum ReceiverSettleMode
 {
-    FIRST, SECOND;
+    FIRST(0),
+    SECOND(1);
 
-    public UnsignedByte getValue()
-    {
-        return UnsignedByte.valueOf((byte)ordinal());
+    private UnsignedByte value;
+
+    private ReceiverSettleMode(int value) {
+        this.value = UnsignedByte.valueOf((byte)value);
     }
 
+    public static ReceiverSettleMode valueOf(UnsignedByte value) {
+
+        switch (value.intValue()) {
+            case 0:
+                return ReceiverSettleMode.FIRST;
+            case 1:
+                return ReceiverSettleMode.SECOND;
+            default:
+                throw new IllegalArgumentException("The value can be only 0 (for FIRST) and 1 (for SECOND)");
+        }
+    }
+
+    public UnsignedByte getValue() {
+        return this.value;
+    }
 }
--- a/proton-j/src/main/java/org/apache/qpid/proton/amqp/transport/SenderSettleMode.java
+++ b/proton-j/src/main/java/org/apache/qpid/proton/amqp/transport/SenderSettleMode.java
@@ -27,11 +27,32 @@ import org.apache.qpid.proton.amqp.UnsignedByte;
 
 public enum SenderSettleMode
 {
-    UNSETTLED, SETTLED, MIXED;
+    UNSETTLED(0),
+    SETTLED(1),
+    MIXED(2);
 
-    public UnsignedByte getValue()
-    {
-        return UnsignedByte.valueOf((byte)ordinal());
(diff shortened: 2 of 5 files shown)


proton-j/src/main/java/org/apache/qpid/proton/amqp/transport/ReceiverSettleMode.java
```diff
@@ -25,13 +25,32 @@
...
-    {
-        return UnsignedByte.valueOf((byte)ordinal());
+    static {
+        for (ReceiverSettleMode mode: ReceiverSettleMode.values()) {
+            map.put(mode.value, mode);
+        }
     }
 
+    public static ReceiverSettleMode valueOf(UnsignedByte value) {
+        return map.get(value);
```
gemmellr: I think its worth throwing IAE if an out of bound value is provided, rather than just returning null.

proton-j/src/main/java/org/apache/qpid/proton/amqp/transport/SenderSettleMode.java
```diff
@@ -25,13 +25,33 @@
...
-    {
-        return UnsignedByte.valueOf((byte)ordinal());
+    static {
+        for (SenderSettleMode mode: SenderSettleMode.values()) {
+            map.put(mode.value, mode);
+        }
     }
 
+    public SenderSettleMode valueof(UnsignedByte value) {
+        return map.get(value);
```
gemmellr: As with the other mode.

proton-j/src/main/java/org/apache/qpid/proton/message/impl/MessageImpl.java
```diff
@@ -769,6 +769,10 @@ public String toString()
             sb.append("properties=");
             sb.append(_properties);
         }
+        if (_messageAnnotations != null) {
```
gemmellr: This should have its own JIRA. We can skip the separate PR, this time ;)

proton-j/src/main/java/org/apache/qpid/proton/amqp/transport/ReceiverSettleMode.java
tabish121: The map here seems unnecessary to me, the compiler adds a values method that returns an array of the enum values which you can just use to just do something like this

```
    public SenderSettleMode valueof(UnsignedByte value) {
        try {
            return values()[value.intValue()];
        } catch (Exception e) {
            throw new IllegalArgumentException();
        }
    }
```

ppatierno: It's true but it works in our case just because enum values go from 0,1 to 2 (so they are good as indexes in the array). If one of the value was 100 for example, using value.intValue() as index for the values() array will throw an out of bound exception. Your suggestion was the way I was using outside this enum in my application before opening the PR for doing the cast. :-)

ppatierno: @gemmellr I could be ok with switch but provided here as "out of box" :)

gemmellr: I'm actually fine with adding the method here, which is why I didn't say that originally ;)

tabish121: For this case I'd go with the switch statement if we decided we need this method, simple, effective.  Given the values are sequential there no need to use a solution based on what they could be.

gemmellr:
Including a message as to why the IAE is thrown wouldn't go amiss :)

I think you missed my comment about raising a separate JIRA for the unrelated annotations change, then including its key in the commit too. Can you use a simple "key(s): message" format for the log, e.g. "PROTON-1234, PROTON-5678: foo" (see the previous commits for more: https://github.com/apache/qpid-proton/commits/master).

gemmellr:
I hadn't realised you had done separate commits...can you add the relevant JIRA key to the first commit, then squash the other two together as a single commit and give it the other key.

PR #90, opened 2016-11-18
PROTON-1355: Set ssl.peer_hostname to virtual_host if specified

lulf:


proton-c/bindings/python/proton/reactor.py
```diff
@@ -565,7 +565,7 @@ def _connect(self, connection, reactor):
             if not self.ssl_domain:
                 raise SSLUnavailable("amqps: SSL libraries not found")
             self.ssl = SSL(transport, self.ssl_domain)
-            self.ssl.peer_hostname = url.host
+            self.ssl.peer_hostname = self.virtual_host if self.virtual_host != None else url.host
```
astitcher: I'm pretty sure that None has truthiness of false so this can be written more simply as:
... = self.virtual_host if self.virtual_host else url.host

lulf: Agreed. I chose to write it this way to be consistent with the rest of the code which seems to explicitly check for None.

astitcher: actually you can write it even more clearly ...
... = self.virtual_host or url.host

PR #91, opened 2016-12-10
PROTON-1375: add explicit cast to avoid clang 4.0 warning

jdanekrh:


astitcher:
This PR should be fixed by 79d06f07

PR #92, opened 2016-12-23
PROTON-1379: [C++ binding] Add #if'd #include <functional>

jdanekrh:


astitcher:
Yep this looks like a mistake - I 'll pull it in.

PR #93, opened 2017-01-10
PROTON-1312: fix memory leak on BlockingConnection.close()

cliffjansen:
Instrumenting the C code shows that Python BlockingConnections never
free their underlying connections, sessions and links, and frequently
the reactor too.

The BlockingConnection stops processing its dedicated reactor abruptly
on close preventing normal shutdown and cleanup operations from taking
place. Plus there are some circular references that keep some
structures alive.

The reactor C code keeps a connection and its children alive until
after the PN_CONNECTION_FINAL has completed dispatch processing.  The
BlockingConnection must continue processing the container at least
that far.  However, any unprocessed event can can contain a relevant
refcounted context so the safe route is to drain all remaining events
(including the PN_REACTOR_FINAL).

The SessionPerConnection class keeps the session alive by assigning a
context that is never used perhaps for some previous use in a handler.
This class is not currently used as a handler and its
on_session_remote_close is never called.  Removing the context
eliminates one needless refcount.

self.conn.free() is required to trigger a pn_connection_release() on
the blocking connection.

### DIFF
--- a/proton-c/bindings/python/proton/__init__.py
+++ b/proton-c/bindings/python/proton/__init__.py
@@ -2581,6 +2581,9 @@ and SASL layers to identify the peer.
     """
     self._update_cond()
     pn_connection_close(self._impl)
+    if hasattr(self, '_session_policy'):
+      # break circular ref
+      del self._session_policy
 
   @property
   def state(self):
--- a/proton-c/bindings/python/proton/utils.py
+++ b/proton-c/bindings/python/proton/utils.py
@@ -132,10 +132,12 @@ class BlockingReceiver(BlockingLink):
             raise LinkException("Failed to open receiver %s, source does not match" % self.link.name)
         if credit: receiver.flow(credit)
         self.fetcher = fetcher
+        self.container = connection.container
 
     def __del__(self):
         self.fetcher = None
-        self.link.handler = None
+        self.link.handler = None  # fails if reactor finalizes first
+        self.container = None
 
     def receive(self, timeout=False):
         if not self.fetcher:
@@ -222,12 +224,19 @@ class BlockingConnection(Handler):
             self, self.container.create_receiver(self.conn, address, name=name, dynamic=dynamic, handler=handler or fetcher, options=options), fetcher, credit=prefetch)
 
     def close(self):
+        if not self.conn:
+            return
         self.conn.close()
         try:
             self.wait(lambda: not (self.conn.state & Endpoint.REMOTE_ACTIVE),
                       msg="Closing connection")
         finally:
+            self.conn.free()
+            # For cleanup, reactor needs to process PN_CONNECTION_FINAL
+            # and all events with embedded contexts must be drained.
+            self.run() # will not block any more
             self.conn = None
+            self.container.global_handler = None # break circular ref: container to cadapter.on_error
             self.container = None
 
     def _is_closed(self):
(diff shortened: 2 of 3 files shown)


proton-c/bindings/python/proton/__init__.py
```diff
@@ -2581,6 +2581,9 @@ def close(self):
     """
     self._update_cond()
     pn_connection_close(self._impl)
+    if hasattr(self, '_session_policy'):
```
ssorj: Out of curiosity, why isn't _session_policy a stable attribute of the object?  That is, could we simply assign None here?  I would prefer that a bit if it makes no difference.  This likely predates your change, but I usually try to avoid having to test for the presence of attributes in Python objects.

proton-c/bindings/python/proton/utils.py
```diff
@@ -132,10 +132,12 @@ def __init__(self, connection, receiver, fetcher, credit=1):
             raise LinkException("Failed to open receiver %s, source does not match" % self.link.name)
         if credit: receiver.flow(credit)
         self.fetcher = fetcher
+        self.container = connection.container
 
     def __del__(self):
         self.fetcher = None
-        self.link.handler = None
+        self.link.handler = None  # fails if reactor finalizes first
```
ssorj: What happens with the error if it fails?  I suppose in that case it doesn't matter that the subsequent statement is not reached?

proton-c/bindings/python/proton/__init__.py
cliffjansen: Before my delta, the existing use of the attribute was via a hasattr() test.  Elsewhere in the code, hasattr attributes seemed to be decommissioned via del instead of None (more pythonic?). Functionally I have no preference between the two.

proton-c/bindings/python/proton/utils.py
cliffjansen: A core dump in pn_record_get.  That's why the container reference is set on init (line 135) and unset in the next line (line 140).  The order of decref to 0 differs for BlockingSenders and BlockingReceivers.  Sometimes the reactor is last, sometimes it precedes the blocking endpoints.  This bug lurked but was never hit because none of these objects were free'd before.

proton-c/bindings/python/proton/__init__.py
astitcher: The issue with just using 'self._session_policy = None' is that it would create a useless attribute at his point if it doesn't already exist.

But either solution works.

proton-c/bindings/python/proton/utils.py
ssorj: Maybe clarify that comment then.  As it stands, it appears to describe as normal a state that you have taken trouble to avoid.

proton-c/bindings/python/proton/__init__.py
ssorj: Gotcha, consistent with the existing code.

proton-c/bindings/python/proton/utils.py
alanconway: +1 the comment should read something like "this would cause a core dump if the reactor finalized first, but that is impossible because *explain refcounting tricks here*" It is important to leave some explanation, this is exactly the sort of thing that someone in future will think "this doesn't seem necessary" and change it without realizing they are re-introducing a heisenbug.

cliffjansen:
Adopted with review suggestions.  Thanks folks.

98e26f6 PROTON-1312: fix memory leak on BlockingConnection.close()

PR #95, opened 2017-02-21
PROTON-1412 Add fuzzers to proton-c tests

jdanekrh:
I am hoping to get some feedback about the directories where I placed the new files and about what I did in the CMakeFiles and about the fuzzers themselves and anything else I might've done wrong.

I will clean up the PR when I get general approval and advice about required changes.

For the final version I expect to

- [x] write proper commit messages (I did not mention Jira number so that it is not linked there)
- [x] add a Dockerfile to simplify building and running the fuzzers
- [x] rename files in corpus as they are apparently invalid on Windows
- [x] option(FUZZ_TEST must be OFF by default
- [x] LLVMFuzzerTestOneInput must be extern "C" when BUILD_WITH_CXX
- [x] delete commented out code and all those things
- [x] squash/fixup the commits


proton-c/src/tests/fuzz/fuzz-sniff-header.c
```diff
@@ -0,0 +1,15 @@
+#include <stdlib.h>
+#include <stdio.h>
+
+#include "core/autodetect.h"
+
+#ifdef __cplusplus
```
kcc: No need for extern "C" and surrounding ifdefs if the file is .c

proton-c/src/tests/fuzz/README.md
```diff
@@ -0,0 +1,45 @@
...
+    FLAGS="-fsanitize-coverage=trace-pc-guard -fsanitize=address -fno-omit-frame-pointer -fno-optimize-sibling-calls"
+    
+    CC=~/Work/third_party/llvm-build/Release+Asserts/bin/clang \
+    CXX=~/Work/third_party/llvm-build/Release+Asserts/bin/clang++ \
+    CFLAGS="$FLAGS" \
+    CXXFLAGS="$FLAGS" \
+    LDFLAGS="-L/path/to/LibFuzzingEngine/directory" \
+    cmake -DCMAKE_BUILD_TYPE=Debug -DFUZZ_TEST=ON ..
+
+### with afl-fuzz
```
kcc: in oss-fuzz AFL will work *exactly* like libFuzzer. 
Our environment will simply use different `libFuzzingEngine.a`

proton-c/src/tests/fuzz/fuzz-sniff-header.c
jdanekrh: There is a project specific cmake option BUILD_WITH_CXX which will build .c files with C++ compiler. See e.g. https://github.com/apache/qpid-proton/blob/9ec97b51c1fed9a77e5c3dc537e301fb7ae3c5fe/proton-c/src/tests/CMakeLists.txt#L30 Thank you very much for review.

astitcher: Even so, you shouldn't need this outside header files. whether the internal symbol names are C++ or C is not important unless they are visible externally.

astitcher:
I really like the idea of putting fuzzing tests into the source tree (guarded appropriately by cmake tests).

I'm not very keen about putting binary only files into the corpus. Could we use hex files perhaps? I'd even prefer in the future to use a text format like used by my project https://github.com/astitcher/amqp-value-parser.git

proton-c/src/tests/fuzz/fuzz-sniff-header.c
jdanekrh: The `LLVMFuzzerTestOneInput` function is to be visible externally. libFuzzingEngine provides a `main` function which calls it.

proton-c/src/tests/fuzz/README.md
jdanekrh: and different compiler options, presumably. Anyways, I have to have a separate doc explaining where to get afl-fuzz and what compiler flags to use. This is intended for local debugging. (Although I suppose running remote gdb in a docker container is workable...)

proton-c/src/tests/fuzz/fuzz-sniff-header.c
astitcher: The declaration only has to be extern "C" so you could put:
    #ifdef __cplusplus
    extern "C"
    #endif
    int LLVMFuzzerTestOneInput(uint8_t *data, size_t size);
In a header file. That is sufficient to give the function the correct linkage. In fact I'm surprised there isn't already a declaration like this in libfuzzer.

jdanekrh: Thanks, good point.

LubFuzzer does not have a header file. It is a single `.a` library. That is probably why.

jdanekrh:
I am not fan of storing binary data either, but I think it is not unreasonable given the circumstances.

The corpus was generated automatically by the fuzzer. There is a greedy algorithm for this: every time newly discovered input extends the total fuzzing coverage, it is added to the corpus.

* There are two seed files which I captured with Wireshark, the rest of the corpus is autogenerated. There would have to be something that parses the generated files into a text format.

* Some of the files in the corpus are not valid AMQP, the text format would have to be more flexible than what you linked, to capture arbitrary binary data, not only valid AMQP frames.

* The only advantage of text format I can see is ease of manual inspection and more meaningful git versioning. This is IMO not what will be happening. The operations done with the corpus in the future will be most likely occasional corpus minimization (to compact the number of files needed to reach the current coverage).

There is a possibility of a regression discovered with a file in the corpus (which is unlikely, because developers will be able to easily run all corpus files through fuzzer before they push changes (takes below a second), and fixing issues in a fresh change tends to be easier). In this case, you would use the fuzzer to give you a minimal input that still triggers the failure and you would analyze the minimal input, not the original one.

In summary, I think that corpus should remain in binary form, but there should be AMQP parser to help analyze files that trigger failures.

@astitcher btw, how did you go about fixing the two fuzzing issues. Did you run proton with the reproducer inputs attached in Jira in a debugger and stepped through until something unexpected occurred, or did you analyze the files manually to understand what is wrong with them?

jdanekrh:
I forgot to comment about hex files.

Sure, but then there would have to be a conversion step to hex when adding file to corpus and a conversion step from hex to binary before starting the fuzzer. The second conversion could be made part of the cmake build process. As I said previously, because the content of the files in the corpus never changes, I don't see a value in this, though.

astitcher:
I'd be happy with just hex files, but the binary files are virtually impossible to use sensibly without extra tools as a developer.

It might be a bit better with the fuzzing code checked in as you could just feed the failing test into the tester, but without it I essentially had to construct tests with a hex editor, as the binary sequences in the bug reports couldn't be simply injected into a running example program.

I really found and fixed the fuzzing bugs by reviewing the failing code! After looking at the stack traces.

But I had great difficulty in reproducing the original bugs so that I could show that I'd fixed them.

astitcher:
Also, in general I'm not usually happy to check in generated (especially) binary files to a source code repository.

I understand the purpose here, but it seems messy to me.

astitcher:
In general though it's more important to get this into the repo so it can be used going forward and then maybe it can be tidied up later.

kcc:
We use the approach with binary files in a few other repos (e.g. https://github.com/google/boringssl/tree/master/fuzz) and it works perfect. 
Note that this is not only a seed for fudzzing, but also a great way to extend your regression testing -- 
just run the fuzz target (w/o fuzzing it) on all inputs in the corpus in your pre-submit check and in CI. 

If you are reluctant to store the binary files in the same repo, you may create a separate one, 
like e.g. nss does: https://github.com/mozilla/nss-fuzzing-corpus

The files should contain the exact bytes that are used to pass into the target function, 
so if you store hex (text), the target function should consume hex (text), or otherwise you should have a conversion tools (messy)

kcc:
Re declaration of LLVMFuzzerTestOneInput -- libFuzzer does have a single interface header, 
FuzzerInterface.h, but we typically ask *not* to use it because we want the fuzz targets to be 100% independent of any particular fuzzing engine, so that e.g. AFL (or any other) 
can be used even if LibFuzzer is not present. 

This does cause discussions like this, I know. But otherwise is not very problematic. 
In the worst case if you declare LLVMFuzzerTestOneInput incorrectly, linking will fail.

astitcher:
Assuming that any compatible fuzzer has to have the same interface (assuming they all use LLVMFuzzerTestOneInput) then they all need to supply a compatible definition.

It's not as if this function is useful to compile unless there is a fuzzer present.

So I don't really buy that argument.

kcc:
on oss-fuzz we use adapters between e.g. AFL and LLVMFuzzerTestOneInput. 
See e.g. https://github.com/llvm-mirror/llvm/blob/master/lib/Fuzzer/afl/afl_driver.cpp
So, a fuzzer does not have to be compatible out of the box, but it should be possible to implement an adapter. (usually it is). 

> It's not as if this function is useful to compile unless there is a fuzzer present.

This is what I am trying to explain: this function *is* useful even if there is no fuzzing engine
linked to it. If there is a good corpus of inputs this function can and should be used for regression testing by linking it to a simple driver, e.g. [this one](https://github.com/llvm-mirror/llvm/blob/master/lib/Fuzzer/standalone/StandaloneFuzzTargetMain.c)

> So I don't really buy that argument.

Which one?

astitcher:
Ok, if the function is *useful* without the fuzzer then it makes sense to compile it without the fuzzer.

Sold.

proton-c/CMakeLists.txt
```diff
@@ -729,6 +729,12 @@ if (BUILD_PYTHON)
 
 endif (BUILD_PYTHON)
 
+# fuzz tests: tests/fuzz
+option(FUZZ_TEST "Enable building fuzzers and regression testing with libFuzzer" OFF)
+if (FUZZ_TEST)
+  add_subdirectory(src/tests/fuzz)
+endif (FUZZ_TEST)
```
jdanekrh: This should be probably changed. The fuzz tests should compile always, and switch like this will decide whether to link with actual fuzzer or something like `StandaloneFuzzTargetMain.c` (that would have to be committed here).

proton-c/src/tests/fuzz/fuzz-sniff-header.c
jdanekrh: done. I created the header file as Andrew suggested.

proton-c/CMakeLists.txt
jdanekrh: done, added another cmake variable `FUZZING_ENGINE` which decides whether to link with libFuzzingEngine or with a simple driver. It defaults to the simple driver.

jdanekrh:
I've done the cleanup.

One thing I came across is that I was not able to easily build the fuzzers with a simple driver on Windows, because MSVC does not have weak linking. Since both afl-fuzz and libFuzzer do not work on Linux anyway, I think it is OK to disable this with MSVC.

If I enable building and running the fuzzers with the simple driver, then Windows build is unchanged and Linux build shows results of the tests. The reason it is passing in Travis and not for me locally is that Travis does not enable sanitizers like `-fsanitize=address`.

CI builds with fuzzer enabled:

https://ci.appveyor.com/project/ke4qqq/qpid-proton/build/master.1117
https://travis-ci.org/apache/qpid-proton/builds/205374841

astitcher:
@jdanekrh I'm looking at this PR now. And I don't understand the purpose in renaming libFuzzer -> libFuzzingEngine.

Could you explain why you do that.

jdanekrh:
@astitcher Because there are at least two interchangeable libraries that can be used (libFuzzer and afl-fuzz). So instead of configuring this in the build with CMake, I adopted [the convention in OSS-Fuzz](https://github.com/google/oss-fuzz/blob/master/docs/new_project_guide.md#buildsh) and instead require that the intended library is renamed (or linked with `ln -s`) to that specific name. I wrote this in the README.md, https://github.com/apache/qpid-proton/pull/95/commits/85ff6e7d209aff30a46105b607e6946ef7b41222#diff-44029c0f93a63a627194c3f9fda0da85R64

I do not insist on doing it this way...

astitcher:
Ah, sorry, I didn't read it through to the end (it is perhaps a little wordy).

If the libs are interchangeable then which one is used need to be selectable from CMake. I'll see if I can make this work.

I've some other small changes to do anyway.

jdanekrh:
@astitcher About the proactor fuzzer. When I was preparing PR, proactor was broken, so I newer saw it run in the regressions mode.

I can run corpus minimization on the proactor corpus (I did not do it because proactor was broken) and see if it shrinks)

Alternatively, I would not mind removing the proactor tests. It forks itself for every input and there is a 1s timeout somewhere. Not good design for quick testing.

jdanekrh:
@astitcher anything more I could do around this, or do you want to take it up after the 0.18.0 release, or some other event like that?

jdanekrh:
BTW: adding fuzzers to qpid-proton will not automatically enable fuzzing in OSS-Fuzz. For that, somebody would then have to commit a Dockerfile and a shell script to the OSS-Fuzz repository. Which can be postponed until qpid-dispatch is ready for that. (Unless somebody just goes ahead and does it; but they could add it anyways, they would just have to keep the fuzzers in their own repository and enhance the bash script to apply them as a patch, or something...)

ssorj:
We'll try to get this for 0.20.0.

astitcher:
I've now committed a slightly modified version of this PR. So it can be closed.

jdanekrh:
@astitcher I will close the PR and I will have a look at your merge. There should be the Dockerfile for oss-fuzz in my checkout of proton somewhere. I will have a look into updating it and submitting PR against oss-fuzz project. (Do we want to wait with that for something more to happen?)

PR #98, opened 2017-03-13
PROTON-1434: add test blacklist to tox tests

kgiusti:
Helps speed up the time spent running the tox tests


astitcher:
Looks like a good idea to me

PR #99, opened 2017-03-21
PROTON-1408 -- fix for msg id rollover bug

mgoulish:
I'm getting a different hang now, but it looks like this part of the code is doing the right thing.
Before this, the original implementation would loop forever if 'last' equal to INT_MAX.

### DIFF
--- a/proton-c/src/core/transport.c
+++ b/proton-c/src/core/transport.c
@@ -109,9 +109,11 @@ void pn_delivery_map_del(pn_delivery_map_t *db, pn_delivery_t *delivery)
   if (delivery->state.init) {
     delivery->state.init = false;
     delivery->state.sent = false;
-    pn_hash_del(db->deliveries, delivery->state.id);
+    pn_hash_del(db->deliveries, delivery->state.id & BOTTOM_32_BITS);
   }
 }
+
+#undef BOTTOM_32_BITS
 
 static void pni_delivery_map_clear(pn_delivery_map_t *dm)
 {
--- a/proton-c/src/core/transport.c
+++ b/proton-c/src/core/transport.c
@@ -82,9 +82,11 @@ void pn_delivery_map_free(pn_delivery_map_t *db)
   pn_free(db->deliveries);
 }
 
+#define BOTTOM_32_BITS 0x00000000FFFFFFFF
+
 static pn_delivery_t *pni_delivery_map_get(pn_delivery_map_t *db, pn_sequence_t id)
 {
-  return (pn_delivery_t *) pn_hash_get(db->deliveries, id);
+  return (pn_delivery_t *) pn_hash_get(db->deliveries, id & BOTTOM_32_BITS);
 }
 
 static void pn_delivery_state_init(pn_delivery_state_t *ds, pn_delivery_t *delivery, pn_sequence_t id)
@@ -98,7 +100,7 @@ static pn_delivery_state_t *pni_delivery_map_push(pn_delivery_map_t *db, pn_deli
 {
   pn_delivery_state_t *ds = &delivery->state;
   pn_delivery_state_init(ds, delivery, db->next++);
-  pn_hash_put(db->deliveries, ds->id, delivery);
+  pn_hash_put(db->deliveries, ds->id & BOTTOM_32_BITS, delivery);
   return ds;
 }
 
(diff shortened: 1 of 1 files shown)


proton-c/src/core/transport.c
```diff
@@ -1619,6 +1619,17 @@ static int pn_scan_error(pn_data_t *data, pn_condition_t *condition, const char
   return 0;
 }
 
+/*
+  This operator, copied from code for the qpid cpp broker, gives the correct
+  result when comparing sequence numbers implemented in a signed integer type.
+*/
+static bool sequence_less_than ( pn_sequence_t a, pn_sequence_t b )
```
astitcher: Why not return less than 0 for less than 0 for equal and greater than 0 for greater than (like strcmp)

proton-c/src/core/transport.c
```diff
@@ -1648,7 +1659,7 @@ int pn_do_disposition(pn_transport_t *transport, uint8_t frame_type, uint16_t ch
   bool remote_data = (pn_data_next(transport->disp_data) &&
                       pn_data_get_list(transport->disp_data) > 0);
 
-  for (pn_sequence_t id = first; id <= last; id++) {
+  for (pn_sequence_t id = first; sequence_less_than(id, last) || (id == last); id++) {
```
astitcher: Then this test would be sequence_cmp(id, last)<=0

proton-c/src/core/transport.c
astitcher: ie code would be

static int sequence_cmp(pn_sequence_t a, pn_sequence_t b) {
    return a-b;
}

proton-c/src/core/transport.c
astitcher: Of course this test will still do the wrong thing if the wrap-around is big enough (ie if first > last at teh start of the loop)

mgoulish:
I have updated to use Andrew's cleaner sequence_cmp() idea, and added a new change: preventing sign-extension from pn_sequence_t to uintptr_t in the calls to pn_hash_put() and pn_hash_get().   This sign extension appears to be happening in one pathway but not the other, so that , when we wrap around to negative p_sequence_t values, pn_hash_get() and pn_hash_put() are getting different hash values.

I know how this sounds -- but with this change my test got through the 2^31 barrier for the first time ever.
It sounds like heresy -- but -- eppur si muove.

This is probably not the end of the story -- I had runaway memory consumption later, but the test got to 2.25 billion messages before I had to kill it.  Well past the 2^32 barrier, which == 2.147 billion.

mgoulish:
By the way -- to be able to check this much faster, rather than running a test that has to count all the way up to 2^31 messages, I have been changing one line in engine.c, from this:
pn_delivery_map_init(&ssn->state.outgoing, 0);
to this:
pn_delivery_map_init(&ssn->state.outgoing, 2145483648);
// NOTE -- that number is 2^31minus 2 million.

However, after this new fix passed that shortened test several times, I ran a full-sized test, starting at zero.  That was successful also.

proton-c/src/core/transport.c
```diff
@@ -1620,12 +1622,12 @@ static int pn_scan_error(pn_data_t *data, pn_condition_t *condition, const char
 }
 
 /*
-  This operator, copied from code for the qpid cpp broker, gives the correct
+  This operator, inspired by code for the qpid cpp broker, gives the correct
   result when comparing sequence numbers implemented in a signed integer type.
 */
-static bool sequence_less_than ( pn_sequence_t a, pn_sequence_t b )
+static bool sequence_cmp ( pn_sequence_t a, pn_sequence_t b )
```
astitcher: This should be ```inline```

proton-c/src/core/transport.c
```diff
@@ -82,9 +82,11 @@ void pn_delivery_map_free(pn_delivery_map_t *db)
   pn_free(db->deliveries);
 }
 
+#define BOTTOM_32_BITS 0x00000000FFFFFFFF
```
astitcher: Isn't this just UINT32_MAX?

astitcher:
Having looked a bit more it seems less clear that the bug is in pn_hash!

Essentially the lower 32 bit masking is used as a more determinate way to cast the int32_t of the pn_sequence_t to the uintptr_t (uint64_t) of the hash code I don't understand (yet) why this casting sometimes gives different results, but I *strongly suggest* creating a better named inline function to express the intent rather than doing the masking with no real explanation.

Viz:

    static inline uintptr_t pni_sequence_make_hash(pn_sequence_t i) {
        return i & 0xFFFFFFFFUL;
    }

astitcher:
Also I *strongly suggest* you test this code on both 32 and 64 bit machines to make sure that the casting works equally well when uintptr_t is 32 vs 64 bits.

This is independent of whatever solution you end up with.

mgoulish:
Closing this in favor of doing a new one, with latest improvement.  New PR will have just one checkin.

PR #101, opened 2017-03-30
PROTON-1453: properly terminate output string buffer

kgiusti:
Fixes the occasional segfault when running travis (as encountered by aconway)

### DIFF
--- a/proton-c/src/ssl/openssl.c
+++ b/proton-c/src/ssl/openssl.c
@@ -762,6 +762,7 @@ bool pn_ssl_get_protocol_name(pn_ssl_t *ssl0, char *buffer, size_t size )
   const SSL_CIPHER *c;
 
   pni_ssl_t *ssl = get_ssl_internal(ssl0);
+  *buffer = '\0';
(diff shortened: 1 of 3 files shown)


alanconway:
Ship It! Wow, that's an obscure bug found by obscure means.

astitcher:
@alanconway If the calls are documented to always return a value in the string (even upon failure) I won't complain.

As a matter of API design though, a filing call should not alter anything.

alanconway:
Disagree. IMO it is good design for all  return values of a failing call to be set to known failure indicators provided a) there are such indicators,  b) it is not costly to do so, and c)  it is unlikely there was valuable data in them before call. Provided the API doc is clear that the value will always be set (to "" in an error) that seems to me more robust than leaving random data lying around. 

I agree the tests are poorly written for not checking the return value but I think the more robust behavior leads to less support calls.

kgiusti:
I agree with Alan on this - but in this specific case proton is actually *inconsistent* in the handling of this buffer.  In the schannel impl the buffer is null terminated regardless of outcome in one logic path.  The openssl null terminates the fingerprint, cipher name, but not the protocol name.   This patch removes the inconsistency.

PR #102, opened 2017-03-31
NO-JIRA: revert temporary workaround for PROTON-1453

kgiusti:
Looks like the fix for PROTON-1453 resolves the problems travis hit for the tox tests.  This change reverts just the change where the travis config filtered out py33 and py34.


gemmellr:
Looks like the change was applied, but was no longer the head commit at the time, and didn't have "This closes #102" in the message of the commit itself or a merge commit. Can you close it?

PR #105, opened 2017-05-17
Port of C++ binding to Proton Proactor

astitcher:
This set of changes reimplements the C++ binding do that it no longer uses the Proton Reactor library at all and now only links to the Proton Core and Proton Proactor libraries.

It is not 100% finished yet, missing pieces include:
* Multi thread support - Currently the API is thread safe if you obey the thread safety rules, but only one thread is used to service handler callbacks - This will be fixed soon.
* Reconnect is not implemented - If a network connection fails no reconnect action is attempted - This will also be fixed, but ntoe that the C++ binding implementation of reconnect is subject to change as we don't have a lot of experience with it yet.

Included in this PR is a reworking of the mechanism for injecting work into safe contexts. The concepts have been renamed for clarity (`proton::work_queue` & `proton::work`) and some convenience functions have been added to help those (poor benighted souls) using C++03. There is some more work to come which should also improve injecting work using the scheduler.


proton-c/bindings/cpp/include/proton/message.hpp
```diff
@@ -329,11 +329,7 @@ class message {
     mutable annotation_map message_annotations_;
     mutable annotation_map delivery_annotations_;
 
-    /// Decode the message corresponding to a delivery from a link.
```
ssorj: How does this relate to the other changes here?

proton-c/bindings/cpp/include/proton/work_queue.hpp
```diff
@@ -0,0 +1,368 @@
+#ifndef PROTON_EVENT_LOOP_HPP
```
ssorj: This needs a rename.

astitcher: Will fix.

proton-c/bindings/cpp/include/proton/message.hpp
astitcher: This is a private function that didn't need exposing. So I moved it to the internals.

I can't remember anymore if this change was enabled by other things changed or not. But the code moved into message_adapter.cpp which also changed a lot in the same change.

astitcher:
I've now rebased and slightly modified some the previous commits.

Substantively I've addressed the work_queue header guard issue; finished the work dealing with scheduling work to occur in different contexts; renamed proton::defer to proton::schedule_work.

Note that the commits in the branch *are not* in the order display in this PR because they have been heavily reworked and reordered in rebases. So the creation time does not correspond to the commit order.

proton-c/bindings/cpp/src/handler.cpp
```diff
@@ -49,17 +50,33 @@ void messaging_handler::on_transport_error(transport &t) { on_error(t.error());
...
+        l.open(l.container().receiver_options());
+    }
+}
 void messaging_handler::on_receiver_detach(receiver &) {}
 void messaging_handler::on_sender_close(sender &) {}
 void messaging_handler::on_sender_error(sender &l) { on_error(l.error()); }
-void messaging_handler::on_sender_open(sender &) {}
+void messaging_handler::on_sender_open(sender &l) {
+    if (l.uninitialized()) {
+        l.open(l.container().sender_options());
```
alanconway: This looks like it will throw if the listener is not associated with a container. I want to still support pure connection_driver based application that might not have a "real" container impl  - we could have a "null container" with only a container ID. That can be fixed later - not an objection to this commit.

astitcher:
I've now updated this with what I intend to be the final rebase before merging.

Everything is functionally there except for reconnect, which will be addressed in the 0.19 timescale.

* Multhtreaded container is now implemented
* The (still) experimental connection_driver now works.

alanconway:
I'm still happy with this. I have some ideas about simplifying the returned<>/thread_safe<> story but lets get this up on master as a starting point.

PR #106, opened 2017-05-22
Added command line parameters for configure URL and address

ppatierno:
It refers to this https://issues.apache.org/jira/browse/PROTON-1488

### DIFF
--- a/examples/python/server.py
+++ b/examples/python/server.py
@@ -20,10 +20,9 @@
 
 from __future__ import print_function
 import optparse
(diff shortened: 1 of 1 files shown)


grs:
One minor issue: this doesn't work at present with the defaults. It seems to be because urlparse requires a scheme.  You can use Url from proton e.g. as in sync_client.py.

ssorj:
@ppatierno , would you close this.  It's now part of 0.18.0.

PR #108, opened 2017-06-09
PROTON-1500: Implement plugin interface for SASL implementations

astitcher:
- Allow sasl implementation to be selected per connection
- Compile both default and cyrus implementation (if we can)
- No current change in behaviour (get cyrus is available and default otherwise)
- But now have frqamework to solve PROTON-1209; Stil to do for this:
- Need to have API to turn on cyrus, with default being the initial implementation
- Need to deprecate cyrus specific APIs:
-- pn_sasl_extended
-- pn_sasl_config_path
-- pn_sasl_config_name
- Need to deprecate pn_sasl_done


astitcher:
@grs @rgodfrey @alanconway @ssorj You may want to take a look at this.

PR #114, opened 2017-07-27
[Ruby code Example] fix Messenger package

kbrock:
Some ruby examples have the correct package for `Messenger`, while others have a truncated package.

Thanks for the great gem and packages


kbrock:
Looks like master is not passing

kbrock:
thanks

PR #115, opened 2017-08-09
PROTON-1533: Fix swig deprecation warnings

ssorj:
For review

### DIFF
--- a/proton-c/bindings/php/CMakeLists.txt
+++ b/proton-c/bindings/php/CMakeLists.txt
@@ -34,7 +34,7 @@ list(APPEND SWIG_MODULE_cproton_EXTRA_DEPS
     ${CMAKE_SOURCE_DIR}/proton-c/include/proton/cproton.i
     ${PROTON_HEADERS}
 )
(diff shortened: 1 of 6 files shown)


proton-c/bindings/php/CMakeLists.txt
```diff
@@ -34,7 +34,7 @@ list(APPEND SWIG_MODULE_cproton_EXTRA_DEPS
     ${CMAKE_SOURCE_DIR}/proton-c/include/proton/cproton.i
     ${PROTON_HEADERS}
 )
-swig_add_module(cproton php ${CMAKE_CURRENT_SOURCE_DIR}/php.i)
+swig_add_library(cproton LANGUAGE php SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/php.i)
```
astitcher: Not directly related to your change, but I think the ${CMAKE_CURRENT_SOURCE_DIR} is extraneous here and is different from all the other uses of swig_add_library in the other bindings.

PR #118, opened 2017-08-25
PROTON-1326: update python setup.py to avoid using incompatible opens…

kgiusti:
…sl 1.1.0

Proposing this patch against 0.17.x to fix a problem in the python-qpid-proton pypi packages.

I don't think this requires another 0.17.x release since the change only relates to python packaging.  I simply need to rebuild the python packages on pypi with this fix.

### DIFF
--- a/proton-c/bindings/python/setup.py.in
+++ b/proton-c/bindings/python/setup.py.in
@@ -164,12 +164,15 @@ class Configure(build_ext):
         # Check whether openssl is installed by poking
         # pkg-config for a minimum version 0. If it's installed, it should
         # return True and we'll use it. Otherwise, we'll use the stub.
-        if misc.pkg_config_version_installed('openssl', atleast='0'):
+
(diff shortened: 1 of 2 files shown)


astitcher:
I'm assuming you are proposing to check this in branching from the 0.17 release? (It's difficult to tell from the PR). Note that the CI builds both fail - you will want to check this isn't due to this change.

kgiusti:
The failures don't seem to be related - various test failures but not the python-tox tests (which rely on setup.py patched by this patch).

Yes - I would have to create a 0.17.x branch from 0.17.0 tag for this commit.  The github gui didn't allow me to create a pull request against a tag so this PR is against master as that's the best I can do.

astitcher:
@kgiusti I think this PR is now irrelevant - as 0.18 is out and supports openssl 1.0 and 1.1.

Could you close it?

PR #120, opened 2017-09-26
PROTON-1369: Add deprecation warnings to the C++ binding

ssorj:
Changes

 - Add warnings for default_container.hpp, function.hpp, url.hpp, and some methods in value.hpp
 - Removed some deprecated type conversion functions that were never published as API
 - Convert examples that used the URL class to instead split the connection URL and address into two args
 - Some minor example improvements

Notes and questions

 - This approach requires users define PN_CPP_USE_DEPRECATED_API if they want to carry on without moving to the preferred APIs.
    - The alternative is #pragma message (just a warning), but that's not as portable as #error.  #warning is regrettably non-standard.
 - I'm using #define PN_CPP_USE_DEPRECATED_API/#undef for internal uses - what's the other approach?  I remember a recommendation to use an ORed set of internal variables, but I unfortunately lost the details.
 - I wasn't able to find the cmake support for deprecation macros.

### DIFF
--- a/proton-c/bindings/cpp/include/proton/default_container.hpp
+++ b/proton-c/bindings/cpp/include/proton/default_container.hpp
@@ -22,6 +22,8 @@
  *
  */
 
+#include "./internal/export.hpp"
+
 /// @file
 /// @copybrief proton::default_container
 


astitcher:
Note the #pragma is portable in that compilers are required to ignore pragmas they don't understand.

You can use _Pragma to use macros to expand to #pragma to make it more platform portable.

astitcher:
This is the CMake generated version of the deprecation macros:
```
#  ifndef PN_DEPRECATED
#    if defined(PN_COMPILER_CXX_ATTRIBUTE_DEPRECATED) && PN_COMPILER_CXX_ATTRIBUTE_DEPRECATED
#      define PN_DEPRECATED [[deprecated]]
#      define PN_DEPRECATED_MSG(MSG) [[deprecated(MSG)]]
#    elif PN_COMPILER_IS_GNU || PN_COMPILER_IS_Clang
#      define PN_DEPRECATED __attribute__((__deprecated__))
#      define PN_DEPRECATED_MSG(MSG) __attribute__((__deprecated__(MSG)))
#    elif PN_COMPILER_IS_MSVC
#      define PN_DEPRECATED __declspec(deprecated)
#      define PN_DEPRECATED_MSG(MSG) __declspec(deprecated(MSG))
#    else
#      define PN_DEPRECATED
#      define PN_DEPRECATED_MSG(MSG)
#    endif
#  endif
```

I note that it doesn't include the declaration itself - I take this to mean that you can choose the correct place to put the macro do it works for all the supported compilers - This certainly seems neater to me.

astitcher:
I'm not sure that it's actually worth putting out deprecation messages for including the files but not using the deprecated APIs there.

If the APIs aren't used it will be trivial for anyone to just remove the ``#include`` when the header files are actually removed.

ssorj:
"I note that it doesn't include the declaration itself - I take this to mean that you can choose the correct place to put the macro do it works for all the supported compilers - This certainly seems neater to me."

I don't get this one.  Isn't the idea that VS puts the attribute before the declaration, and GCC after, so you need to explicitly position the declaration?

ssorj:
Changes made per the review.

I took part of the logic from the cmake deprecation macro, but kept some of the conditions from the previous iteration because they are more consistent with the rest of the defines in export.hpp.

I will squash these before commit.

astitcher:
Now that I've merged the work_queue/work API changes I think you should deprecate the use of void_function0 and the work_queue/container APIs that use it.

astitcher:
Otherwise this looks about right to me.

PR #121, opened 2017-09-26
PROTON-1368: Remove parser from the public API

ssorj:


astitcher:
I'm not sure what the logic of this change is:

This parser API is not used internally (as far as I can tell) in proton-c so removing the external API whilst leaving the internal code intact probably makes no sense unless you are trying to maintain ABI compatibility. But in this case there is no need to even keep the header file internally as there is no internal code that uses this API.

astitcher:
Note the build failures are from internal code including proton/parser.h.

So even though I didn't find anything internally using the parser maybe my code browser isn't as good as I thought.

astitcher:
looks good.

ssorj:
https://github.com/apache/qpid-proton/commit/ea02b93374b527206eef3a1c7def939eff7804fb

PR #125, opened 2017-10-13
Steps to make openssl.c thread safe, marked with FIXME comments.

alanconway:
This is not a real PR, just for comments.


alanconway:
For discussion only, will be addressed by 
PROTON-1665 |   | Review locking - TLS / SSL thread safety with proactor

PR #128, opened 2017-11-02
PROTON-1342: CI on OS X

RoddieKieley:
This contains the travis yaml updates to configure CI to run osx type os's as well as matrix'd variants for 10.11-10.13 and Xcode 7.3.1 - Xcode 9 respectively. 

IMPORTANT: This alone will simply result in failed builds for os: osx, and requires the work in PROTON-522 for working os: osx builds. PR to follow.

NOTE: This PR also contains an updated README.md that shows the status of the OSX builds on PROTON-522 at the moment.  Please edit as required.

Feedback welcome!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,19 +17,30 @@
 # under the License
 #
 
-os: linux
 sudo: false
 language: cpp
 compiler:
 - gcc
 - clang
-env:
-- QPID_PROTON_CMAKE_ARGS=
-- QPID_PROTON_CMAKE_ARGS=-DCMAKE_BUILD_TYPE=Coverage
+
 matrix:
-  exclude:
-  - compiler: clang
-    env: QPID_PROTON_CMAKE_ARGS=-DCMAKE_BUILD_TYPE=Coverage
+  include:
+  - os: linux
+    env:
+    - QPID_PROTON_CMAKE_ARGS='-DCMAKE_BUILD_TYPE=Coverage'
+
+  - os: osx
+    osx_image: xcode7.3
+    env:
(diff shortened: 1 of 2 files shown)


astitcher:
A few points - 
* I'm not sure how familiar you are with git/github: but there seem to be a lot of fairly meaningless changes here. Could you squash all of these changes into a single change. That would make it easier to see what's actually changed.
* I think you've introduced way too many builds to the CI - do we really need to build every possible version of Xcode? I'd prefer to pick just the current and earliest supported version - that should suffice for CI, shouldn't it?
* I don't think you need to introduce a conditional to the before_script. The purpose of the QPID_PROTON_CMAKE_ARGS variable is to change exactly what you are changing - I think you can set it for the new matrix elements you are introducing.

RoddieKieley:
- git fairly / github not so much, and agreed, will squash.
- Almost certainly do not need every version, but what do you think about the latest version for every major version? So in this case say 7.3.1, 8.3, and 9.0?
- Will see about an update using the QPID_PROTON_CMAKE_ARGS rather than the if blocks and update further.

gemmellr:
I agree on reducing the number of builds. The /apache org has a shared job capacity for all the foundations projects, so we dont want to be overloading things, and tripling the number just to add OSX seems like its getting there. I'd probably do what Andrew said and pick the latest plus something earlier.

RoddieKieley:
@gemmellr didn't know that. That's good incentive to go with earliest and latest in that case - will update with 7.3.1 and 9.

astitcher:
I like this latest version, but 2 remaining things:

* Please, please rebase and squash this to be a single commit - this chain of development makes no sense as part of the project history.
* A niggle really - you didn't follow the existing formatting of the yaml file. Specifically you indent follow on lists, this just pushes everything further right and makes it harder to read long lines. The existing style is like this:
```yaml
before_install:
- eval ...
- if ...
- if ...
```

PR #130, opened 2017-11-17
PROTON-2221 Add coverage reporting for the Ruby binding

jirkadanek:


ssorj:
We'll get this when 0.20.0 opens.

jiridanek:
I'm few years too late to make it for 0.20, but 0.32 is also nice, maybe? ;-P

Nowadays, I'd probably go for https://sonarcloud.io instead of Codecov, but well, that should work too.

PR #131, opened 2017-12-07
PROTON-1711: SSL_IMPL none SASL_IMPL none build attempts to build c-s…

RoddieKieley:
…sl-tests and fails

Added simple check to only build c-ssl-tests if SSL_IMPL was not none.

NOTE: This simple change also exposes the fact that the python binding does not build with SSL_IMPL none and all other configuration variables default at the moment. I disabled the python binding for a successful build as I didn't need it at the moment.

Should be addressed here or via separate issue?


astitcher:
The tests should still pass if SSL_IMPL is none as there is a stub implementation. If not then we need to fix that bug.

astitcher:
Oops - I guess I did that when I added in the ssl tests.

astitcher:
@RoddieKieley  I've fixed this issue now - can you close the PR (stupidly I forgot to add that to the commit).

PR #132, opened 2017-12-22
PROTON-1728: WIP: Reorganize the source tree and remove deprecated bindings

ssorj:
WIP! For review only.

Note that this doesn't attempt any significant clean up of the cmake logic.  I left that for another day.


ChugR:
* This seems like a great improvement overall. It's easier to find stuff without needing the folklore to locate it.
* The presentation of the source tree gives the project a mature look and feel.
* Compiled on windows VS2012 x64 and ran tests OK.

A few observations:
* Why is there a directory src/extra? This seems odd after such a massive cleanup.
* As a downstream consumer of the binary artifacts the minor reorganization is acceptable.

ssorj:
One more change I'd like to consider making is moving the language-specific examples with their binding code.  That's consistent with how we currently handle docs, and it will make the cmake logic a bit more natural.

astitcher:
Can I suggest breaking this down into 3 separate changes (PRs branches etc.):
* Removal of deprecated bindings
* Removal of obsolete doc
* Moving the remainder to new locations and fixing up build

I think it'll be easier to see what's going on in the important part (the third) if there were fewer individual commits to look at. This would be easier I think if you just remove all the stuff from the tree.

In any case the removals are much easier to review/test individually.

PR #133, opened 2017-12-24
PROTON-1796 branch for automated periodic Coverity Scan runs

jdanekrh:
This commit could be added as a new branch. Then, TravisCI can be set up
to build it daily (or weekly).

What it does is that it
1) checks out the current master branch
2) builds it under the Coverity tool
2) submits the build to Coverity Scan

The `COVERITY_SCAN_TOKEN` would need to be replaced with a correct one.

The new branch would need to be created.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,53 @@
+os: linux
+dist: trusty
+sudo: false
+language:
+- c
+- cpp
+compiler:
+- gcc
+addons:
+  apt:
+    packages:
+    - cmake
+    - libssl-dev
+    - libsasl2-dev
+    - sasl2-bin
+    - swig
+    - python-dev
+    - valgrind
+    - ruby
+    - ruby-dev
+    - python3-dev
+    - php5
+    - golang
+    - lcov
+
+env:
+  global:
+    # COVERITY_SCAN_TOKEN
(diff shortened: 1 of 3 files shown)


jdanekrh:
The CI failure is expected. Build in correctly configured branch looks like https://travis-ci.org/msgqe/travisci/builds/321105069

jdanekrh:
@ssorj The coverity PR for daily (weekly) build. This is the same thing that is set up for Artemis, https://github.com/msgqe/travisci/tree/activemq-artemis-coverity.

gemmellr:
I don't think we have ability to configure periodic builds in Travis on this mirror, over which we have limited control in which case this doesn't really make sense.

jdanekrh:
It does not have to be placed on this mirror. The Artemis coverity run is under /msgqe/ organization on github. The only requirement is that the coverity token needs to be uploaded to travis under that specific account.

So, for example, if I was given rights on coverity website to upload new builds for the project, I would be able to set up the coverity build. (That is exactly how it is working for Artemis right now.)

gemmellr:
I meant the PR probably doesnt make sense, i.e we wouldnt have a branch here, which it seems you agree with.

jdanekrh:
I have to agree. Without the ability to have periodic travis build, there won't be a point. Closing the PR, then.

PR #135, opened 2018-01-04
PROTON-1732: [OSX] Enable swig for the Travis CI build

RoddieKieley:
Minor update to enable swig for the OSX Travis CI. BUILD_PERL=OFF for the moment as it attempts to build i386 and fails on OSX.


astitcher:
Hmm, not quite sure about this:
* The actual change seems fine, although ISTR putting in an explicit turning off for PERL for apple builds some years ago, so I'm surprised you needed to do this (but not an actual problem).
* All of the ruby tests are crashing, so maybe turning ruby off until we investigate and fix might be good, we can't have merging this change make the Travis master CI fail.
* Even though cmake does find and use swig, the most important thing - the python tests - still don't build/run because cmake finds a mismatched libpython with a different version from the python interpreter it finds.

So this is a useful step, but not a real advance over where are already are.

RoddieKieley:
@astitcher Thanks for the quick feedback. Will address the 2nd and 3rd points and push changes for review as well as take a look at the BUILD_PERL configuration.

.travis.yml
```diff
@@ -23,29 +23,37 @@ compiler:
...
 
+env:
+- PIP='pip'
+
 matrix:
   include:
   - os: linux
     env:
     - QPID_PROTON_CMAKE_ARGS='-DCMAKE_BUILD_TYPE=Coverage'
+    - PIP='pip'
```
astitcher: You don't need this as it is the same as the default above.

RoddieKieley: I didn't have it first and the non matrix'd builds didn't succeed; see [buildlog](https://travis-ci.org/RoddieKieley/qpid-proton/jobs/325304860). Was there something I missed maybe?

astitcher: I think we may be talking about the opposite case:
- I'm suggesting that this line which part of a specific matrix build is not necessary because there is a line in the non matrix case.
- I think you are saying that not having the non matrix line doesn't work. I agree with this.

RoddieKieley:
No luck with having the global PIP env var utilized by the linux os matrix [build](https://travis-ci.org/RoddieKieley/qpid-proton/jobs/327947850). Was there something I missed?

astitcher:
Maybe I'm the one who misunderstands how env and matrix interact then.

Only one other thing I might suggest is to just use pip2 everywhere without a variable at all - It looks like that should work on Ubuntu and will avoid bringing attention to this irrelevant detail.

astitcher:
But failing that I'm happy with what's there now.

PR #137, opened 2018-03-01
[cpp] Fix update method missing 'name' variable

p32blo:
The `name` property in the sender and receiver is not being updated correctly when passed to the `open_sender` and `open_receiver`.

Here is a simple example demonstrating the problem:
```cpp
#include <proton/sender_options.hpp>
#include <proton/receiver_options.hpp>
#include <proton/container.hpp>

#include <iostream>

int main()
{
    proton::container c;

    proton::sender_options so;
    proton::receiver_options ro;

    so.name("qpid");
    ro.name("qpid");

    proton::sender sender = c.open_sender("", so);
    proton::receiver receiver = c.open_receiver("", ro);

    std::cout << sender.name() << std::endl;
    std::cout << receiver.name() << std::endl;
}
```
The output is currently a random uuid, but it should be "qpid" in both cases.

### DIFF
--- a/proton-c/bindings/cpp/src/receiver_options.cpp
+++ b/proton-c/bindings/cpp/src/receiver_options.cpp
@@ -101,6 +101,7 @@ class receiver_options::impl {
         dynamic_address.update(x.dynamic_address);
         source.update(x.source);
         target.update(x.target);
+        name.update(x.name);
     }
 
 };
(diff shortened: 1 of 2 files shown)


astitcher:
Thanks for this fix - I will commit it when we have some tests for this bug to prevent any regressions.

If you want to get this into master quicker then I suggest you write a simple test for this bug which fails without this fix and succeeds with the fix.

It looks like there isn't an obvious existing unit test file that this would fit into sadly, so it might mean creating a new one - mostly by copying an existing one.

PR #139, opened 2018-03-07
PROTON-1787: Pinned the homebrew python version to 2 to stay current with

RoddieKieley:
current change to default of python to python 3 in homebrew.
See associated travis job [140](https://travis-ci.org/RoddieKieley/qpid-proton/builds/350289134)


astitcher:
This seems like an entirely unnecessary change - the Travis build job already only uses python2 - why do we need to change it?
Homebrew building has no relation with Travis building does it?

PR #140, opened 2018-03-08
PROTON-1638, PROTON-1728: Reorganize the source tree

ssorj:
I plan to squash this on merge.


gemmellr:
I gave things a skim over, changes seemed reasonable from my doesnt-do-cmake-etc perspective.

Noting for later reference: this builds on / incorporates WIP from #136 and #138.

PR #141, opened 2018-04-11
PROTON-2215: Windows build fails if CMAKE_MODULE_PATH is not empty

matlo607:


jiridanek:
This looks correct to me. I'd feel still better about this after going through a list of other CMake variables, if there is something that allows specifying a shorter path. CMAKE_MODULE_PATH is a list of directories, not a single directory, so the build file as it is now is certainly wrong, just because of the type mismatch. I'll create a Jira.

jiridanek:
There does appear to be a way to specify shorter path, by using [`CMAKE_CURRENT_LIST_DIR`](https://cmake.org/cmake/help/latest/variable/CMAKE_CURRENT_LIST_DIR.html) or [`CMAKE_CURRENT_FUNCTION_LIST_DIR`](https://cmake.org/cmake/help/latest/variable/CMAKE_CURRENT_FUNCTION_LIST_DIR.html). The first of those is available in CMake 2.8.12.

jiridanek:
This script will hopefully go away in Proton 0.32 thanks to #245.

PR #142, opened 2018-04-25
PROTON-1831: Incorrect open/close sequence for same link name.

alanconway:
Fixes the server side part of PROTON-1831 - transport error if a duplicate link
name is attached instead of seg fault.

Client side problem remains: pni_process does not respect the order of individual open/close calls,
and it is possible to generate an invalid sequence that attempts a duplicate
attach when it should not. This commit includes a test that illustrates the
problem, the test runs but status is ignored for now.

### DIFF
--- a/c/src/core/transport.c
+++ b/c/src/core/transport.c
@@ -1253,7 +1253,7 @@ int pn_do_begin(pn_transport_t *transport, uint8_t frame_type, uint16_t channel,
   return 0;
 }
 
-pn_link_t *pn_find_link(pn_session_t *ssn, pn_bytes_t name, bool is_sender)
+static pn_link_t *pni_find_link(pn_session_t *ssn, pn_bytes_t name, bool is_sender)
 {
   pn_endpoint_type_t type = is_sender ? SENDER : RECEIVER;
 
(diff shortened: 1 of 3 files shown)


gemmellr:
The use of "invalid-field" for the error feels like it is maybe a little off. Its possible/likely there is nothing wrong with the information in the fields of the attach, as none of it is really actually checked (e.g whether its the same remote handle, itself a session:handle-in-use error) other than someone got there previously with the name. Perhaps "illegal-state" might be more applicable given the actual check being done?

That said though, its not actually 100% clear it is an illegal state. The spec covers attaching an active link a second time as resulting in the first attachment being detached with the "stolen" link error. Its described for handling across different connections (something proton thus cant actually do on its own) e.g during reconnect, but its not clear that precludes it being the case on a single session or connection. However, it does seem far more likely to be in result of an error for the situation to ever occur on a single session/connection so perhaps treating it as such is the way to go in the end (plus any alternative would I guess be much harder to achieve, so maybe its the only choice for now).

alanconway:
On Fri, Apr 27, 2018 at 4:57 AM, Robbie Gemmell <notifications@github.com>
wrote:

> The use of "invalid-field" for the error feels like it is maybe a little
> off. Its possible/likely there is nothing wrong with the information in the
> fields of the attach, as none of it is really actually checked (e.g whether
> its the same remote handle, itself a session:handle-in-use error) other
> than someone got there previously with the name. Perhaps "illegal-state"
> might be more applicable given the actual check being done?
>
> That said though, its not actually 100% clear it is an illegal state. The
> spec covers attaching an active link a second time as resulting in the
> first attachment being detached with the "stolen" link error. Its described
> for handling across different connections (something proton thus cant
> actually do on its own) e.g during reconnect, but its not clear that
> precludes it being the case on a single session or connection. However, it
> does seem far more likely to be in result of an error for the situation to
> ever occur on a single session/connection so perhaps treating it as such is
> the way to go in the end (plus any alternative would I guess be much harder
> to achieve, so maybe its the only choice for now).
>
Thanks - I didn't know about stolen. I think we could implement what the
spec says on the server side - close the old link/handle with "stolen" and
create a new link under the new handle. That would fix the crash, which was
because of multiple handles to the same  link object - we'd have a seprate
link object for each handle, all but one of which are closed with "stolen".

I think on the client side its always an error to try to attach if you
already have a link with same name an a handle. After disconnect all
handles are cleared, so it is OK an error to re-attach during reconnect.  I
want to fix the client to refuse to create a link at all in this case. That
would prevent this happening on the wire and give more immediate feedback
to code that is mistakenly trying to double-attach. The fix is not trivial
due to proton's batch-processing madness or I'd have done it first.

Does that sound right?
Cheers,
Alan.

gemmellr:
It seems you already committed this before Gordon and I commented, but neglected to close the PR with the commit or otherwise. It either needs closing, or the commit reverted, depending what your intent is after the comments.

alanconway:
The commit makes things better than they were but is not the final fix, as per comments from you and Gordon. Closing the PR as the JIRA has the comments - I think they are valid, just haven't got back to it yet.

alanconway:
Merged, see JIRA

PR #143, opened 2018-05-15
NO-JIRA: [c] link fuzzing binaries using CXX linker

jdanekrh:
This should allow for removal of patch file https://github.com/google/oss-fuzz/blob/1909d92b8bb068dd7aa94bdf22da4028ada385ea/projects/qpid-proton/c_tests_fuzz_CMakeLists.patch


jdanekrh:
out of date, see https://github.com/apache/qpid-proton/pull/172

PR #144, opened 2018-05-18
Python proactor

astitcher:
This is a fairly large change to the structure of the Proton python binding.
* Initially it removes support for really old versions of python - anything before 2.5.
* Then it goes on to split up the monolithic __init__.py file into IMO) more manageable pieces.

I've also used this to restrict the usable API to only those bits that are supposed to be API (again IMO).

This is the initial work which I'm using as a base for more substantial changes, so I will be grateful for any comments and review.


astitcher:
Pushed a new version of these changes according to feedback from @ssorj

PR #145, opened 2018-05-24
Potential chained epoll impl for PROTON-1842

cliffjansen:
Removes temp fix, adds chained/secondary epollfd_2.

Still has failure in helgrind threaderciser racecheck.  Still unsure if a problem or false positive.

Previous temp fix also has the same threaderciser error, so at least not a regression.

I have not been able to detect any change in performance with my various epoll proactor load tests.

Instrumented runs show the secondary/chained arming occurs about 0.1% in proton ctest.  It is closer to 0.001% in dispatch router runs under heavy load, but that may be less indicative of the overall frequency.

Chaining is unlikely to occur if socket output does not fill the kernel buffer (!EWOULDBLOCK), or if a write event mask is desired as a result of input (say a flow event) and EPOLLIN is also necessary with the EPOLLOUT.

Chaining *is* likely to occur if the socket is quiet (i.e. no recent output and waiting on EPOLLIN), and the pconnection gets a pn_connection_wake to do enough output to fill the kernel output buffer.

### DIFF
--- a/c/src/proactor/epoll.c
+++ b/c/src/proactor/epoll.c
@@ -280,11 +280,6 @@ const char *AMQP_PORT_NAME = "amqp";
 // and increases latency.
 #define HOG_MAX 1
 
-// pn_proactor_t.deferred_free_list, pconnection_t.free_in_progress, pconnection_deferred_free()
-// are part of a temporary mitigation strategy for PROTON-1842 pending a comprehensive solution.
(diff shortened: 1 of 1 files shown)


alanconway:
I think this makes sense but it makes my head hurt a bit.

cliffjansen:
Substantially adopted but with similar valgrind cleanups as in 37cd347.

PR #146, opened 2018-05-31
NO-JIRA: [c] prefer linking with static library in fuzz tests

jdanekrh:
This should allow for removal of patch file https://github.com/google/oss-fuzz/blob/1909d92b8bb068dd7aa94bdf22da4028ada385ea/projects/qpid-proton/c_CMakeLists.patch


astitcher:
I'm  a little confused by this change - what creates the libqpid-proton-static lib that this changes uses? It doesn't seem that this changes creates that library anywhere.

astitcher:
Ok - I see this is dependant on @alanconway 's recent work.

jdanekrh:
out of date, see https://github.com/apache/qpid-proton/pull/172

PR #150, opened 2018-07-09
PROTON-1888: allow configuration of connection details from a config …

grs:
…file

The file to use can be specified using the CONNECT_CONFIG environment variable. If
that is not set, it will look for a file name connect.json, first in the current
directory, then in the users local configuration directory (~/.config/messaging)
and then in the system config directory (/etc/messaging/)

The config file is in json format and the following properties can be specified:

scheme
host
port
sasl_mechanisms
user
password
cert_file
private_key_file
private_key_password
ca_cert_file


grs:
Rebased and updated to align with commit for PROTON-1935 for c++

PR #152, opened 2018-08-03
PROTON-1886: Dump the thread's SSL error queue in the event that SSL_new fails.

k-wall:
Uses the established mechanism to dump the thread's SSL error queue to established tracer in the event that SSL_new fails.


astitcher:
This looks good - I'll pull it in.

I wonder though if there is scope for some refactoring of the ssl error logic to make this error logic in common.

PR #156, opened 2018-09-17
PROTON-1936: lower case include and updated ifdefs

marcelmeulemans:


astitcher:
You don't explain what problem you are solving with these changes. Please give more details about why you need to -
1. change the case of the includes - I thought windows filesystems are case insensitive.
2. Add some platform detection as well as compiler detection - I think that in openssl.c the detection should probably be platform detection instead of compiler detection. But in the other cases it platform detection makes no sense to me as things being selected are compiler things not platform things.

marcelmeulemans:
Sorry, you are right, I could have been a bit more verbose in explaining the issue being tackled here (also in the Jira issue) and how I solve them.

What this pull request is trying to achieve is to allow proton to be cross compiled for windows from a linux host using mingw. The are basically two main things in the current code base preventing this (both of which you are asking questions about):

1. **File name casing**. The windows API is not really very consistent when is comes to the casing of file names (can differ per VS version, or platform), so as windows file systems are not case sensitive and linux file systems are, inconsistent casing causes a problem on linux (i.e. `fatal error: Windows.h: No such file or directory # include <BaseTsd.h>`). Mingw "overcomes" this by lower casing all include and library files. So one part of this pull request lower cases all includes and libraries. As windows is case insensitive, this changes nothing on windows but allows compilation to succeed on linux.
2. **Preprocessor based platform detection** In several places in the proton code base platform detection is being done using preprocessor definitions. In a few of these cases it is doing so by using a define that identifies the compiler, not the platform. In these places I have added a check for a define that identifies the _platform_ (which as you mention should be the correct way of doing so). I did not feel confident enough to remove the existing compiler identification check, but I would be happy to do so.

astitcher:
>
>     1. **File name casing**. The windows API is not really very consistent when is comes to the casing of file names (can differ per VS version, or platform), so as windows file systems are not case sensitive and linux file systems are, inconsistent casing causes a problem on linux (i.e. `fatal error: Windows.h: No such file or directory # include <BaseTsd.h>`). Mingw "overcomes" this by lower casing all include and library files. So one part of this pull request lower cases all includes and libraries. As windows is case insensitive, this changes nothing on windows but allows compilation to succeed on linux.

I'm completely fine with these changes.

>
>     2. **Preprocessor based platform detection** In several places in the proton code base platform detection is being done using preprocessor definitions. In a few of these cases it is doing so by using a define that identifies the compiler, not the platform. In these places I have added a check for a define that identifies the _platform_ (which as you mention should be the correct way of doing so). I did not feel confident enough to remove the existing compiler identification check, but I would be happy to do so.

A few points here:
* An aside really, but I would say that if you are using Windows you'd be better to use the schannel TLS implementation rather than openssl. But having said that the test in openssl.c should definitely be a platform test not a compiler test.
* However I'm not at all clear whether the tests around snprintf, vsnprintf, va_copy are actually platform dependent.
  * snprintf, vsnprintf are C runtime library things, so they could either come from the platform or the compiler. I had assumed they came from the compiler in which case they don't need any new selection logic, but if they come from the platform then the selection login should be changed. 
  * va_copy is definitely compiler dependent and not platform dependent although I guess it might depend on the ABI in use.

Did you find that you needed both these changes to make the code compile or work?

astitcher:
Having said all that on the whole if the Windows CI build works and the Linux CI build is unaffected I'm ok with merging this change. I'd just prefer to make the platform/compiler selection logic as simple and clear as possible, without explanationless voodoo present.

c/src/platform/platform.h
```diff
@@ -81,7 +81,7 @@ int pni_vsnprintf(char *buf, size_t count, const char *fmt, va_list ap);
 
 #endif
 
-#if defined _MSC_VER || defined _OPENVMS
+#if defined _MSC_VER || defined(_WIN32) || defined _OPENVMS
```
jiridanek, line 84: I am surprised this is needed. I always believed mingw is a version of GCC, and therefore there wouldn't be a version of mingw that does not define the va_copy macro.

It would be nice to switch to Visual Studio 2015+, and stop supporting 2013. Then there would be va_copy always defined even on MSVC.

jiridanek: Note to myself: The switch I mentioned in previous comment already happened, so va_copy on Windows can be relied on.

PR #157, opened 2018-09-19
WIP: extract connection_options_impl for tests and internal use

alanconway:


astitcher:
Basically don't like this approach because it exposes implementation details to the library interface.

As the unit tests don't only need to be built in the build environment. Myself I'd be happy with direct linking, but I'm not clear why you need/want to test the implementation when the actual external interface pretty much directly passes through anyway.

Perhaps you can give a simple example of a test that can't use the exposed API.

alanconway:
I want to unit test connection_config::parse by parsing various configs into connection_options, and then verifying that the connection_options receives the correct values. connection_options public API is set-only, no gets. I considered adding public get methods, if you think that's a good idea too I'll do that.

Otherwise the only thing you can do with connaction_options is apply them to a connection. The connection, transport, sasl and ssl objects also have mostly set-only interfaces, so that doesn't help.

I could write test to verify the *effects* of the values by making a connection, but I would much rather test parser as a unit, independently of tests for SASL/TLS functionality. Currently there are no tests  for SASL or TLS except the examples, I don't think connection parser tests are the right place to add them.

astitcher:
Honestly given the alternative, adding gets for the options seems fairly innocuous!

To be (a little) less flippant Unit testing is actually a good reason to make the options readable too. Although I'm not sure if this is a good idea for password though.

I think the original reason for not making the options readable was that it was unecessary as the client program should know what options it set, but unit testing is important too.

Perhaps an alternative would be a "dump" method for the options (akin to ```pn_inspect()``` or __str__) which outputs the options in json for comparison although you'd have to parse it and compare to be safe. But hey you've got a json parser linked in anyway!

PR #159, opened 2018-09-24
PROTON-1940: [c] normalize encoding of multiple="true" fields

alanconway:
AMQP spec allows several ways to encode a "multiple" field, we allow all in
incoming data or pn_data_t values created in the user, but when writing
to the wire we always encode:

- empty array as null
- single value array as a single value
- multiple values - encoded as array

This is the most compact encoding and seems to be best for interop, as some AMQP
clients (.NET) have trouble with an empty array in a multiple value field.

### DIFF
--- a/c/src/core/codec.c
+++ b/c/src/core/codec.c
@@ -481,6 +481,84 @@ static int pni_data_intern_node(pn_data_t *data, pni_node_t *node)
   return 0;
 }
 
+
+/*
+   Append src to data after normalizing for "multiple" field encoding.
+
+   AMQP composite field definitions can be declared "multiple", see:
+
+   - http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-types-v1.0-os.html#doc-idp115568
+   - http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-types-v1.0-os.html#section-composite-type-representation
+
+   Multiple fields allow redundant encoding of two cases:
+
+   1. empty: null or an empty array.
+   2. single-value: direct encoding of value, or array with one element
+
+   For encoding compactness and inter-operability, normalize multiple
+   field values to always use null for empty, and direct encoding for
+   single value.
+*/
+static int pni_normalize_multiple(pn_data_t *data, pn_data_t *src) {
+  int err = 0;
+  pn_handle_t point = pn_data_point(src);
+  pn_data_rewind(src);
+  pn_data_next(src);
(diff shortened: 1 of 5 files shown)


c/src/core/codec.c
```diff
@@ -657,7 +657,46 @@ int pn_data_vfill(pn_data_t *data, const char *fmt, va_list ap)
         }
       }
       break;
-    default:
+     case 'M': {                /* multiple=true field value */
+       /* Encode a pn_data_t * as a normalized multiple="true" field.
+          See file:///home/aconway/specs/amqp-core-complete-v1.0.pdf
```
astitcher: What is 'multiple="true" field'? There is nothing called this in the standard is there? The ="true' naming seems to indicate boolean fields, but this is not the case afaict - and this really applies to a list of things that can also be empty in which case we just use null instead of the empty list.

Also url to local file system is not very helpful!

c/src/core/codec.c
```diff
@@ -529,15 +529,15 @@ int pn_data_vfill(pn_data_t *data, const char *fmt, va_list ap)
     case 'd':
       err = pn_data_put_double(data, va_arg(ap, double));
       break;
-    case 'Z':
+    case 'Z':                   /* encode binary, must not be NULL */
```
astitcher: As much as I agree the format letters need documenting I think that doing it inline is *not* the right way to do it - IMO a single block comment above would be much easier to encompass all the different codes.

alanconway:
Changes prompted by astitcher's comments:

- better explanation of mulitple fields, corrected AMQP spec links
- added block comment for vfill format letters
- factored out pni_normalize_multiple to reduce visual churn in _vfill() impl

Not changed: still using a new format letter. 2 reasons - consistency and efficiency/ease of impl:

consistency: the multiple attribute is part of an AMQP composite field definition. The purpose of _vfill() is to fill in composite fields correctly - so IMO that is where this belongs. I personally hate the vfill char-based mini-language; it is hard to read, hard to remember and error-prone. However it seems like something we should replace entirely or live with for now, rather than have inconsistent approaches for different fields.

ease of impl: as part of vfill, we can copy exactly as much of the source data as we need to the target. A normalize function outside of vfill would either have to modify the source value in-place (side-effects on caller's data value) or create an intermediate copy (expensive, extra memory management) Both are inconsistent with existing use of vfill.

alanconway:
Merged, if there are still objections to using a vfill code lets open a new issue

PR #160, opened 2018-09-28
WIP: option<T> version of connection_options getters

alanconway:
This is the template approach to option getters - memory safe.
Needs to be extended to all the _options classes, but I want agreement on the approach and option<T> clss before I complete the job. @astitcher @cliffjansen

### DIFF
--- a/cpp/include/proton/connection_options.hpp
+++ b/cpp/include/proton/connection_options.hpp
@@ -27,6 +27,7 @@
 #include "./internal/config.hpp"
 #include "./internal/export.hpp"
 #include "./internal/pn_unique_ptr.hpp"
+#include "./option.hpp"
 #include "./symbol.hpp"
 #include "./types_fwd.hpp"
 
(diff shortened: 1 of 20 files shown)


astitcher:
I definitely like making option<T> a class in common rather than duplicating it.

I think that this is the correct way to expose options in the API, but perhaps we could make the API a bit more easy to use. Perhaps we could have a C++11 only explicit bool conversion as well as is_set(). However we would need to experiment a  bit as some of the options themselves are also bools so this might be confusing!
Another thing we should consider is throwing an exception if get() is called when the option is not set.

alanconway:
> Perhaps we could have a C++11 only explicit bool conversion as well as is_set(). However we would need to experiment a bit as some of the options themselves are also bools so this might be confusing!

I left it off because of the confusion. I could probably be persuaded but I'm not sure of the value - I think 
    if (opt.is_set()) use(opt.get())
is clearer and not really harder to use than
    if (opt) use(opt.get())

> Another thing we should consider is throwing an exception if get() is called when the option is not set.
No objection, I'll make it so. Technically it is a logic_error. Do you think  we should add proton::logic_error, or fudge it and just throw proton::error?

alanconway:
Here's the final version, I'll hold off till after release to take a bit of time to consider API impact. @astitcher @ssorj

alanconway:

PR #161, opened 2018-10-03
NO-JIRA: Removed python upgrade for xcode7.3 travis build to fix pip issue

RoddieKieley:
Tried a few different updates to rectify the [travis xcode7.3 build failure](https://travis-ci.org/apache/qpid-proton/jobs/436640885) with this having the least impact. Originally we needed this to get a working python version but now we don't and the upgrade to 3.7.0 breaks pip for the moment.


astitcher:
This looks good to me - it makes me wonder if we could do the same for the Xcode 9 build too.

astitcher:
Another thought - is there any way to pin the python version using brew? That would make the CI build instructions more robust.

PR #168, opened 2018-11-15
PROTON-1887: [c] Convert C tests to use Catch2 harness.

alanconway:
Convert c/tests from home-baked test macros to the Catch2 test harness.
We are using version 1.x until we can move on from supporting C++03.
https://github.com/catchorg/Catch2/tree/Catch1.x

Added files:

test/include/
- catch.hpp - the Catch2 version 1.x framework in a single header.
- catch_extra.hpp - added Catch::Matcher support for C-string comparisons

c/tests/
- pn_test.hpp/.cpp - tools for testing the core library
  - auto_free<> - scoped freeing of pn_xxx_t*
  - matchers to check pn_condition/error_t with useful output
  - driver class to pump a pn_connection_driver_t and check events
    - trivial in-memory copy transport, no networking needed.
- pn_test_proactor.hpp/cpp - tools for testing the proactor library
  - proactor to pump a pn_proactor and check events.
  - same handler as driver above.

Renamed tests files: c/tests/*.c => c/tests/*_test.cpp

Tests are now linked into 3 executables:

- c-core-test - tests for libqpid-proton-core API
- c-extra-test - tests for the "extra" deprecated API in libqpid-proton
- c-proactor-test - tests for the proactor API in libqpid-proton-proactor

Individual tests or subsets can be run via command line arguments, run with
"-h" for a summary, or see https://github.com/catchorg/Catch2/blob/Catch1.x/docs/command-line.md

NOTE: Catch2 allows multiple "tags" to be associated with each test to easily
define different test sets. We're not using this feature yet, we could use it to
identify a fast self-test set, a long-running stress test set etc.


astitcher:
This has been merged

PR #169, opened 2018-11-27
Clang format

alanconway:
Format command: clang-format -i $(git ls-files -- *.cpp *.hpp *.c *.h)

Style file is /.clang-format, with the following contents:

Language: Cpp                   # also applies to C
BasedOnStyle: LLVM
IndentWidth:     4
AllowShortIfStatementsOnASingleLine: true
AllowShortLoopsOnASingleLine: true

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,5 @@
+Language: Cpp                   # also applies to C
+BasedOnStyle: LLVM
+IndentWidth:     4
+AllowShortIfStatementsOnASingleLine: true
+AllowShortLoopsOnASingleLine: true
--- a/c/examples/broker.c
+++ b/c/examples/broker.c
@@ -31,449 +31,474 @@
 #include <stdlib.h>
 #include <string.h>
 
-/* The ssl-certs subdir must be in the current directory for an ssl-enabled broker */
+/* The ssl-certs subdir must be in the current directory for an ssl-enabled
+ * broker */
 #define SSL_FILE(NAME) "ssl-certs/" NAME
 #define SSL_PW "tserverpw"
 /* Windows vs. OpenSSL certificates */
 #if defined(_WIN32)
-#  define CERTIFICATE(NAME) SSL_FILE(NAME "-certificate.p12")
-#  define SET_CREDENTIALS(DOMAIN, NAME)                                 \
-  pn_ssl_domain_set_credentials(DOMAIN, SSL_FILE(NAME "-full.p12"), "", SSL_PW)
+#define CERTIFICATE(NAME) SSL_FILE(NAME "-certificate.p12")
+#define SET_CREDENTIALS(DOMAIN, NAME)                                          \
+    pn_ssl_domain_set_credentials(DOMAIN, SSL_FILE(NAME "-full.p12"), "",      \
+                                  SSL_PW)
 #else
-#  define CERTIFICATE(NAME) SSL_FILE(NAME "-certificate.pem")
-#  define SET_CREDENTIALS(DOMAIN, NAME)                                 \
-  pn_ssl_domain_set_credentials(DOMAIN, CERTIFICATE(NAME), SSL_FILE(NAME "-private-key.pem"), SSL_PW)
+#define CERTIFICATE(NAME) SSL_FILE(NAME "-certificate.pem")
+#define SET_CREDENTIALS(DOMAIN, NAME)                                          \
+    pn_ssl_domain_set_credentials(DOMAIN, CERTIFICATE(NAME),                   \
+                                  SSL_FILE(NAME "-private-key.pem"), SSL_PW)
 #endif
 
 /* Simple re-sizable vector that acts as a queue */
-#define VEC(T) struct { T* data; size_t len, cap; }
+#define VEC(T)                                                                 \
+    struct {                                                                   \
+        T *data;                                                               \
+        size_t len, cap;                                                       \
+    }
 
-#define VEC_INIT(V)                             \
-  do {                                          \
-    void **vp = (void**)&V.data;                \
-    V.len = 0;                                  \
-    V.cap = 16;                                 \
-    *vp = malloc(V.cap * sizeof(*V.data));      \
-  } while(0)
+#define VEC_INIT(V)                                                            \
+    do {                                                                       \
+        void **vp = (void **)&V.data;                                          \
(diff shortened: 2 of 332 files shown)


astitcher:
__I 100% support having a consistent automated format for the proton codebase(s)__

However , I don't think this is quite it (IMO):
- I think we should just leave the proton-c codebase with an indent of 2 - A lot of changes for marginal gain there.
  - We can have a 4 space indent for the C++ code, I don't think having slightly different settings is a problem if the format is automatically applied.
- I think we can allow wider lines - it looks like it is wrapping at 80 chars, I think 100 would be fine (conservative even) for modern screens.
- I'm not happy with the reformatting of the header files:
  - Especially reformatting the 2nd and subsequent parameter at a variable indent - not cool for variable width fonts!.
  - Seems to have reformatted blah* x(); to blah *x(); which is fine for C code, but not for C++ code. Another area where the setting should probably be different.
  - Connected to the first sub point: if the function name and first parameter is too long wrapping both:
blah *pn_xxxx(pn_param1 *x);
To
blah *
pn_xxxx(pn_param1 *x);

I think the mess of the header files is actually the most important issue given that this is user visible documentation. It should look as nice as we can make it. Consistency is important, but it shouldn't be ugly to view - I currently opine that the reformat has made the header files ugly.

dnwe:
When I've applied clang-format to existing projects, I knocked up a bit of shell script to iterate over the repos with a matrix of various BasedOnStyle and some of the override params and recording the git diff -w --shortstat each time. I then chose whichever config settings caused the smallest diffstat delta going forward.

alanconway:
On Wed, Nov 28, 2018 at 12:29 PM Andrew Stitcher <notifications@github.com>
wrote:

> *I 100% support having a consistent automated format for the proton
> codebase(s)*
>
> However , I don't think this is quite it (IMO):
>
Have a bash: `clang-format -style=X -dump-style`  will give you all the
gory details, also check the clang site for the list of pre-defined styles
(Mozilla and others), you may find one more to your liking.

>
>    - I think we should just leave the proton-c codebase with an indent of
>    2 - A lot of changes for marginal gain there.
>       - We can have a 4 space indent for the C++ code, I don't think
>       having slightly different settings is a problem if the format is
>       automatically applied.
>
> Easy to do, just put separate .clang-format files in /c and /cpp . I'd
prefer a common style for both however. The first big reformat will be a
pain (mitigated by automated reformatting for outstanding commits) but the
pain will pass.
If we use different styles for C and C++, should C++ files in the C tests
tree be formated like C or like the C++ binding? The latter will make
automated reformatting more complicated.

>
>    - I think we can allow wider lines - it looks like it is wrapping at
>    80 chars, I think 100 would be fine (conservative even) for modern screens.
>
> +1, I went with  minimal changes from LLVM but 100 is fine, I find 80 a
bit tight.


>
>    - I'm not happy with the reformatting of the header files:
>       - Especially reformatting the 2nd and subsequent parameter at a
>       variable indent - not cool for variable width fonts!.
>
> No space-based alignment method works for variable width fonts. Are you
proposing we move to tab-based indentation, or to a style that never aligns
anything with the previous line? I'm pretty sure clang-format can do that,
just not sure if that's what you mean.

>
>    -
>       - Seems to have reformatted blah* x(); to blah *x(); which is fine
>       for C code, but not for C++ code. Another area where the setting should
>       probably be different.
>
> Depends who you ask, lots of C++ style guides use blah *x; as it more
accurately reflects the precendence of *.
For example `blah *x, *y` means what it looks like; `blah* x, y` does not
(y is a blah, not  a blah*)
If you want this I believe the clang-format variable is `PointerAlignment:
Left`. Again I'd prefer we be consistent in C and C++.

> I think the mess of the header files is actually the most important issue
> given that this is user visible documentation. It should look as nice as we
> can make it. Consistency is important, but it shouldn't be ugly to view - I
> currently opine that the reformat has made the header files ugly.
>
I went with adopting an existing style guide with minimal changes, but no
objection if you can find or craft a more beautiful one.

alanconway:

jiridanek:
+1, I'd like to have a reasonable `.clang-format` in the project. It is nowadays _the_ way to specify C/C++ code formatting across editing tools (together with `.editorconfig`, but that is simpler tooling that cannot take language semantics into account). It would be very useful when writing a new file, so that I know what is the projects desired code formatting.

I probably wouldn't start by mass reformatting of the entire codebase. That can come later. (That is, for C/C++. I probably would want to reformat all Python right away.)

Good thing about autoformatters is that after codebase reformat, they can be also applied to any work-in-progress code in branches, and that will hopefully significantly decrease the effort to adopt the style there, when merging.

It would be good if other Qpid projects (i.e. Dispatch) could adopt the same style. I do not have particular preference on what the style should be.

alanconway:
Last time we never managed to converge on a style. I suggest you attach a proposed .clang-format and announce a time limit (a few weeks?) Anyone with objections MUST attach a new .clang-format that implements a style they would be happy with; objections without a new implementation are ignored. That should converge pretty quickly.

jiridanek:
I messed with it for a while, and I found out I am not the correct person to propose a code style for the project. So, let me propose to adopt whatever dispatch ends up choosing, in https://github.com/apache/qpid-dispatch/pull/744/files#diff-88d5907f11b8b6ca0b213de568bf4582

Currently the proposal there is

```
BasedOnStyle: Google
IndentWidth: 4
ColumnLimit: 120
AllowShortFunctionsOnASingleLine: false
SpaceAfterCStyleCast: true
DerivePointerAlignment: false
PointerAlignment: Left
AlignConsecutiveAssignments: true
AlignConsecutiveDeclarations: true
KeepEmptyLinesAtTheStartOfBlocks: false
```

On the other hand, my favorite settings include something like

```
BasedOnStyle: LLVM
ExperimentalAutoDetectBinPacking: true
ColumnLimit: 100
AccessModifierOffset: -4
AllowShortIfStatementsOnASingleLine: Always
AlignConsecutiveDeclarations: true
FixNamespaceComments: true
NamespaceIndentation: None
DerivePointerAlignment: false
PointerAlignment: Right
SpaceAfterCStyleCast: true
```

alanconway:
Being consistent with dispatch is a good idea. Good luck getting agreement :)

alanconway:
Closing as I can't follow up on this, someone else can take it over if there's still interest.

PR #175, opened 2019-02-05
src/ssl/openssl: add libressl compatibility

rc-matthew-l-weber:
Similar to https://github.com/FreeRDP/FreeRDP/issues/5049
libressl has `#define OPENSSL_VERSION_NUMBER ` defined the same as
openssl 1.1.x which results in SSL_CTX_set_security_level() getting used.

This patch prevents SSL_CTX_<get/set>_security_level() from being used with
libressl.

Signed-off-by: Matthew Weber <matthew.weber@rockwellcollins.com>

### DIFF
--- a/c/src/ssl/openssl.c
+++ b/c/src/ssl/openssl.c
@@ -522,7 +522,7 @@ pn_ssl_domain_t *pn_ssl_domain( pn_ssl_mode_t mode )
   // Mitigate the CRIME vulnerability
   SSL_CTX_set_options(domain->ctx, SSL_OP_NO_COMPRESSION);
 #endif
-#if OPENSSL_VERSION_NUMBER >= 0x10100000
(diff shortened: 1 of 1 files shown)


gemmellr:
I'll defer review/merge to folks familiar with these bits, but can you please raise a JIRA issue at https://issues.apache.org/jira/browse/PROTON and reference it in your commit message (see previous commits for example).

rc-matthew-l-weber:
https://issues.apache.org/jira/browse/PROTON-2004

astitcher:
This change is missing the definition of default_seclevel at line 78 - this is only used when SSL_CTX_get_security_level and SSL_CTX_set_security_level are present so should also be excluded in this case.

```
#if OPENSSL_VERSION_NUMBER >= 0x10100000
  int default_seclevel;
#endif

```

astitcher:
This PR should now be resolved with commit SHA 87898b3d71aea5bfd9e4157d6da04071e1461339

PR #176, opened 2019-02-17
NO-JIRA: [c] Fix Coverity warning of buffer overrun in pn_proactor_addr

jirkadanek:
```
Error: OVERRUN (CWE-119):
qpid-proton-0.27.0/c/tests/proactor_test.cpp:613: overrun-buffer-arg: Overrunning array "sl" of 1024 bytes by passing it to a function which accesses it at byte offset 1024 using argument "1024UL".
#  611|
#  612|     pn_transport_t *st = pn_connection_transport(s);
#  613|->   pn_netaddr_str(pn_transport_local_addr(st), sl, sizeof(sl));
#  614|     CHECK_THAT(cr, Equals(sl)); /* client remote == server local */
```

### DIFF
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -505,7 +505,7 @@ endif(MSVC)
 
 if (qpid-proton-proactor)
   set(HAS_PROACTOR True)
-  add_library (qpid-proton-proactor SHARED ${qpid-proton-proactor})
+  add_library (qpid-proton-proactor SHARED ${qpid-proton-proactor} ${qpid-proton-platform})
   target_link_libraries (qpid-proton-proactor  LINK_PUBLIC qpid-proton-core)
   target_link_libraries (qpid-proton-proactor  LINK_PRIVATE ${PLATFORM_LIBS} ${PROACTOR_LIBS})
   set_target_properties (qpid-proton-proactor
@@ -515,7 +515,7 @@ if (qpid-proton-proactor)
     LINK_FLAGS "${CATCH_UNDEFINED} ${LTO}"
   )
   if (BUILD_STATIC_LIBS)
-    add_library (qpid-proton-proactor-static STATIC ${qpid-proton-proactor})
+    add_library (qpid-proton-proactor-static STATIC ${qpid-proton-proactor} ${qpid-proton-platform})
   endif(BUILD_STATIC_LIBS)
 endif()
 
--- a/c/src/proactor/proactor-internal.c
+++ b/c/src/proactor/proactor-internal.c
@@ -21,6 +21,9 @@
 /* Common platform-independent implementation for proactor libraries */
 
 #include "proactor-internal.h"
+
+#include "platform/platform.h"
+
 #include <proton/error.h>
 #include <proton/listener.h>
 #include <proton/netaddr.h>
@@ -39,14 +42,7 @@ static const char *AMQPS_PORT_NAME = "amqps";
 const char *PNI_IO_CONDITION = "proton:io";
 
 int pn_proactor_addr(char *buf, size_t len, const char *host, const char *port) {
-  /* Don't use snprintf, Windows is not C99 compliant and snprintf is broken. */
-  if (buf && len > 0) {
-    buf[0] = '\0';
-    if (host) strncat(buf, host, len);
-    strncat(buf, ":", len);
-    if (port) strncat(buf, port, len);
-  }
-  return (host ? strlen(host) : 0) + (port ? strlen(port) : 0) + 1;
+  return pni_snprintf(buf, len, "%s:%s", host ? host : "", port ? port : "");
 }
 
 int pni_parse_addr(const char *addr, char *buf, size_t len, const char **host, const char **port)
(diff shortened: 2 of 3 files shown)


astitcher:
I don't think this is the correct fix. There is no problem using strncat that requires inventing a new (and possibly wrong) alternative that does the same thing but perhaps a bit different.

In any case as far as I can see there isn't a buffer overrun problem with the fixed code at all! pn_proactor_addr can only overrun the buffer it's given if you give it the wrong length.

astitcher:
Thinking more...
> I don't think this is the correct fix. There is no problem using strncat that requires inventing a new (and possibly wrong) alternative that does the same thing but perhaps a bit different.
> 

This is still my view.

Rereading the code and man pages the code is incorrect, but introducing a new strncat isn't necessary. Just reduce the len parameter to strncat by the size of the copied host and ':' chars to avoid the overrun. You then only need the do strlen on the host.

Also doing strncat(host); strncat(':'); strncat(port) is wrong. The output buffer should start with strncpy() as we don't want to append to anything already in the buffer but overwrite it (as documented in the header file)

> In any case as far as I can see there isn't a buffer overrun problem with the fixed code at all! pn_proactor_addr can only overrun the buffer it's given if you give it the wrong length.

This is just wrong - sorry!

astitcher:
How about something like (compiles and passes tests with valgrind, but not tested with sanitisers or coverity):

```
int pn_proactor_addr(char *buf, size_t len, const char *host, const char *port) {
  /* Don't use snprintf, Windows is not C99 compliant and snprintf is broken. */
  size_t hostlen = host ? strlen(host) : 0;
  size_t portlen = port ? strlen(port) : 0;
  if (buf && len > 0) {
    if (host) {
        strncpy(buf, host, len);
    } else {
        buf[0] = '\0';
    }
    if (hostlen+2 < len) {
        buf[hostlen] = ':';
        buf[hostlen+1] = '\0';

        if (port) strncat(buf, port, len-hostlen-1);
    }
    buf[len-1] = '\0';
  }
  return hostlen + portlen + 1;
}
```

jdanekrh:
> Also doing strncat(host); strncat(':'); strncat(port) is wrong. The output buffer should start with strncpy() as we don't want to append to anything already in the buffer but overwrite it (as documented in the header file)

`strncat` should be perfectly fine. First thing the code does (both Alan's version and mine) is `buf[0] = '\0';`. strncat will be then overwriting whatever was in buf, because it sees it as empty string to start with.

> How about something like (compiles and passes tests with valgrind, but not tested with sanitisers or coverity):

Looks reasonable, and does not require new function...

`if (port) strncat(buf, port, len-hostlen-1);` I am thinking maybe `-2`, because strncat always adds `\0`. I'd need to run this in valgrind myself...

I don't know (yet) if proton has tests that exercise this logic, or if all hostnames are relatively short compared to size of that buffer.

jdanekrh:
I am debating whether to update the PR with Andrew's fix, or whether to propose requiring MSVC 2015 for Proton and simply use `snprintf` after the change is made. I'm inclined to go with snprintf, as it is straightforward and lasting.

jdanekrh:
There is `pni_snprintf`, which on Windows behaves as compatibility wrapper! I did not know about this before.

I am updating the PR to use that.

c/src/platform/platform.h
```diff
@@ -65,7 +65,7 @@ int pn_i_error_from_errno(pn_error_t *error, const char *msg);
  */
 int64_t pn_i_atoll(const char* num);
 
-int pni_snprintf(char *buf, size_t count, const char *fmt, ...);
+PN_EXTERN int pni_snprintf(char *buf, size_t count, const char *fmt, ...);
```
astitcher: Can't do this pni_ prefix means internal only not to be exported.
If you need to use this in a different library you need to copy the code.

jdanekrh: I saw two `pni_` symbols in qpid-proton-cored.lib already, so I did it this way. Ok, I'll link it directly.

    dumpbin /EXPORTS qpid-proton\out\build\x64-Debug\c\qpid-proton-cored.lib

    [...]

                  pn_void_refcount
                  pn_work_head
                  pn_work_next
                  pni_message_get_extra
                  pni_message_with_extra
                  pnx_sasl_clear_password
                  pnx_sasl_error
                  pnx_sasl_fail_authentication
                  pnx_sasl_get_allow_insecure_mechs
                  pnx_sasl_get_auth_required
                  pnx_sasl_get_context
                  pnx_sasl_get_external_ssf
    [...]

astitcher: Yes, those symbols are the subject of an issue (PROTON-1637) they should either be public API with a well defined meaning or not exported.
pni == ProtoN Internal.

jdanekrh: I've made the change. Is this what you've meant? Just realized I'd have to do the same thing to `qpid-proton-proactor-static` if I wanted to go this way.

jiridanek:
This is better fixed by #245.

gemmellr:
Which also has a much-needed JIRA ;)

jiridanek:
#245 is not enough, need also https://github.com/apache/qpid-proton/pull/267

PR #179, opened 2019-02-21
NO-JIRA: [c] missing break in send-ssl.c example

jdanekrh:
CID 214944 (#1 of 1): Missing break in switch (MISSING_BREAK)unterminated_case: The case for value
    PN_CONNECTION_REMOTE_OPEN is not terminated by a 'break' statement.

This looks genuinely wrong. Either there is `break` missing, or there should be some `/* fallthrough */` comment to signal this is intentional.


astitcher:
I agree it's wrong (although actually harmless) . It's fairly harmless because the event will have no link and pn_link_credit will return 0 if it is called with no link - although maybe it should just error!

PR #181, opened 2019-05-02
PROTON-2040 [cpp] Allow connection options to be updated for reconnect

alanconway:
Added connection::reconnect_update(const connection_options&) which can be
called during on_transport_error() to change the failover list and other
connection options before the next automatic reconnect.

### DIFF
--- a/cpp/include/proton/connection.hpp
+++ b/cpp/include/proton/connection.hpp
@@ -181,6 +181,10 @@ PN_CPP_CLASS_EXTERN connection : public internal::object<pn_connection_t>, publi
     /// @see reconnect_options, messaging_handler
     PN_CPP_EXTERN bool reconnected() const;
 
+    /// **Unsettled API** - update the connection options prior to an automatic reconnect.
+    /// Note this will over-write existing options like connection_options::update()
+    PN_CPP_EXTERN void reconnect_update(const connection_options&);
+
     /// @cond INTERNAL
   friend class internal::factory<connection>;
   friend class container;
--- a/cpp/src/connection.cpp
+++ b/cpp/src/connection.cpp
@@ -56,7 +56,7 @@ void connection::open() {
 }
 
 void connection::open(const connection_options &opts) {
-    opts.apply_unbound(*this);
+    if (uninitialized()) opts.apply_unbound(*this);
     pn_connection_open(pn_object());
 }
 
@@ -197,5 +197,10 @@ bool connection::reconnected() const {
     reconnect_context* rc = cc.reconnect_context_.get();
     return (rc && rc->reconnected_);
 }
+
+void connection::reconnect_update(const connection_options &opts) {
+    opts.apply_unbound(*this);
+}
+
 
 } // namespace proton
--- a/cpp/src/connection_options.cpp
+++ b/cpp/src/connection_options.cpp
@@ -79,11 +79,6 @@ class connection_options::impl {
      */
     void apply_unbound(connection& c) {
         pn_connection_t *pnc = unwrap(c);
-
-        // Only apply connection options if uninit.
-        bool uninit = c.uninitialized();
-        if (!uninit) return;
-
         if (reconnect.set)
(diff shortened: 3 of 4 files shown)


astitcher:
I don't think I like this design much.

If I understand it correctly when you get to the reconnect you can't modify the connection options but only overwrite them completely.

This means that you have to keep around the original connection options and then update them with any changes. This is IMO inconvenient, but worse isn't what the API seems to offer with the name `reconnect_update()` - even if the comment is correct.

I think it would it be better for the connection to keep around the connection options used and then you could really update them. If we implement it this way we could offer a more general `connection::connection_options()` accessor and allow that to be updated with the existing `connection_options::update()` API and not need the `reconnect_update()` API which is strangely specifically named for something that actually just overwrites the connection options and is not really related to reconnect (except by intention).

In any event, I think that in order to sensibly offer this functionality someone has to remember the connection options so it might as well be the binding.

alanconway:
On Fri, May 3, 2019 at 11:20 AM Andrew Stitcher <notifications@github.com>
wrote:

> I don't think I like this design much.
>
> If I understand it correctly when you get to the reconnect you can't
> modify the connection options but only overwrite them completely.
>
Nope the new options are applied over the existing ones in the manner of
connection_options::update(), there's no explicit call to update but the
application of the options is additive.


> This means that you have to keep around the original connection options
> and then update them with any changes. This is IMO inconvenient, but worse
> isn't what the API seems to offer with the name reconnect_update() - even
> if the comment is correct.
>
> I think it would it be better for the connection to keep around the
> connection options used and then you could really update them. If we
> implement it this way we could offer a more general
> connection::connection_options() accessor and allow that to be updated
> with the existing connection_options::update() API and not need the
> reconnect_update() API which is strangely specifically named for
> something that actually just overwrites the connection options and is not
> really related to reconnect (except by intention).
>
That is effectively what happens - I'm not using update() explicitly but
apply_unbound() effectively does the same thing - I believe, if I'm wrong
then your approach would be better.

> In any event, I think that in order to sensibly offer this functionality
> someone has to remember the connection options so it might as well be the
> binding.
>
It's the pn_connection_t objects that effectively "remember" the options,
along with some of the internal C++ context objects. Probably I should add
a test to prove this.

astitcher:
> It's the pn_connection_t objects that effectively "remember" the options, along with some of the internal C++ context objects. Probably I should add a test to prove this.

I don't think this is the case exactly. Looking at the reconnect code (which I just did to remind myself!) I see that we actually now do keep araound a copy of the `connection_options` for a connection in the `connection_context`. This is there because at reconnect time you actually have to get the failover urls and parameters from the connection options as they aren't part of the underlying proton-c model at all. 

Also looking at the reconnect code I see that at reconnect time it is the *cached away* version of the options that is applied to the new transport of the new connection attempt - I don't think your changes will affect the cached away version of the connection_options - have you tested them?

So I'm not convinced your changes even work - unless you've got a test to prove that they do!

In summary I think that my suggestion actually is feasible - provide an accessor to the connections options - this would allow you to update them and then they would get used on the next connection attempt.

astitcher:
Another separate point - although this change is small, its affect is not. It absolutely needs some sort of test before it can be merged.

alanconway:
Updated PR with tests.

gemmellr:
The tests hang in one of my build envs. I'm guessing this could be because my ISP, like several others, have their DNS servers attempt to be 'helpful' with unresolvable addresses and so actually do resolve to a landing page. I expect the test isnt getting the exceptions its looking for as a result.

astitcher:
@gemmellr That might make the test hang, but then so would several others because they also rely on the same trick to elicit a transport error.

astitcher:
See #182 For a different implementation of the same approach.

gemmellr:
In leaving it, the test does eventually pass, it just takes 8 minutes. The total suite normally takes 30-45sec.

alanconway:
Closing stale PRs - reopen if this is still live.

PR #182, opened 2019-05-07
PROTON-2040: Alternative suggestion for updating connection options:

astitcher:
- Renamed connection method - update_options()
  * This is the actual function of the method - to update any connection
    options. so that updating eg SSL/SASL options will take affect too.
- Corrected implementation
  * The previous implementation did not actually update the stored
    connection options, rather applying them immediately. This did
    accidentally update the reconnection options and the connection
    based options, but not the transport based ones.

### DIFF
--- a/c/src/core/connection_driver.c
+++ b/c/src/core/connection_driver.c
@@ -77,12 +77,13 @@ int pn_connection_driver_bind(pn_connection_driver_t *d) {
 
 pn_connection_t *pn_connection_driver_release_connection(pn_connection_driver_t *d) {
   if (d->transport) {           /* Make sure transport is closed and unbound */
-      pn_connection_driver_close(d);
-      pn_transport_unbind(d->transport);
+    pn_connection_driver_close(d);
+    pn_transport_unbind(d->transport);
   }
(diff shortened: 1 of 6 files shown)


astitcher:
@alanconway This is basically the same idea as your PR #181 

Implementing this I noticed that there are a couple of issues with this approach:
* There is no way to replace/change the initial connection url.
   For full control of reconnect we definitely need to be able to control this. maybe we need to treat it as a connection option. So that we can replace it by updating the options.
* When the initial connection url gets used is a little obscure. Your version and mine actually have different behaviours here which are caused by the slightly different approach to updating the reconnect context. But it is a little unintuitive to me that you get the initial url used again after failing then updating the failover url list.
   If we had a way to change the initial connection url then this might be a moot point.

alanconway:
This looks good to me.
On initial URL I noticed that also. We could make  URL a connection_option and provide a connect() that takes only options - I think we do that in some other bindings like ruby.
If you use a connect(url, opts) call, then I would say the URL from connect() should over-write any URL in the options to avoid confusion.

alanconway:

PR #183, opened 2019-05-13
PROTON-2047: Multithreaded examples for C++0x

ssorj:


astitcher:
**The naming here is wrong**
Your new examples are for C++0**3**.
[There is no such thing as C++0x, when C++11 was being developed it was referred to by that name, as it was assumed it would be finished in that decade! This naming style has pretty much died out now as the standards committee has very effectively managed to time box the last few releases to 3 year cycles - C++11, C++14, C++17 and coming soonish C++20]

astitcher:
I'll rephrase that - as it's not exactly to point.
Your code is not for *any* standardised version of C++ before C++11. It will only work for the gcc 4.8 compiler with the -std=c++0x flag *or* an actual C++11 compiler.
Given this I'm not actually sure it makes that much sense as an example.

ssorj:
I moved these examples to a different location, outside of the Proton source tree:

https://github.com/ssorj/menagerie/tree/master/qpid-proton-cpp-multithreading-el6

PR #184, opened 2019-07-18
PROTON-2076 compile c-threaderciser.c without test_main.cpp

jdanekrh:


jdanekrh:
Already fixed in PROTON-2120.

PR #185, opened 2019-07-23
PROTON-2079: Fix usage of formatting placeholders in printfs

jdanekrh:


jdanekrh:
Aha, I am actually supposed to use the definitions from c/src/platform/platform_fmt.h for the placeholders.

c/tools/msgr-send.c
```diff
@@ -289,7 +289,7 @@ int main(int argc, char** argv)
         check_messenger(messenger);
     }
 
-    LOG("Messages received=%llu sent=%llu\n", received, sent);
+    LOG("Messages received=%" PRIu64 " sent=%" PRIu64 "\n", received, sent);
```
astitcher: I don't think these C99 macros exist in MSVC before VS2013

jdanekrh: They don't, but it's in "platform/platform.fmt"

https://github.com/apache/qpid-proton/blob/97b62d2b7ce22c84e7d8db1ace8aa0355ea5fbe7/c/src/platform/platform_fmt.h#L61-L65

astitcher: True, but is that header included? It's internal to proton and (I think) this is code that only uses the exposed proton API (albeit the old messenger API.

astitcher: Ah - there are defines in c/tools/msgr-common.h which are included.

jdanekrh: Seems I was more lucky than knowledgeable, then.

astitcher:
+1 Feel free to rebase and merge this.

PR #187, opened 2019-08-23
PROTON-2185 cpp/CMakeLists.txt: fix qpid-proton-cpp linking

ffontaine:
gpid-proton-cpp must use LINK_PUBLIC to link with qpid-proton-core and
qpid-proton-proactor otherwise build will fail if those libraries are
not yet installed on the system:

[ 81%] Linking CXX executable value_test
/home/fabrice/buildroot/output/host/lib/gcc/arm-buildroot-linux-uclibcgnueabihf/8.3.0/../../../../arm-buildroot-linux-uclibcgnueabihf/bin/ld: warning: libqpid-proton-proactor.so.1, needed by libqpid-proton-cpp.so.12.6.0, not found (try using -rpath or -rpath-link)
/home/fabrice/buildroot/output/host/lib/gcc/arm-buildroot-linux-uclibcgnueabihf/8.3.0/../../../../arm-buildroot-linux-uclibcgnueabihf/bin/ld: warning: libqpid-proton-core.so.10, needed by libqpid-proton-cpp.so.12.6.0, not found (try using -rpath or -rpath-link)
/home/fabrice/buildroot/output/host/lib/gcc/arm-buildroot-linux-uclibcgnueabihf/8.3.0/../../../../arm-buildroot-linux-uclibcgnueabihf/bin/ld: libqpid-proton-cpp.so.12.6.0: undefined reference to `pn_data_put_described'

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -179,7 +179,7 @@ if(BUILD_STATIC_LIBS)
   add_library(qpid-proton-cpp-static STATIC ${qpid-proton-cpp-source})
 endif(BUILD_STATIC_LIBS)
 
-target_link_libraries (qpid-proton-cpp LINK_PRIVATE ${PLATFORM_LIBS} qpid-proton-core qpid-proton-proactor ${CONNECT_CONFIG_LIBS})
+target_link_libraries (qpid-proton-cpp LINK_PRIVATE ${PLATFORM_LIBS} ${CONNECT_CONFIG_LIBS} LINK_PUBLIC qpid-proton-core qpid-proton-proactor)
 
 set_target_properties (
   qpid-proton-cpp


astitcher:
Can you explain your specific situation more because this fix doesn't entirely make sense:
These libraries should be LINK_PRIVATE as far as cmake semantics are concerned because they provide symbols only to the qpid-proton-cpp library itself and are not needed by any of it's clients.
LINK_PUBLIC means that any client of qpid-proton-cpp would need to explicitly link with these libraries and this is not needed or desired.
Are you perhaps trying to build the cpp directory of the qpid-proton tree by itself? Currently the only way to build the C++ API is as part of the full tree build.

franzhollerer:
I found that this commit fixes a problem I had with cross compiling qpid-proton-cpp.

http://qpid.2158936.n2.nabble.com/Cross-Compiling-qpid-proton-0-30-0-with-Yocto-SDK-td7689081.html

Alternatively, adding 
```
link_directories(${PN_C_LIBRARY_DIR})
```
to cpp/CMakeLists.txt solved too.

I am unsure which one is the better fix.

@astitcher wrote:
> Are you perhaps trying to build the cpp directory of the qpid-proton tree by itself? Currently the only way to build the C++ API is as part of the full tree build.

I think @ffontaine  faced a similar problem than I. I tried to build the C++ API with a cross compiler as part of the full tree build. The build failed because the libqpid-proton-proactor.so was not found during the linking step.

Building qpid-proton for the host succeeded. Thus it may be some incompatibility with the toolchain file generated provided by Yocto. This fix solved this issue.

ffontaine:
It should be noted that this patch is not needed if tests are disabled: https://issues.apache.org/jira/browse/PROTON-2170

PR #188, opened 2019-08-23
PROTON-2091: fix `len(queue)` in python/examples and reformat for PEP8

jdanekrh:


astitcher:
I think you left extraneous testing code in the bottom of the file.
Otherwise looks good to me.

jdanekrh:
I meant to keep the test code there. It will only execute if user does `python -m unittest discover`, otherwise it is inert.

astitcher:
Please remove it - this is meant to be example code. No other example has test code included.

If you really think we need to include unit tests for something that you think is complex enough to need it then use a testing tool that can extract unit tests from comments and put the unit test comment with the class/function that is being tested.

jdanekrh:
Ok, it is trivial enough it does not need a test. Adding this as a python doctest would be similarly distracting, I guess

PR #190, opened 2019-09-25
PROTON-2109: [Python] Correct default JSON schema to amqps

rkubis:
According to the [documentation](https://github.com/apache/qpid-proton/blob/master/docs/connect-config.md), default port should be `amqps`.

### DIFF
--- a/python/proton/_reactor.py
+++ b/python/proton/_reactor.py
@@ -1214,7 +1214,7 @@ class Container(Reactor):
         """
         if not url and not urls and not address:
             config = _get_default_config()
-            scheme = config.get('scheme', 'amqp')
+            scheme = config.get('scheme', 'amqps')
             _url = "%s://%s:%s" % (scheme, config.get('host', 'localhost'), config.get('port', _get_default_port_for_scheme(scheme)))
             _ssl_domain = None
             _kwargs = kwargs
--- a/python/tests/proton_tests/connect.py
+++ b/python/tests/proton_tests/connect.py
@@ -32,7 +32,7 @@ def write_connect_conf(obj):
         json.dump(obj, outfile)
 
 class Server(MessagingHandler):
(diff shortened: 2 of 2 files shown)


astitcher:
I'm not sure if this is the best thing or to change the documentation to match the implementation.
As far as I remember the default should be 'amqp' not 'amqps' that is what the other json connect config readers use as default.
@ssorj - Do you have insight here?

ssorj:
I worked from the written spec and docs from Alan, which has "amqps" as the default.

astitcher:
I just checked and the C++ implementation uses amqps as default so I will merge this PR.

Thanks @rkubis .

astitcher:
@rkubis - In order to keep the git log good would you raise a JIRA for this change and then repush this change with a commit log like:
PROTON-2xxx: [Python] correct default json schema to amqps

rkubis:
> @rkubis - In order to keep the git log good would you raise a JIRA for this change and then repush this change with a commit log like:
> PROTON-2xxx: [Python] correct default json schema to amqps

Yes, no problem, I will raise a JIRA and then repush the change.

rkubis:
JIRA https://issues.apache.org/jira/browse/PROTON-2109 raised and the change re-pushed.

astitcher:
@rkubis Sorry I just realised this change makes the python tests hang. You need to go into the python tests for the connect.json file and fix them up to carry on working. look in python/tests/proton_tests/connect.py

rkubis:
@astitcher The Python tests are fixed now.

I was not sure if the fix of the Python tests should be in a separate commit or not, I squashed the Python tests fix with the Python library fix into one commit.

jdanekrh:
@rkubis afaik same commit. Clebert was once proposing test fixes in separate commits, in Artemis project, but then changed his mind soon.

PR #191, opened 2019-09-26
PROTON-2110: [Python] Expand tilde symbol as user's home directory wh…

rkubis:
…en searching for connect.json file


rkubis:
Yes, I thought about use of the `os.path.expanduser(...)` in the loop, but I see that there is just one location with the tilde symbol in the location list now, so I decided to use one call of the `os.path.expanduser(...)` only to decrease the number of calls of this function.

astitcher:
Ok, fair enough - I'd have decided the other way, but we can change it easily enough.

PR #193, opened 2019-09-27
NO-JIRA: add .asf.yaml

jdanekrh:
> When making use of a feature in .asf.yaml, please ensure that you have discussed this with the entire project first, and have understood what the configuration changes will do to your workflow and project resources.
> ~~ https://cwiki.apache.org/confluence/display/INFRA/.asf.yaml+features+for+git+repositories

I believe that the github config part (setting tagline and project labels) is mostly harmless.

I'll create mailing list thread about choosing tagline and the labels. Initial pull request does not change the current tagline, adds link to project website, and adds lots of labels collected from other similar GitHub repositories.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -1,6 +1,6 @@
 github:
   description: "Mirror of Apache Qpid Proton"
-  homepage: https://qpid.apache.org/proton/index.html
+  homepage: https://qpid.apache.org/proton
   labels:
     - qpid
     - amqp
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -1,0 +1,23 @@
+github:
+  description: "Mirror of Apache Qpid Proton"
+  homepage: https://qpid.apache.org/proton/index.html
+  labels:
+    - qpid
+    - amqp
+    - messaging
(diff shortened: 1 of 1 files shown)


gemmellr:
Drop the index.html from the URL, its shorter and more readable.

I'd question how useful some of the labels are, but I guess they don't do any harm either, and the currently populated ones (not sure they were always there) are clearly not applicable in some cases.

.asf.yaml
```diff
@@ -0,0 +1,23 @@
+github:
+  description: "Mirror of Apache Qpid Proton"
+  homepage: https://qpid.apache.org/proton/index.html
```
jdanekrh:
```suggestion
  homepage: https://qpid.apache.org/proton
```

gemmellr:
Since its just updating the tags, and many of the existing tags are entirely wrong thanks to being taken from a common source, I've just applied this regardless of the discussion thread. It helps folks see what it actually looks like applied also.

I added a licence header to avoid updating e.g the RAT config or anything else, though I see form the other PR you already noticed that.

PR #196, opened 2019-10-21
PROTON-2122: Increase hardcoded limit from 16 to 32 SASL mechanisms

jdanekrh:
Temporary measure before the issue is fully addressed. This allows more qpid-dispatch tests to pass on macOS.


RoddieKieley:
@jdanekrh I double checked the 2 appveyor checks that had issues and from previous discussions surrounding that failure it is known unrelated.

PR #197, opened 2019-10-21
PROTON-2123 Escape every char >127 printed into logs

jdanekrh:
As mentioned on the DISPATCH issue, I don't know what exactly is causing me problems, and I don't have standalone reproducer, but this does fix the issue for me and the patch does not look completely unreasonable on its own.

### DIFF
--- a/c/src/core/util.c
+++ b/c/src/core/util.c
@@ -38,7 +38,8 @@ ssize_t pn_quote_data(char *dst, size_t capacity, const char *src, size_t size)
   for (unsigned i = 0; i < size; i++)
   {
     uint8_t c = src[i];
-    if (c < 128 && isprint(c)) {
+    // output printable ASCII, ensure '\' always introduces hex escape
+    if (c < 128 && c != '\\' && isprint(c)) {
       if (idx < (int) (capacity - 1)) {
         dst[idx++] = c;
       } else {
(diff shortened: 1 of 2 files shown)


jdanekrh:
When this change is considered from the perspective of ensuring that logs are always valid utf-8 encoded text, then I think this is a valid fix. Irrespective of what `isprint` thinks, every character outside of ASCII range will be escaped, which means that the result will always be valid utf-8. (Since ASCII is subset of utf-8.)

c/src/core/util.c
```diff
@@ -38,7 +38,7 @@ ssize_t pn_quote_data(char *dst, size_t capacity, const char *src, size_t size)
   for (unsigned i = 0; i < size; i++)
   {
     uint8_t c = src[i];
-    if (isprint(c)) {
+    if (c < 128 && isprint(c)) {
```
gemmellr: The bit I expect is directly comparable from proton-j uses:
`if (c > 31 && c < 127 && c != '\\')`

The < 127 and < 128 difference isnt important since DEL (127) isnt printable, but I'd note that proton-j also excludes \ (92) from direct inclusion since it could clash with the \x prefix used for the alternative hex-format output case.

jdanekrh: Yes, it makes lots of sense to make this reversible!

ChugR:
Ship it.

PR #199, opened 2019-10-24
PROTON-2124 Disable GS2-KRB5 SASL mechanism if it is not explicitly enabled

jdanekrh:


jdanekrh:
I have cyrus-sasl as a single package on my system, the .so file is in `cyrus-sasl-2.1.27/lib/sasl2/libgs2.so`. Same on macOS, there MacPorts installs it all. (I guess it can be somehow reconfigured when installing package from source.)

Looking in Fedora, it should be this one, https://apps.fedoraproject.org/packages/cyrus-sasl-gs2, or on RHEL, this one https://access.redhat.com/downloads/content/cyrus-sasl-gs2/2.1.27-1.el8/x86_64/fd431d51/package

On RHEL 7 it is in rhel-7-server-optional-rpms, on RHEL 8 in rhel-8-for-x86_64-baseos-rpms.

astitcher:
When I install cyrus-sasl-gs2 I also get the mechanism GS2-IAKERB - did you not disable this mechanism for a specific reason? Or did you not see it offered?
I suggest preemptively disabling this mechanism too.

jdanekrh:
I saw it, it is mentioned on https://issues.apache.org/jira/browse/PROTON-2122 and also PROTON-2124 in comment by @RoddieKieley. It was not causing any problems for me with the tests or examples, so I decided to ignore it for the time being.

I guess I could disable it as well.

PR #200, opened 2019-10-25
PROTON-2126 Fix Appveyor CI build

jdanekrh:
Previously, the build would fail when installing setuptools and tox:

```
pip install --user setuptools tox
Traceback (most recent call last):
  File "c:\python27\lib\runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "c:\python27\lib\runpy.py", line 72, in _run_code
    exec code in run_globals
  File "C:\Python27\Scripts\pip.exe\__main__.py", line 9, in <module>
TypeError: 'module' object is not callable
Command exited with code 1
```

* Use pip systemwide, not with --user flag
* Use explicit Appveyor image versions in build matrix
* Install VC 9.0, because Python 2.7 needs it

https://www.appveyor.com/docs/windows-images-software/

https://wiki.python.org/moin/WindowsCompilers#Microsoft_Visual_C.2B-.2B-_9.0_standalone:_Visual_C.2B-.2B-_Compiler_for_Python_2.7_.28x86.2C_x64.29

* Don't try detecting sasl on Windows, because MSBuild would fail the py wheel build
* Build also with Python 3.7 x64 to check that is possible
* Fix bad indent in setup.py.in
* Use os.makedirs in setup.py.in, to avoid failue:

```
Error [WinError 3]
	The system cannot find the path specified: 'build\\temp.win-amd64-3.7\\Release\\srcs'
	py_pkg_wheel
	C:\Users\Administrator\qpid-proton\build\python\CUSTOMBUILD	1
```

* Install wheel with pip

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -2,22 +2,44 @@ version: '{branch}.{build}'
 configuration: RelWithDebInfo
 environment:
   matrix:
+    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
+      CMAKE_GENERATOR: Visual Studio 16 2019
+      PYTHON: "C:\\Python37-x64"
+      VCPKG_INTEGRATION: '-DCMAKE_TOOLCHAIN_FILE=C:/Tools/vcpkg/scripts/buildsystems/vcpkg.cmake -A x64'
+      VCPKG_DEFAULT_TRIPLET: x64-windows
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
-      CMAKE_GENERATOR: Visual Studio 15
+      CMAKE_GENERATOR: Visual Studio 15 2017
+      PYTHON: "C:\\Python27"
       VCPKG_INTEGRATION: '-DCMAKE_TOOLCHAIN_FILE=C:/Tools/vcpkg/scripts/buildsystems/vcpkg.cmake'
-    - CMAKE_GENERATOR: Visual Studio 12
-    - CMAKE_GENERATOR: Visual Studio 10
+      VCPKG_DEFAULT_TRIPLET: x86-windows
+    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
+      CMAKE_GENERATOR: Visual Studio 14 2015
+      PYTHON: "C:\\Python27"
+    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
+      CMAKE_GENERATOR: Visual Studio 12 2013
+      PYTHON: "C:\\Python27"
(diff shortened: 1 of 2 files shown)


jdanekrh:
It might be possible to delete the `Visual Studio 12 2013` and `Visual Studio 11 2012` `CMAKE_GENERATOR`s to speed up the CI, as per https://issues.apache.org/jira/browse/PROTON-2077 Support Visual Studio 2015 and up.

astitcher:
@jdanekrh I've merged a slightly simplified version of this PR. Thank you for figuring out the root of these failures.

gemmellr:
For reference, the simplified variant noted seems to be via commits https://github.com/apache/qpid-proton/commit/4a69c9ce2653ece427182f83c113091924eec9a2 and https://github.com/apache/qpid-proton/commit/5dae7dd21c61dd08fafbc0aab0fefa00b12f0a94 (which unlike this PR, look to make the Appveyor build only test with Python 3.5, so maybe it isnt the title of PROTON 2096 that is wrong, but rather the changes on it incomplete given the Travis build is using 2.7? @astitcher?).

jdanekrh:
What I thought was important in this PR was to run pip using `"%PYTHON% -m pip install`, because that ensures the pip used corresponds to version of python being used. Appveyor VMs have 5+ versions of python installed at the same time, 32bit and 64bit variants. Not doing this was causing the original appveyor issue. That got merged.

When you install the wheel package, or when you use python venv which automatically brings in wheel package, then the setup.py needs also updating. And this got merged also.

I also thought that it is a good idea to drop the --user flag from pip invocation, since 1) test runs on throwaway VM 2) appveyor files I examined did not bother with it 3) one less flag to worry about.

astitcher:
For reference I choose python 3.5 because it is present on every appveyor image and so avoided messing with a matrix of different versions.
I left the --user flag not because it's needed here, but because it allows the build logic to be common across different CI platforms even if the actual 'code' is different - ie it will work on a non privileged account as well on Windows or Unix.
Also note that the python binding (rather than the python build code) is tested using tox (if present) and this will test py26, py27, py35, py36, py37 & py38 (I think that's the correct list) if it can find them regardless of the python used for the build.
I removed the VS9 install part because once you move to python3 for the build it's not needed (and doesn't fail the tox test) and is not important for this kind of CI run IMO.

gemmellr:
The appveyor build is no longer testing 2.x as it they aren't working under tox according to the output.

jdanekrh:
@gemmellr That's only on `APPVEYOR_BUILD_WORKER_IMAGE=Visual Studio 2017`. There you'd need to add the VS9 install part. Tox with python 2.7 runs fine on 2015 images.

gemmellr:
ah, oops :)

PR #203, opened 2019-11-07
PROTON-2132 Skip compatibility hacks with libc++

jdanekrh:


astitcher:
Actually I'm pretty sure that the "MacOS" and definitely the FreeBSD tests are actuallly trying to test for libc++ - I just couldn't find the correct way to do it - so well done.
So this means that the test can be simplified to just libc++ and Visual studio.

jdanekrh:
> `#if !defined(_LIBCPP_VERSION) && !defined(_MSC_VER)`

Works on macOS.

PR #209, opened 2019-11-20
PROTON-2138: properly parse the remote terminus properties

kgiusti:
The coordinator terminus only has a capabilities field.  This is
different from the tranditional endpoint attach frame and must be
handled appropriately.


astitcher:
I don't currently have the protocol knowledge to fully review this change.
I'd feel a lot more comfortable with it if there was a a test added that would fail with the current code and succeed with this change. (Probably in the python tests)

PR #210, opened 2019-11-21
PROTON-2137: Removing ssl init from ssl_server_options default constructor

rabih-mourad:


rabih-mourad:
Weird, I just renamed the commit and travis-ci failed...

cpp/src/connection_options.cpp
```diff
@@ -175,7 +175,8 @@ class connection_options::impl {
             }
         } else if (!client && ssl_server_options.set) {
             pn_ssl_t *ssl = pn_ssl(pnt);
-            if (pn_ssl_init(ssl, ssl_server_options.value.impl_->pn_domain(), NULL)) {
+            pn_ssl_domain_t* ssl_domain = ssl_server_options.value.impl_ ? ssl_server_options.value.impl_->pn_domain() : NULL;
+            if (pn_ssl_init(ssl, ssl_domain, NULL)) {
```
astitcher: This change is wrong unfortunately.
the comparison with the client case is misleading. `pn_ssl_init(ssl, NULL, NULL)` will create an ssl connection with the default domain - however the default domain is a *client* domain so you can't do this if the connection is a server.
So you need to replace he NULL with something like `pn_ssl_domain(PN_SSL_MODE_SERVER)`.

astitcher: It is a shame that this wasn't picked up by any tests - I thought we had some ssl server tests.

rabih-mourad: @astitcher  I replaced NULL by pn_ssl_domain(PN_SSL_MODE_SERVER) like you mentioned

astitcher:
@rabih-mourad Could you squash this into a single commit please - then I can rebase and merge it.

rabih-mourad:
@astitcher squash done. Will we add a test in this pull request or we will do another one for it?

astitcher:
I'm happy to commit this as is. I think testing c++ ssl servers is probably a different issue.

jdanekrh:
That travis failure was unrelated to the PR. Something changed either on Travis or in Brew (brew.sh) so that openssl .dynlib is no longer found when `PKG_CONFIG_PATH=/usr/local/opt/openssl/lib/pkgconfig` is set.

I think I know what to do about it.

PR #212, opened 2019-11-21
QPID-8383: Lazy creation of various link related objects

grs:
to reduce per-link memory overhead


astitcher:
@grs I think you may have entered the JIRA into the wrong issue database - it should be PROTON-xxxx.

grs:
@astitcher sorry!

PR #214, opened 2019-11-22
Get AF_INET adresses only in unit tests

ali25hadi:


astitcher:
This is just incorrect - You need AF_UNSPEC to get both ipv4 and ipv6 addresses.

PR #218, opened 2019-11-27
PROTON-2138: add test to verify coordinator target capability parsing

kgiusti:


astitcher:
Does this test fail for the code pre change? If so then just commit it - it looks fine.

kgiusti:
Yes - this test does fail pre-fix.  Checking in...

PR #220, opened 2019-11-28
PROTON-2115 fix typo in tests/py/test_unittest.py breaking fd-limit test

jdanekrh:


jdanekrh:
Nothing, I think. Only fd-limit test uses this. I forgot to commit the other change that needs to be done there for it all to work. I'll try another commit and be more careful.

PR #227, opened 2020-01-22
PROTON-2172 Permit send to exit with ecode 1, as well as hang and then exit with 0

jirkadanek:
Previous version of the test, see git history, was ok with ecode 1.


jiridanek:
Obsoleted by commit https://github.com/apache/qpid-proton/commit/b496997

PR #230, opened 2020-02-27
PROTON-2184: fix assert in pn_session_set_context

kgiusti:


kgiusti:
Did a quick lookover the use of assert in the engine code - LGTM!

PR #231, opened 2020-03-26
PROTON-2200 Fix memory leak in go message unmarshal

d98762625:
When using this library in an application to receive messages from QPID, we found our containers were getting OOM killed in kubernetes. 

This fixes the memory leak we uncovered.

### DIFF
--- a/go/pkg/amqp/message.go
+++ b/go/pkg/amqp/message.go
@@ -326,6 +326,7 @@ func (m *message) Marshal(v interface{}) { m.body = v }
 
 func (m *message) Unmarshal(v interface{}) {
 	pnData := C.pn_data(2)
+	defer C.pn_data_free(pnData)
 	marshal(m.body, pnData)
 	unmarshal(v, pnData)
 }


astitcher:
This fix looks good to me @alanconway what do you think?

go/pkg/amqp/url_test.go
```diff
@@ -32,8 +32,9 @@ func ExampleParseURL() {
 		"amqps://host",
 		"/path",
 		"",
-		":1234",
-                // Taken out because the go 1.4 URL parser isn't the same as later
+		// Taken out because url.Parse error formatting changed in 1.13
```
alanconway: The "taken out" lines can just be deleted - they're not valid URLs, those forms were only ever supported for backwards compat and I'm pretty sure aren't needed anymore.

d98762625: Great, I'll remove those.

d98762625:
I don't know if I should or not, but I am unable to merge this PR myself. I assume we just wait for one of the maintainers to merge it?

jiridanek:
Created a Jira. I am deleting the commits concerning `url_test.go` which was fixed in a different way meanwhile. Should be good to go when CI passes.

PR #235, opened 2020-03-27
[go] Add ability to set capability on the terminus struct

d98762625:
We found that we needed to be able to set capabilities on the terminus struct, otherwise we couldn't share queues.

This PR adds a getter and setter for capabilities on the terminus struct.

### DIFF
--- a/go/pkg/amqp/url_test.go
+++ b/go/pkg/amqp/url_test.go
@@ -32,13 +32,6 @@ func ExampleParseURL() {
 		"amqps://host",
 		"/path",
 		"",
-		":1234",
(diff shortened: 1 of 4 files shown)


go/pkg/amqp/url_test.go
```diff
@@ -32,13 +32,6 @@ func ExampleParseURL() {
 		"amqps://host",
 		"/path",
 		"",
-		":1234",
```
d98762625: Removing this invalid case and the commented out ones, as per the comment on [this PR](https://github.com/apache/qpid-proton/pull/231#discussion_r399227856)

PR #237, opened 2020-04-07
PROTON-2116 [python] add test for the object leak issue in BlockingConnection

jiridanek:
The issue fixed in https://issues.apache.org/jira/browse/PROTON-2116 is fixed on Linux, but not on Windows: the test fails there.

@astitcher This one takes 18 s. I guess it can be easilly sped up by running the outer for loop 3 times, instead of 5 (so 40 % improvement), then it would take larger changes to make it even faster.

```27: Test timeout computed to be: 1500
27: test_blocking_connection_object_leak (blocking_connection_object_leak_test.BlockingConnectionObjectLeakTests)
27: Kills and restarts broker repeatedly, while client is reconnecting. ... FAIL
27: 
27: ======================================================================
27: FAIL: test_blocking_connection_object_leak (blocking_connection_object_leak_test.BlockingConnectionObjectLeakTests)
27: Kills and restarts broker repeatedly, while client is reconnecting.
27: ----------------------------------------------------------------------
27: Traceback (most recent call last):
27:   File "C:\projects\qpid-proton\python\tests\integration\blocking_connection_object_leak_test.py", line 149, in test_blocking_connection_object_leak
27:     self.assertEqual([0] * 4, diffs, "Object counts should not be increasing")
27: AssertionError: Lists differ: [0, 0, 0, 0] != [0, 30, 60, 90]
27: 
27: First differing element 1:
27: 0
27: 30
27: 
27: - [0, 0, 0, 0]
27: + [0, 30, 60, 90]
27: ?     +   +   +
27:  : Object counts should not be increasing
27: 
27: ----------------------------------------------------------------------
27: Ran 1 test in 17.445s
27: 
27: FAILED (failures=1)
27/28 Test #27: python-integration-test ..........***Failed   17.77 sec
test 28
```

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -227,9 +227,10 @@ set (py_bin "${CMAKE_CURRENT_BINARY_DIR}")
 set (py_dll "$<TARGET_FILE_DIR:_cproton>")
 set (py_bld "$<TARGET_FILE_DIR:qpid-proton-core>") # For windows
 set (py_tests "${py_src}/tests")
+set (tests_py "${py_src}/../tests/py")
 
 set (py_path ${CMAKE_BINARY_DIR}/c/tools ${py_bld} $ENV{PATH})
-set (py_pythonpath ${py_tests} ${py_src} ${py_bin} ${py_dll} $ENV{PYTHONPATH})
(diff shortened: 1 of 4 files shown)


jiridanek:
I'll create single PR for all the Python leak tests instead.

PR #239, opened 2020-04-09
PROTON-2190 Fix url_test.go on macOS

jiridanek:


go/pkg/amqp/url_test.go
```diff
@@ -32,13 +35,6 @@ func ExampleParseURL() {
 		"amqps://host",
 		"/path",
 		"",
-		":1234",
-                // Taken out because the go 1.4 URL parser isn't the same as later
```
RoddieKieley, line 36: Go 1.4 or the latest Go 1.14 in this case?

RoddieKieley:
@jiridanek sorry for the delay, haven't been checking personal e-mail where this goes. Looks good to me as I see the test run passed.

go/pkg/amqp/url_test.go
jiridanek: It is an old comment. It is really about go 1.4 when the test apparently broke for the first time, due to parsing differences.

PR #242, opened 2020-04-27
Fix build of SASL related code on Windows

josusky:
The code that uses Cyrus SASL library failed to build on Windows because
of "pthead.h" include. There is for sure a nicer way how to make
locking in a platform independent way. I have tried to keep the
pthread code for Unix and have added a corresponding code for Windows.

### DIFF
--- a/c/src/sasl/cyrus_sasl.c
+++ b/c/src/sasl/cyrus_sasl.c
@@ -32,7 +32,6 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
-#include <pthread.h>
 
 #ifndef CYRUS_SASL_MAX_BUFFSIZE
 # define CYRUS_SASL_MAX_BUFFSIZE (32768) /* bytes */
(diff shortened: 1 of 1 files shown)


jiridanek:
Windows build does not support using Cyrus SASL. Instead, it uses the stub, which supports SASL Anonymous and Plain.

https://github.com/apache/qpid-proton/blob/69f5dfe7eb22432dbec666b5c7a1dca786c93657/python/setup.py.in#L170-L184

Jira: https://issues.apache.org/jira/browse/PROTON-2216

josusky:
The Cyrus SASL library is multiplatform. I do not see a reason why Qpid-proton was restricting itself to less SASL methods on Windows platform. 
Now when the JIRA issue is finally loaded, I can see that you have summed my intention in that issue [PROTON-2216] and added few more notes. What does it mean for me? Should I take any action?

jiridanek:
@josusky I don't know much about SASL and Cyrus SASL in Proton, so I cannot really review the PR, beyond the initial triage (calling this a new feature). I think nothing is required from you at this point.

If we want to support SASL on Windows, update to README.md or INSTALL.md is probably necessary, with build instructions. The code I linked above, in python/setup.py.in, is then also in need of update, I think. I won't ask that of you, though, because I am not certain how this PR is going to proceed.

astitcher:
@josusky Cyrus SASL may well be multiplatfrom, but there is no readily available of the library for Windows as far as I can tell. It is not available on vcpkg for instance and there is no windows package available from CMU.

I'm assuming that you are compiling it yourself - if not could you point to your source. If so, perhaps some brief compilation instructions or a pointer to them would be useful.

jiridanek:
vcpkg request for cyrus sasl, https://github.com/microsoft/vcpkg/issues/5894, currently abandoned

astitcher:
> vcpkg request for cyrus sasl, [microsoft/vcpkg#5894](https://github.com/microsoft/vcpkg/issues/5894), currently abandoned

Unfortunately this is only an issue, rather than an actual PR with the changes necessary to actually build cyrus under vcpkg so it needs someone to do the work. However it does have a pointer to somebodies Windows port on sourceforge.

josusky:
@astitcher Yes, we are building Cyrus SASL from source files. Please see https://github.com/cyrusimap/cyrus-sasl/tree/master/win32

PR #245, opened 2020-05-11
PROTON-2077 Remove pni_snprintf; not needed with Visual Studio 2015 and up

jiridanek:
This fixes overflow issue found earlier by Coverity in pni_snprintf.


gemmellr:
Presumably this would go in after whatever changes Andrew proposed making to actually require 2015?

gemmellr:
Unless this is actually the extent of them that is. I see the JIRA is essentially more specifically 'require 2015'.

jiridanek:
This change is one of the type that Andrew proposed. I think the proposed changes were all essentially "quality of life", motivated by "what can we simplify if our lowest supported VS is 2015 and we can use C99 compiler".

If we have VS2015 as minimal supported VS, we can remove pni_snprintf and replace it with C99 snprintf.

gemmellr:
Yep, I just meant that if there are bigger or more explicit 'drop <2015' changes to be made separately, they should ideally go in before rather than after a smaller less specific change that still has effect of requiring 2015

jiridanek:
Ok, I will not rush with this one.

astitcher:
@jiridanek This is exactly one of the changes I was thinking of, so thank you for doing this. But, I have bigger changes to do with C99 vs C++ code in the Windows code, so don't rush to put this in. Minimally I want to change the README then make the substantive changes.

astitcher:
Also we need to change appveyor to not build on unsupported VS versions so the tests don't fail.

jiridanek:
I've just rebased this without making any other changes, to see how it runs on current Travis and Appveyor configs.

PR #246, opened 2020-05-13
PROTON-2219 Upgrade .travis.yml to test on a good cross-section of available Ubuntu versions

jiridanek:
Travis now fails because clang-10 compilation fails, https://issues.apache.org/jira/browse/PROTON-2207

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -141,8 +141,9 @@ addons:
 
 install:
 - echo "Using PYTHON=${PYTHON:=python}"
+# use older version of virtualenv to workaround https://github.com/pypa/virtualenv/issues/1873
 - ${PYTHON} -m pip install --user --upgrade pip
-- ${PYTHON} -m pip install --user coverage setuptools wheel tox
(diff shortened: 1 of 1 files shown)


astitcher:
We should leave this until I finish the C99 changes as that is the issue causing the clang 10 failure I'm pretty sure - certainly I have this issue on Fedora - it's PROTON-2207: https://issues.apache.org/jira/browse/PROTON-2207

jiridanek:
If I can figure out what's causing python-tox-test to fail and fix that, I'd prefer to commit this, using clang 9 to work around the other issue. This PR used to run fine in CI when I submitted it (except for the clang 10 thing), and only broke in the meantime.

Leaving unfinished PRs languish seems to be only a recipe for misery down the line, trying to put them back into shape.

astitcher:
Sure - I think there's a good chance I'll do it today or Monday anyway, but if not then don't wait longer.

astitcher:
@jiridanek I've finished my C99 changes. So this should now work for clang 10.

PR #253, opened 2020-05-30
PROTON-1800: issue not completely fixed, ignore the failing test

jiridanek:
@astitcher this implements the decorator you suggested to skip test if `AssertionError`s occurred.


astitcher:
I think the name of the new decorator is too general. How about something like "skipOnFailure"?

PR #254, opened 2020-05-30
PROTON-2125: suppress annoying deprecation warning from Minitest

jiridanek:
I am not completely convinced this is a good idea, but it is the only easy thing I could think of (since on ruby <2.7.0 it is not possible to suppress deprecation warnings selectively by a switch).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,11 +19,14 @@
 
 sudo: false
 language: cpp
+env:
+  global:
+  # PROTON-2125 suppress annoying deprecation warning from Minitest in Ruby tests
+  - RUBYOPT='-W0'  # ruby 2.7.0 introduced a more targeted -W:no-deprecated
 
 matrix:
   include:
   - os: linux
-    dist: bionic
     dist: xenial
     sudo: true
     language: cpp


astitcher:
Did you mean to remove the bionic image from the build matrix?

jiridanek:
> Did you mean to remove the bionic image from the build matrix?

Yes, that is in response to message from travis-ci, "Build config validation: matrix.include: duplicate key: dist".

```
matrix:
  include:
  - os: linux
    dist: bionic	
    dist: xenial
```

astitcher:
> > Did you mean to remove the bionic image from the build matrix?
> 
> Yes, that is in response to message from travis-ci, "Build config validation: matrix.include: duplicate key: dist".
> 
Ok, in that case do we want to actually split that change out from this one (as it definitely seems like it needs fixing) and create a separate matrix element for bionic then as I'm guessing the intent was to test on both bionic and xenial.

astitcher:
I don't really like this fix as it only affects one specific CI build rather than all the testing.

I think this fix (if we're going to us it at all) should be in the ruby CMakeLists.txt.

jiridanek:
> Ok, in that case do we want to actually split that change out from this one (as it definitely seems like it needs fixing) and create a separate matrix element for bionic then as I'm guessing the intent was to test on both bionic and xenial.

I don't know what the intention was. I think I introduced that bionic line and it was by mistake ;( My deliberate suggestion now is to have xenial, bionic, but mostly focal in the matrix, as proposed in #246.

> I think this fix (if we're going to use it at all) should be in the ruby CMakeLists.txt.

That's what I wanted to specifically avoid touching. If I am setting options in CMake, I might just as well monkeypatch the warning out of minitest, or possibly even migrate the tests from the deprecated methods (and monkeypatch them back in when using the rhel 7 version of minitest).

Having quickly checked, it looks like `gem` is available as a package in rhel 7. That possibly means that using newer version of minitest from gem, instead of the packaged version, may be possible. Given that it is a test-only dependency.

jiridanek:
Got a better idea what to do here: install old version of minitest using gem in Travis. By old I mean the minimal supported one, whatever is packaged in RHEL 7.

jiridanek:
Ok, Dropped 2.7, used 3.6, 3.7 and 3.8. I got compilation errors from vcpkg on vs2015 with python 3.6, https://ci.appveyor.com/project/ke4qqq/qpid-proton/builds/33887709

```C:\Tools\vcpkg\toolsrc\include\vcpkg/base/stringview.h(46): error C3259: 'constexpr' functions can only have one return statement (compiling source file ..\src\vcpkg\visualstudio.cpp) [C:\Tools\vcpkg\toolsrc\vcpkglib\vcpkglib.vcxproj]
C:\Tools\vcpkg\scripts\bootstrap.ps1 : Building vcpkg.exe failed. Please 
ensure you have installed Visual Studio with the Desktop C++ workload and the 
Windows SDK for Desktop C++.
At line:1 char:4
+ & {& 'C:\Tools\vcpkg\scripts\bootstrap.ps1' }
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorExcep 
   tion
    + FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorExceptio 
   n,bootstrap.ps1
 
ScriptHalted
At C:\Tools\vcpkg\scripts\bootstrap.ps1:411 char:5
+     throw
+     ~~~~~
    + CategoryInfo          : OperationStopped: (:) [], RuntimeException
    + FullyQualifiedErrorId : ScriptHalted
 
Command exited with code 1
```

and python issues on vs2019 with python 3.8 in python-test and python-integration-test. Tox tests passed fine. Probably some misconfiguration? https://ci.appveyor.com/project/ke4qqq/qpid-proton/builds/33886498/job/qiqg570b0n2cgwwk

```
27: Test timeout computed to be: 1500
27: Traceback (most recent call last):
27:   File "C:/projects/qpid-proton/python/tests/proton-test", line 24, in <module>
27:     from proton_tests.main import main
27:   File "C:\projects\qpid-proton\python\tests\proton_tests\__init__.py", line 22, in <module>
27:     from . import codec
27:   File "C:\projects\qpid-proton\python\tests\proton_tests\codec.py", line 25, in <module>
27:     from proton import *
27:   File "C:\projects\qpid-proton\python\proton\__init__.py", line 35, in <module>
27:     from cproton import PN_VERSION_MAJOR, PN_VERSION_MINOR, PN_VERSION_POINT
27:   File "C:\projects\qpid-proton\BLD\python\cproton.py", line 15, in <module>
27:     import _cproton
27: ImportError: DLL load failed while importing _cproton: The specified module could not be found.
27/29 Test #27: python-test ......................***Failed  Required regular expression not found. Regex=[Totals: .* 0 failed
]  0.18 sec
test 28
      Start 28: python-in
```

jiridanek:
Replaced by https://github.com/apache/qpid-proton/pull/266

PR #255, opened 2020-06-01
Fixed idle_timeout connection option not working for SSL connections on Windows.

attila-kun:
Steps to reproduce:
1. Connect your receiver to an SSL-capable broker. The receiver has to be running on Windows.
2. Set the `idle_timeout` connection option to e.g. 2 seconds in your `proton::messaging_handler` instance.
2. In the `on_message` callback, sleep for longer than `idle_timeout`. This should trigger the `on_transport_error` callback with a `amqp:resource-limit-exceeded: local-idle-timeout expired` error. On Windows this does not happen.

I saw that the above scenario worked correctly when connecting to a non-SSL broker but failed with an SSL broker. I looked at the difference between the two code paths and found that for the same input (`transport->output_buffer` being empty and `transport->close_sent == true`) [`transport.c` returns `PN_EOS`](https://github.com/apache/qpid-proton/blob/0.31.0/c/src/core/transport.c#L2722) while `schannel.c` returned `0`. After ensuring that `schannel.c` matches the behaviour of `transport.c`, the problem went away.


attila-kun:
Is there anything I can do to make sure this gets merged soon?

gemmellr:
A start would be to create a JIRA to allow tracking the issue, rebasing your commit against master (rather than merging master), and reference the JIRA in your commit (see existing commits).

attila-kun:
@gemmellr It's done.

gemmellr:
Thanks. Hopefully one of the C-inclined folks will be along to review the change.

cpt-codes:
Hi @gemmellr. I've come across the same issue using version 0.37.0 through vcpkg (happened to be the version I had installed, and I couldn't see any changes related to this in newer versions). I applied this little fix and it seems to work a treat. Any chance this could be reviewed soon?

gemmellr:
@cliffjansen @astitcher ^

c/src/ssl/schannel.cpp
```diff
@@ -1765,6 +1765,12 @@ static ssize_t process_output_ssl( pn_transport_t *transport, unsigned int layer
   if (!ssl) return PN_EOS;
   ssl_log( transport, PN_LEVEL_TRACE, "process_output_ssl( max_len=%d )",max_len );
 
+  // This is to match the behaviour of pn_output_write_amqp defined in transport.c.
+  // Without this, the idle_timeout connection option does not work in case of an SSL connection.
+  if (!pn_buffer_size(transport->output_buffer) && transport->close_sent) {
+    return PN_EOS;
+  }
+
```
astitcher, line 1773: I'm somewhat confused about this change - the equivalent openssl code does not do anything like this but as far as I know works correctly in the face of idle timeout. This makes me think that this is not the correct fix to the problem.
Especially as the check should be redundent in that the amqp layer processing which does have this check will still be running and should pick up on the issue. Clearly there is an issue somewhere that is specific to the schannel implementation, but this seems like a 'voodoo' fix.
@cliffjansen @kgiusti do either of you have any thoughts?

astitcher (commented):
I'm not sure that this fix is in the correct layer, as the openssl code works correctly (it does right?)

PR #260, opened 2020-06-08
PROTON-335: Add access to link attach properties (C and Python)

kgiusti:
pn_link_properties()
  pn_link_remote_properties()


jiridanek:
I guess this is good. It looks similar to other such "accessor" code, and there is a C test and a Python test for this.

CC @astitcher

PR #261, opened 2020-06-28
PROTON-1870 reproducers in Python for overly generic TLS errors

jiridanek:
I do realize that the tests are extremely similar to the existing TLS tests in Proton Python. It felt much easier to write it gradually from scratch than to fit it into existing test helpers right from the beginning.

This issue may require writing similar tests for all the other bindings.


jiridanek:
There is something in the test that starts TLS server in Python that is incompatible with older Python 2.x on Ubuntu, and something in all of the TLS tests that expect Proton TLS to work, that fails on Windows. This issue is on a back burner at this point, so I will not look into fixing this for now.

PR #262, opened 2020-06-28
PROTON-2235 set threaderciser test timeout and modernize compilation

jiridanek:


jiridanek:
The test failures look irrelevant with regards to the changes made.

PR #269, opened 2020-08-06
SASL Changes - Authzid and plugin API review

astitcher:
This is a combined PR both adding authorization id capability to proton-c's SASL handling and tidying up the SASL plugin API.

It make sense to combine them into a single change as adding authorization id capability changes the plugin API so as to make any code using need at least some simple changes. So adding in the renamings and other simple changes coming from the API review at the same time avoids more churn than necessary in code using the API.


astitcher:
This has been merged

PR #272, opened 2020-10-05
PROTON-2171: Add BUILD_EXAMPLES option, to optionally skip building examples

jiridanek:
This seems to be working, but I'll want to test it more, if I forgot anything. It was suspiciously easy to accomplish.


astitcher:
I'm not surprised it's easy ,but what do you want to achieve it?

jiridanek:
I am surprised. I had bad experience with CMake before.

I want to fix PROTON-2171. The reason for the request is that building examples takes quite a lot of time, especially the C++ examples. Disabling everything that doesn't necessarily need to be built shortens compilation time needed, to get the libs.

This option IMO makes sense for Dispatch CI, for example. It could save a few seconds of the compilation time.

PR #274, opened 2020-10-16
PROTON-2169 Require distutils to build Python binding

jiridanek:
Without the module present, the build would fail with a
somewhat cryptic final error saying

```
CMake Error at python/CMakeLists.txt:139 (install):
  install FILES given no DESTINATION!
```

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -48,6 +48,12 @@ set_target_properties(${SWIG_MODULE_cproton_REAL_NAME}
 
 find_package(PythonInterp REQUIRED)
 
+# Distutils is required to build the binding
+check_python_module("distutils.sysconfig" DISTUTILS_MODULE_FOUND)
+if (NOT DISTUTILS_MODULE_FOUND)
+  message(FATAL_ERROR "Distutils module not found; cannot build Python binding.")
+endif()
+
 if (CHECK_SYSINSTALL_PYTHON)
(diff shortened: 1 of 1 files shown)


RoddieKieley:
As a quick note here, the failure in the checks above was due to known issue [PROTON-2231](https://issues.apache.org/jira/browse/PROTON-2231) related to the c threaderciser. The failure should be unrelated to this PR.

jiridanek:
> As a quick note here, the failure in the checks above was due to known issue [PROTON-2231](https://issues.apache.org/jira/browse/PROTON-2231) related to the c threaderciser. The failure should be unrelated to this PR.

The reason I got stuck on the PR is that it is actually quite difficult to install python (python-devel) without distutils. Its in fact impossible in ubuntu, due to package dependencies. It looks like doing that is extremely rare and maybe not worth checking for in CMake...

astitcher:
As far as I know distutils is in the Python standard library (although deprecated and replaced by setuptools) and so should be present for any correct python installation. We don't check that other standard libraries we rely on are present,

The actual issue might be using distutils at all! Especially in the Python CMake file. The setup.py already mostly uses setuptools and probably could stop using distutils completely with some amount of work!

So I'd say on balance this change isn't really needed.

jiridanek:
Ack; I'll close this (and the Jira) and will be more careful not to have broken python installs around.

PR #275, opened 2020-10-20
PROTON-1496: epoll proactor - single timer queue and timerfd for all …

cliffjansen:
…connection heartbeat timers


c/src/proactor/epoll-internal.h
```diff
@@ -217,16 +227,15 @@ typedef struct psocket_t {
   uint32_t working_io_events;
 } psocket_t;
 
-typedef struct pconnection_t {
+struct pconnection_t {
   psocket_t psocket;
   pcontext_t context;
-  ptimer_t timer;  // TODO: review one timerfd per connection
   const char *host, *port;
+  pn_tick_timer_t *timer;     // Idle timeout timer. Protected by and sole use by ctimerq.
```
astitcher, line 234: Comment incorrect: I assume this means really part of context.proactor.ctimerq

c/src/proactor/epoll-internal.h
```diff
@@ -347,6 +366,11 @@ bool start_polling(epoll_extended_t *ee, int epollfd);
 void stop_polling(epoll_extended_t *ee, int epollfd);
 void rearm_polling(epoll_extended_t *ee, int epollfd);
 
+bool ptimer_init(ptimer_t *pt, epoll_type_t ep_type);
+void ptimer_finalize(ptimer_t *pt);
+void ptimer_set(ptimer_t *pt, uint64_t t_millis);
+bool ptimer_callback(ptimer_t *pt);
```
astitcher, line 372: These are now misnamed all internal API has the pni_ prefix

And also seeing as they are now API external to epoll.c why not put them in their own file?

c/src/proactor/epoll-internal.h
```diff
@@ -360,6 +384,13 @@ pcontext_t *pni_raw_connection_context(praw_connection_t *rc);
 praw_connection_t *pni_batch_raw_connection(pn_event_batch_t* batch);
 void pni_raw_connection_done(praw_connection_t *rc);
 
+bool ctimerq_init(pn_proactor_t *p);
+void ctimerq_finalize(pn_proactor_t *p);
+bool ctimerq_register(connection_timerq_t *ctq, pconnection_t *pc);
+void ctimerq_deregister(connection_timerq_t *ctq, pconnection_t *pc);
+void ctimerq_schedule_tick(connection_timerq_t *ctq, pconnection_t *pc, uint64_t deadline, uint64_t now);
+pn_event_batch_t *ctimerq_process(connection_timerq_t *ctq, bool timeout);
+
```
astitcher, line 393: These are misnamed all internal API has pni_ prefix

Also ctimerq_t is not the struct name so they are misnamed that that way too. 
If connection_timerq_t is too long in one context it is in all contexts! I don't see the relevance of the connection_ bit. Why not just timer_queue_t? Why is it specific to connections? Shouldn't it also handle the proactor timers too?

astitcher (changes requested):
I think the naming is messed up in this change, see the comments inline.

More generally I'm not convinced by the factoring. I can't see a reason to introduce another timer as well as the the proactor timer so I'd expected the ptimer code to take on the role of queuing all the combined timer functionality and to either dispatch a proactor timer event or an internal connection timer event to a registered callback either in the proactor or the connection.

cliffjansen:
Reworked with naming changes and a single timerfd as suggested.  A few TSAN and Coverity fixes too.

9722a7f9d3afadffe1883305aff64bfaf977f596
2ec1ba5026ad821f0f7f6919cd14762f04dcadd2
06833acb10f30343d1e2e970a31e11cb290e32ff
d31ad9652a1a63f856ed10772e362b4a155ecbf4

PR #276, opened 2020-11-10
PROTON-2295 avoid broken Go 1.15.3 on macOS CI

jiridanek:
This is mainly tp avoid the broken 1.15.3 which is default on macOS.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,6 +28,12 @@ jobs:
     - name: Create Build and Install directories
       run: mkdir -p "${BuildDir}" "{InstallPrefix}"
       shell: bash
+    # PROTON-2295 avoid using go 1.15.3 on macOS, which is broken
+    - name: Setup go
+      uses: actions/setup-go@v2
+      with:
+        go-version: '^1.15.4'
+      if: runner.os == 'macOS'
     - name: Setup python
       uses: actions/setup-python@v2
       with:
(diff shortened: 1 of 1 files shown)


astitcher:
Since this pin is for every CI build environment not just the failing MacOS, I think it would be better to use the *newest* supported and working version rather than the oldest supported version. I pretty sure that is the case for python (at least when 3.6 was pinned newer versions of python failed under CI on at least some of the CI platforms)

gemmellr:
I'd also go with the newest here, since we are presumably testing older versions in e.g Travis already. We can make the change specific to OSX for now though just to get that working since its specific to a particular version issue there.

jiridanek:
I've changed the PR to run conditionally only on macOS. I'll now watch the GHA, how it runs.

gemmellr:
Great. Seems to have worked.

Stylistic nit, I think Its nicer if the if is the first element of a step, it makes it easier to discern if the step is relevant or not when reading the config...admittedly it is clear from the output/status details for the job that a given step has run or not though.

Could perhaps also rename the step to "Setup go (Mac OS)" to distinguish it?

gemmellr:
Looks good, squash + push.

jiridanek:
> -- Found Go: /Users/runner/hostedtoolcache/go/1.15.4/x64/bin/go (go version go1.15.4 darwin/amd64) (1.15).

PR #277, opened 2020-11-11
PROTON-2298: Disable default running of c-threaderciser tests, enable only through cmake option

kpvdr:
The earlier test to set the default for c-threaderciser is commented out and replaced with a hard OFF.


gemmellr:
I have pushed a modified version of the changes in https://github.com/apache/qpid-proton/commit/db12be7fe55a39e473c414f4bd38d6d45ff4a77a. I removed the commented out code (can fish it out from the commit later if its needed) and tweaked the comment, and removed the enabling flag on the various Windows CI build configs as the test doesnt actually work there currently.

PR #278, opened 2020-12-02
PROTON-2307 Allow access to connection properties in cpp binding

pjfawcett:
Allow setting of connection properties via connection_options
Allow reading of properties on incoming connections
Add unit test to container_test.cpp

### DIFF
--- a/cpp/src/connection.cpp
+++ b/cpp/src/connection.cpp
@@ -21,6 +21,7 @@
 
 #include "proton_bits.hpp"
 
+#include "proton/codec/map.hpp"
 #include "proton/codec/vector.hpp"
 #include "proton/connection.hpp"
 #include "proton/connection_options.hpp"
@@ -190,6 +191,15 @@ std::vector<symbol> connection::offered_capabilities() const {
 std::vector<symbol> connection::desired_capabilities() const {
     value caps(pn_connection_remote_desired_capabilities(pn_object()));
     return get_multiple<std::vector<symbol> >(caps);
+}
+
+std::map<symbol, value> connection::properties() const {
+    std::map<symbol, value> props_ret;
+    value props(pn_connection_remote_properties(pn_object()));
+    if (!props.empty()) {
+        get(props, props_ret);
+    }
+    return props_ret;
 }
 
 bool connection::reconnected() const {
--- a/cpp/src/container_test.cpp
+++ b/cpp/src/container_test.cpp
@@ -90,6 +90,7 @@ class test_handler : public proton::messaging_handler {
     std::string peer_container_id;
     std::vector<proton::symbol> peer_offered_capabilities;
     std::vector<proton::symbol> peer_desired_capabilities;
+    std::map<proton::symbol, proton::value> peer_properties;
     proton::listener listener;
     test_listen_handler listen_handler;
 
@@ -112,6 +113,7 @@ class test_handler : public proton::messaging_handler {
             peer_container_id = c.container_id();
             peer_offered_capabilities = c.offered_capabilities();
             peer_desired_capabilities = c.desired_capabilities();
+            peer_properties = c.properties();
             c.close();
         }
         closing = true;
@@ -179,6 +181,21 @@ int test_container_capabilities() {
     ASSERT_EQUAL(th.peer_offered_capabilities[0], proton::symbol("offered"));
     ASSERT_EQUAL(th.peer_desired_capabilities.size(), 1u);
     ASSERT_EQUAL(th.peer_desired_capabilities[0], proton::symbol("desired"));
+    ASSERT_EQUAL(th.peer_properties.size(), 0u);
+    return 0;
+}
+
+int test_container_properties() {
+    proton::connection_options opts;
+    std::map<proton::symbol, proton::value> props;
+    props["qpid.client_process"] = "test_process";
+    props["qpid.client_pid"] = 123;
+    opts.properties(props);
+    test_handler th("", opts);
+    proton::container(th).run();
+    ASSERT_EQUAL(th.peer_properties.size(), 2u);
+    ASSERT_EQUAL(th.peer_properties["qpid.client_process"], "test_process");
+    ASSERT_EQUAL(th.peer_properties["qpid.client_pid"], 123);
     return 0;
 }
 
@@ -475,6 +492,7 @@ int main(int argc, char** argv) {
     RUN_ARGV_TEST(failed, test_container_default_container_id());
     RUN_ARGV_TEST(failed, test_container_vhost());
     RUN_ARGV_TEST(failed, test_container_capabilities());
+    RUN_ARGV_TEST(failed, test_container_properties());
     RUN_ARGV_TEST(failed, test_container_default_vhost());
     RUN_ARGV_TEST(failed, test_container_no_vhost());
     RUN_ARGV_TEST(failed, test_container_bad_address());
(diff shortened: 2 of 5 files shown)


pjfawcett:
The AppVeyor "Visual Studio 2015" build failed do to a problem with Microsoft vcpkg.  It looks like this has now been [fixed](https://github.com/microsoft/vcpkg/pull/14915)
Is there a way to force a re-run of the AppVeyor build or do I need to do a "no-op push"?

jiridanek:
I don't see option to rerun Appveyor, so the empty commit or empty --amend is probably the only way.

pjfawcett:
This is my first Qpid Pull Request hence I am unfamiliar with the process and would welcome some advice.

Firstly, what should I do about the failed builds?  They seem to be related to errors/problems in the build system(s) rather than to changes I have made, but I might be missing something.

Secondly, and connected, do I need to get this PR to pass all the build checks before it is eligible for review, or is there something else I have to do (apart from being more patient) ?

Thanks

gemmellr:
I believe some of the builds are a little flakey in CI (and there are also some tests enabled in CI that aren't in the default build because of that). If you are confident the failures aren't from your changes then feel free to just say so. Another approach is to give the build a prod and see if anything changes (e.g git commit --amend --reset-author to redo the commit, followed by a force push). 

You don't need to get the tests passing to elicit a review, you just need someone with the requisite knowledge to actually come along (which is not me I'm afraid).

pjfawcett:
Thanks Robbie.  I'll force another push. I was wary of doing it too repeatedly having seen some recent messages about Travis. CI limits (though later messages suggest that isn't relevant)

cpp/src/connection.cpp
```diff
@@ -192,6 +193,13 @@ std::vector<symbol> connection::desired_capabilities() const {
     return get_multiple<std::vector<symbol> >(caps);
 }
 
+std::map<symbol, value> connection::properties() const {
+    map<symbol, value> props(pn_connection_remote_properties(pn_object()));
+    std::map<symbol, value> props_ret;
+    get(props, props_ret);
+    return props_ret;
```
astitcher, line 200: I think you mean:
```
    value props(pn_connection_remote_properties(pn_object()));
    if (props.empty()) {
      return std::map<symbol, value>();
    } else {
      return get<std::map<symbol, value> >(props);
    }

```
This avoids building an entirely unecessary proton::map just to throw it away immediately.

cpp/src/container_test.cpp
```diff
@@ -179,6 +181,22 @@ int test_container_capabilities() {
...
+    ASSERT_EQUAL(th.peer_properties.size(), 0u);
+    return 0;
+}
+
+int test_container_properties() {
+    proton::connection_options opts;
+    std::map<proton::symbol, proton::value> props = {
+            {"qpid.client_process", "test_process"},
+            {"qpid.client_pid", 123}
+    };
```
astitcher, line 193: Irritatingly we still support C++03 (for a little while longer only I think) so this syntax is not acceptable just yet.

astitcher (changes requested):
@pjfawcett Firstly many apologies it has taken me so long to review this PR.

I really think this is a useful addition to the API.

I spent a while thinking about whether using std::map<> was a good idea in the context of this API, and I concluded that it makes sense even if it is a rather heavyweight object to be returning. Another alternative might be to just return a proton::map<> which is comparatively lightweight, but not all that convenient without first converting it to a std::map just as you did internally.

I've a few necessary changes below , but otherwise I think this is good essentially as-is.

I also note that session and link objects also have properties (but these maybe not as useful as the connection properties).

If you rebase this change on master I think you can drop the Visual Studio hacks and the tests should pass as appveyor have now fixed their VS2015 image.

cpp/src/container_test.cpp
jiridanek: Do we? @ssorj afaik we don't have to support C++03 any more on what's to become Proton 0.34, or am I wrong?

astitcher: At this point we still support C++03. I think 0.34 will be the last release to do this (and also python 2.7) We should announce a one release deprecation on the mailing list, and have a clean break from Proton 0.34 to 0.35.

jiridanek: True, nothing was announced yet, that's a fact. Alright, so 03 is still supported.

cpp/src/connection.cpp
pjfawcett: Thank you for your comments.  I agree my code is wasteful in the case where there are no properties and I will rectify that.  However, I think it was necessary to create an intermediate `proton::map` in order to get the `decoder` to process all the properties data, rather than just the first key - buy I may be mis-remembering things so  will have another go.

cpp/src/container_test.cpp
pjfawcett: Thanks for pointing this out, I will change it to something C++03 conformant. 
If C++03 is still being supported, might it be beneficial to have a test build that explicitly uses that variant?

astitcher: Yep, it would make a lot of sense to have an C++03 build. However given it really is going away in a short while probably not worth the bother right now - sorry you had to fix this for 03 when it's going away soon!

cpp/src/connection.cpp
astitcher: It's not the no properties case that is wasteful, it's the with properties case. There it builds the proton::map unnecessarily and throws it away. My replacement code doesn't use a proton::map at all. The whole process is still annoyingly heavyweight, but at least this isn't going to be on the hot path in any sensible use.
My guess is that you used code like mine initially but without the null check (because it's the parallel to the capabilities code). But that doesn't work and fails the unit test you provided. 
As the code passes the unit test I'm reasonably sure it doesn't just decode the first element!

pjfawcett: Thanks for your elucidation.  I was mis-remembering my earlier attempts but your suggestion helped me track down why my earlier code had been failing the unit tests  (memory management in the non-NULL case).
I have ended up with a slight variation on your suggestion due to a preference for single return points, but I think it is as efficient - and an improvement on my first submission.

cpp/src/container_test.cpp
pjfawcett: I agree, a bit late in the day.  However, I found the required setting for cmake to build with C++03 - which fortunately showed up another compatibility error in the PR.

pjfawcett:
@astitcher regarding properties on session and link: I'd be happy to implement similar access to those if it was thought to be useful - probably under a follow-on ticket?

cpp/src/connection.cpp
```diff
@@ -192,6 +193,15 @@ std::vector<symbol> connection::desired_capabilities() const {
...
 
+std::map<symbol, value> connection::properties() const {
+    std::map<symbol, value> props_ret;
+    value props(pn_connection_remote_properties(pn_object()));
+    if (!props.empty()) {
+        get(props, props_ret);
+    }
+    return props_ret;
+}
+
```
astitcher, line 204: I agree this is as efficient as my version - My preference is for multiple returns over (IMO) meaningless local variables. But just my preference - this is fine.

astitcher (approved):

astitcher:
> @astitcher regarding properties on session and link: I'd be happy to implement similar access to those if it was thought to be useful - probably under a follow-on ticket?
 
That would be great! I'll try to get to it much sooner next time!

astitcher:
@pjfawcett If you squash rebase these 2 commits then I'll be happy to merge. (I can do it too, but it'll be a little easier for you)

cpp/src/connection.cpp
pjfawcett: Thanks

pjfawcett:
Squashed into single commit.
Ticket PROTON-2330 opened for sessions / links

PR #280, opened 2020-12-19
PROTON-2104 add test for IPv6 in proton::url

jiridanek:


jiridanek:
This does not conclude PROTON-2104, there is still issue present. Test just shows that the `proton::url` part is working fine.

PR #281, opened 2020-12-19
PROTON-2311 make ctest extra verbose in .travis.yml

jiridanek:


jiridanek:
This does not help with getting a better error message from CTest, so lets close this. Some different direction is needed.

PR #283, opened 2021-01-07
PROTON-2318 Pin simplecov in .travis.yml to resolve gem conflict

jiridanek:


jiridanek:
Ci looks better than before, Travis Coverage job has passed. I am not completely sure this is the best I can do, but as far as I understand rubygem version specifiers, this should be it.

PR #284, opened 2021-01-08
PROTON-2319: [Python] Remove code that only exists to support Python 2.6

astitcher:
Python 2.6 is long dead. Now RHEL 6 is out of support remove Python 2.6 support.


tests/py/test_unittest.py
```diff
@@ -1,83 +0,0 @@
...
-            return lambda f: f
-
-        unittest.skipUnless = skipUnless
-
-
-if sys.version_info >= (2, 7):
-    import unittest
-else:
-    try:
-        import unittest2 as unittest
```
jiridanek, line 79: Isn't there something in CMakeLists.txt which deals with checking for unittest2? If so, then it can be also removed.

c/examples/testme
```diff
@@ -21,7 +21,7 @@
 # Run the C examples and verify that they behave as expected.
 # Example executables must be in PATH
 
-from test_unittest import unittest
```
jiridanek, line 24: This file can still be useful for some Python 2.7 vs 3.6 differences, such as assertRegex/assertRegexpMatches as used in dispatch https://github.com/apache/qpid-dispatch/blob/8b91099c6fbbca2e129326e4fcddf5d24a6cd237/tests/system_test.py#L880-L886

I worry that if the file is taken out completely, we may end up adding it back later (with completely different content, of course).

tests/py/test_unittest.py
astitcher: Ah good point. I'll go look

c/examples/testme
astitcher: We're on the way to removing 2.7 too so I'm not going to worry about this at the moment.

tests/py/test_unittest.py
astitcher: I removed all the build system references and checks for unittest2 that I could find as well and repushed the PR. Let's see how this tests.

PR #285, opened 2021-01-14
PROTON-2320 Apply first round of autopep8 (whitespace changes, mostly)

jiridanek:
```bash
for f in `find -name "*.py"`; do autopep8 --max-line-length 120 --in-place $f; done
```


jiridanek:
I went through the diff and it seems sane (except that github diffing is bit weird sometimes). I will read through it again, after some time.

astitcher:
If we make the python binding pep8 clean we should then add the pep8 test into the default tox tests - currently you could run it explicitly (if you can figure out how!) but it produces so much output that it is useless.

astitcher:
I'm happy to make this change at the moment whilst there are few python developments in process.

jiridanek:
I absolutely want to add flake8 (or do you have different preferences?) to tox. At this point it would be premature, I think, because there is few more `--aggressive` rounds of autopep8 needed, before the code becomes pep8 compliant.

python/examples/simple_send.py
```diff
@@ -37,7 +38,7 @@ def on_start(self, event):
 
     def on_sendable(self, event):
         while event.sender.credit and self.sent < self.total:
-            msg = Message(id=(self.sent+1), body={'sequence':(self.sent+1)})
+            msg = Message(id=(self.sent+1), body={'sequence': (self.sent+1)})
```
jiridanek, line 41: interesting, looks like autopep8 got this line wrong, it is missing spaces around `+`

astitcher:
> I absolutely want to add flake8 (or do you have different preferences?) to tox. At this point it would be premature, I think, because there is few more `--aggressive` rounds of autopep8 needed, before the code becomes pep8 compliant.

flake8 is already in the tox tests it's just not a default profile tested. (You can use 'tox -epep8' to get it - if you also want the other python tests you also need to set an environment var to tell tox where the python tests are - at least currently)

python/tests/proton_tests/main.py
```diff
@@ -330,382 +356,401 @@ def filter(self, record):
...
 
-  def inspect(self, obj):
-    return type(obj) == types.FunctionType and self.matches(obj.__name__)
+    def inspect(self, obj):
+        return type(obj) == types.FunctionType and self.matches(obj.__name__)
+
+    def descend(self, func):
+        # the None is required for older versions of python
+        return
+        yield None
```
jiridanek, line 590: @astitcher maybe something that can be also removed; marking this so it can be found in the future

python/tests/proton_tests/engine.py
```diff
@@ -53,2803 +54,2822 @@ def bytearray(x):
...
+        try:
+            self.c1.transport.channel_max = 666
+            assert False, "expected session exception"
+        except:
+            pass
+
+        assert self.c1.transport.channel_max == upper_limit
+
+    # TODO: Currently failing test - PROTON-1759 - skip
+
```
jiridanek, line 288: unfortunate newline; I won't be redoing the PR because of this

python/tests/proton_tests/main.py
```diff
@@ -330,382 +356,401 @@ def filter(self, record):
...
+    def inspect(self, obj):
+        return type(obj) in CLASS_TYPES and self.matches(obj.__name__)
 
-  def descend(self, cls):
-    # the None is required for older versions of python
-    return; yield None
+    def descend(self, cls):
+        # the None is required for older versions of python
+        return
+        yield None
```
jiridanek, line 604: second occurrence of this

python/tests/proton_tests/main.py
jiridanek: this thing is there multiple times

astitcher: Hmm, sure looks like it can go - best to check that the comment doesn't apply any more - what versions are 'older' for the purposes of this comment?

tools/python/mllib/__init__.py
```diff
@@ -43,43 +45,47 @@
...
-  h = parsers.XMLParser()
-  p = xml.sax.make_parser()
-  p.setContentHandler(h)
-  p.setErrorHandler(ErrorHandler())
-  p.setEntityResolver(Resolver(path))
-  p.parse(source)
-  return h.parser.tree
+    if sys.version_info[0:2] == (2, 3):
+        # XXX: this is for older versions of python
+        source = "file://%s" % os.path.abspath(filename)
```
jiridanek, line 76: @astitcher ancient remnant of Pythons long gone

python/tests/proton_tests/main.py
astitcher: Huh, isn't this the same code again?

tools/python/mllib/__init__.py
astitcher: ooh, more code to remove...

tools/python/mllib/dom.py
```diff
@@ -32,285 +32,308 @@
...
-                         (f.__class__.__name__, attrs))
+    def is_type(self, type):
+        cls = self
+        while cls != None:
+            if cls.type == type:
+                return True
+            cls = cls.base
+        return False
+
+    def dispatch(self, f, attrs=""):
```
jiridanek, line 84: autopep deleted spaces here

jiridanek: eh, never mind, it _is_ consistent

python/examples/simple_send.py
jiridanek: it is only putting the spaces there with more --aggressive options added; weird, but ok

python/tests/proton_tests/main.py
jiridanek: No idea. Yield was added in Python 2.5 and yield without that `None` works for me in Python 2.7. So the "older versions" was either 2.5 or 2.6, or both, or maybe Jython?

Anyways, it is trying to create an empty generator function. After some googling, I think that this code can be replaced with

```python
def extract(self, obj):
    return ()
```

or, if we want to maintain the precise type,

```python
def extract(self, obj):
    return iter(())
```

This is better because return; yield can trigger warnings about unreachable code in some tooling.

tools/python/mllib/__init__.py
jiridanek: #288

python/tests/proton_tests/main.py
jiridanek: yes, there are three copies of this in total

PR #290, opened 2021-01-20
PROTON-2326: epoll proactor refactor - "schedule" instead of "wake", …

cliffjansen:
Part 1 of epoll proactor code cleanup:

 wake->schedule
 context->task
 wake_notify -> notify_poller
 WAKE -> EVENT_FD (epoll_type-t)
 wake_pending, wake_list_xxx -> ready, ready_list_xxx
 similar changes to use of wake and ctx as local vars
 unused sched_wakes_pending removed

### DIFF
--- a/c/src/proactor/epoll-internal.h
+++ b/c/src/proactor/epoll-internal.h
@@ -86,8 +86,9 @@ typedef struct task_t {
   pn_proactor_t *proactor;  /* Immutable */
   task_type_t type;
   bool working;
-  bool on_ready_list;
   bool ready;                // ready to run and on ready list.  Poller notified by eventfd.
+  bool waking;
+  bool on_ready_list;        // todo: protected by eventfd_mutex or sched mutex?  needed?
   struct task_t *ready_next; // ready list, guarded by proactor eventfd_mutex
   bool closing;
   // Next 4 are protected by the proactor mutex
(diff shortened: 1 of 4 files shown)


c/src/proactor/epoll-internal.h
```diff
@@ -79,36 +79,36 @@ typedef enum {
...
   bool disconnecting;           /* pn_proactor_disconnect */
   // Protected by schedule mutex
   tslot_t *runner __attribute__((aligned(64)));  /* designated or running thread */
   tslot_t *prev_runner;
-  bool sched_wake;
+  bool sched_ready;
   bool sched_pending;           /* If true, one or more unseen epoll or other events to process() */
-  bool runnable ;               /* in need of scheduling */
-} pcontext_t;
+  bool runnable ;               /* on one of the runnable lists */
```
astitcher, line 103: Looking at all the bools here, I wonder if there is actually a hidden state machine that would make the task state more comprehensible and safer. I expect that these bools aren't truly independent and so naming and coding out the actual allowed states might make the lifecycle of tasks easier to understand.
There might be some locking issues here as the bools seem to be covered by different locks, but this seems to be a code smell in itself anyway as I'd expect things covered by a lock to be in the same structure as the lock not elsewhere!!

astitcher (approved):
This basically looks fine - There is no actual logic change here anyway (is there?)
TBH  the only way to know if it's an improvement is to get someone with no knowledge of the code to try and understand it!

PR #22, opened 2015-04-24
NO-JIRA: README improvements

dnwe:
* remove out-of-date product information from the README and include the more
  current information from the website
* add a new INSTALL file to contain the build and install instructions
  from the existing README
* move all development and testing related info from that to the
  DEVELOPERS file
* better explain how to run the various test suites and how they work
  across proton-c and proton-j in the developers file
* other misc improvements across the contents of the files
* rename DEVELOPERS and INSTALL to have .md suffixes for better
  rendering on GitHub
* add a symlink from README --> README.md for better landing page
  rendering on GitHub

PR #291, opened 2021-01-22
Tkettenb

tkettenbach-dwd:
Hi all,

this is a minor issue, there is no jira Task for this. I am addressing an issue in a cpp example.

Changes:
 - Added 'bad_option' Exception handler, so the help message is displayed.
 - Fix '-c' command line option. It now searches in the given directory for tserver, tclient files, old behaviour was falling back to default setting 'ssl-certs/'

Test Suite fails on one test-case, but i can't tell where it breaks.

    $ make test

fails in:
    Start 22: cpp-connect_config_test
    22/29 Test #22: cpp-connect_config_test ..........***Failed    0.15 sec

BR,
Thomas

### DIFF
--- a/cpp/examples/ssl.cpp
+++ b/cpp/examples/ssl.cpp
@@ -176,6 +176,8 @@ int main(int argc, char **argv) {
         hello_world_direct hwd;
         proton::container(hwd).run();
         return 0;
+    } catch (const example::bad_option& e) {
+        std::cout << opts << std::endl << e.what() << std::endl;
     } catch (const std::exception& e) {
         std::cerr << e.what() << std::endl;
     }
--- a/cpp/examples/ssl.cpp
+++ b/cpp/examples/ssl.cpp
@@ -165,9 +165,10 @@ int main(int argc, char **argv) {
         opts.parse();
 
         size_t sz = cert_directory.size();
-        if (sz && cert_directory[sz -1] != '/')
-            cert_directory.append("/");
-        else cert_directory = "ssl-certs/";
+        if (sz) {
+           if (cert_directory[sz -1] != '/')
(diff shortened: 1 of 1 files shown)


astitcher (commented):
This change is easily sufficient to need a Jira (sorry!) the quantity of change is not relevant - this is a change of logic that is just not obvious so I'd really like more explanation what the change of certificate directory logic is and why it's preferable.
Usually this would be in both commit message and Jira.

cpp/examples/ssl.cpp
```diff
@@ -165,16 +165,19 @@ int main(int argc, char **argv) {
         opts.parse();
 
         size_t sz = cert_directory.size();
-        if (sz && cert_directory[sz -1] != '/')
-            cert_directory.append("/");
-        else cert_directory = "ssl-certs/";
+        if (sz) {
+           if (cert_directory[sz -1] != '/')
+               cert_directory.append("/");
+        } else cert_directory = "ssl-certs/";
```
astitcher, line 171: Bad indentation (yes I know the original was bad too!) please use explicit '{' and '}' and indent. I'd also prefer the else to be indented in a '{' '}' block too, but that's not bad indentation (IMO).

astitcher:
To be clear (now that I've spent a couple of minutes looking at your change it is a fix to previously wrong logic, but nowhere is it explained what the previous bad behaviour is - that is exactly what a Jira is for.

tkettenbach-dwd:
I've prepared [PROTON-2329](https://issues.apache.org/jira/browse/PROTON-2329) for this.

astitcher:
@tkettenbach-dwd thanks for that - I've merged this with small formating changes

PR #294, opened 2021-02-02
PROTON-2217 Prefer first working Python on PATH, prefer Py3 over Py2

jiridanek:
Use https://cmake.org/cmake/help/git-stage/module/FindPython.html

There is few issues with doing this

* branching on CMake version is bad practice
* PYTHON_EXECUTABLE can be specified from outside, this changes user API (depending on CMake version)
* Travis CI apparently uses CMake 3.12 (on all Ubuntu versions); can be upgraded

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,27 @@ include (CheckPythonModule)
 
 find_package (OpenSSL)
 find_package (Threads)
-find_package (PythonInterp REQUIRED)
+# FindPython was added in CMake 3.12, but there it always returned
+#  newest Python on the entire PATH. We want to use the first one.
+if (CMAKE_VERSION VERSION_LESS "3.15.0")
+  find_package (PythonInterp REQUIRED)
+  # forward compatibility with FindPython
+  set(Python_EXECUTABLE "${PYTHON_EXECUTABLE}")
+  set(Python_LIBRARIES "${PYTHON_LIBRARIES}")
+  set(Python_INCLUDE_DIRS "${PYTHON_INCLUDE_PATH}")
+  set(Python_VERSION_STRING "${PYTHON_VERSION_STRING}")
+  # for completeness, these are not actually used now
+  set(Python_VERSION_MAJOR "${PYTHON_VERSION_MAJOR}")
+  set(Python_VERSION_MINOR "${PYTHON_VERSION_MINOR}")
+  set(Python_VERSION_PATCH "${PYTHON_VERSION_PATCH}")
+else ()
+  if (POLICY CMP0094)  # https://cmake.org/cmake/help/latest/policy/CMP0094.html
+    cmake_policy(SET CMP0094 NEW)  # FindPython should return the first matching Python on PATH
+  endif ()
+
+  find_package(Python REQUIRED COMPONENTS Interpreter
+          OPTIONAL_COMPONENTS Development)
+endif()
 find_package (SWIG)
 find_package (CyrusSASL)
 
@@ -47,7 +67,7 @@ set (ProtonCpp_DIR ${CMAKE_CURRENT_SOURCE_DIR}/tools/cmake/config)
 
 ## Variables used across components
 
-set (PN_ENV_SCRIPT "${PYTHON_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/scripts/env.py")
+set (PN_ENV_SCRIPT "${Python_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/scripts/env.py")
 set (PN_C_INCLUDE_DIR "${CMAKE_BINARY_DIR}/c/include")
 set (PN_C_LIBRARY_DIR "${CMAKE_BINARY_DIR}/c")
 set (PN_C_SOURCE_DIR "${CMAKE_BINARY_DIR}/c/src")
@@ -383,8 +403,11 @@ if(SWIG_FOUND)
   # DEFAULT_{uppercase name of binding} to ON
 
   # Prerequisites for Python wrapper:
-  find_package (PythonLibs ${PYTHON_VERSION_STRING} EXACT)
-  if (PYTHONLIBS_FOUND)
+  if (CMAKE_VERSION VERSION_LESS "3.15.0")
+    find_package (PythonLibs ${PYTHON_VERSION_STRING} EXACT)
+    set (Python_Development_FOUND "${PYTHONLIBS_FOUND}")
+  endif ()
+  if (Python_Development_FOUND)
     set (DEFAULT_PYTHON ON)
   endif ()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,27 +32,9 @@ include (CheckPythonModule)
 
 find_package (OpenSSL)
 find_package (Threads)
-# FindPython was added in CMake 3.12, but there it always returned
-#  newest Python on the entire PATH. We want to use the first one.
-if (CMAKE_VERSION VERSION_LESS "3.15.0")
-  find_package (PythonInterp REQUIRED)
-  # forward compatibility with FindPython
-  set(Python_EXECUTABLE "${PYTHON_EXECUTABLE}")
-  set(Python_LIBRARIES "${PYTHON_LIBRARIES}")
-  set(Python_INCLUDE_DIRS "${PYTHON_INCLUDE_PATH}")
-  set(Python_VERSION_STRING "${PYTHON_VERSION_STRING}")
-  # for completeness, these are not actually used now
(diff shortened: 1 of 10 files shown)


CMakeLists.txt
```diff
@@ -32,7 +32,27 @@ include (CheckPythonModule)
...
+  set(Python_VERSION_MINOR "${PYTHON_VERSION_MINOR}")
+  set(Python_VERSION_PATCH "${PYTHON_VERSION_PATCH}")
+else ()
+  if (POLICY CMP0094)  # https://cmake.org/cmake/help/latest/policy/CMP0094.html
+    cmake_policy(SET CMP0094 NEW)  # FindPython should return the first matching Python on PATH
+  endif ()
+
+  find_package(Python REQUIRED COMPONENTS Interpreter
+          OPTIONAL_COMPONENTS Development)
+endif()
```
astitcher, line 55: **This is totally the wrong approach!**
We already have a way to "branch" the processing so that we can use newer features. Create our own FindPython.cmake in tools/cmake/Modules which imitates the cmake 3 find_package(Python) processing so far as we use it.
Also make sure that any installed script will work with cmake 2.8.12 as our target environment is cmake 2.8.12 (sadly).

astitcher (changes requested):

astitcher:
I'd also say that you might as well use find_package(Python3 REQUIRED) as well (obviously you'd need to wait until we drop Python 2 - but not too long now).

CMakeLists.txt
jiridanek: Is this going to work? Because my own FindPython.cmake will end up shadowing the CMake3's find_package(Python), I think. Not sure, I'll have to try that.

jiridanek:
It could be `find_package(Python "3.6.0" REQUIRED)`, maybe. That way, the script will need not have to change when Python 4 comes out ;P

astitcher:
> It could be `find_package(Python "3.6.0" REQUIRED)`, maybe. That way, the script will need not have to change when Python 4 comes out ;P

By all accounts that is *never* going to happen! (2->3 was so painful)

CMakeLists.txt
astitcher: Yes your version will shadow the system one - but you can get it explicitly go look at tools/cmake/Modules/FindThreads.cmake for instance - we've done this before!

jiridanek: Aha, makes sense; I agree that's the correct way. At least while CMake < 3.15 needs to be supported....

tools/cmake/Modules/FindPython.cmake
```diff
@@ -0,0 +1,49 @@
...
+    set(Python_VERSION_MAJOR "${PYTHON_VERSION_MAJOR}")
+    set(Python_VERSION_MINOR "${PYTHON_VERSION_MINOR}")
+    set(Python_VERSION_PATCH "${PYTHON_VERSION_PATCH}")
+
+    find_package (PythonLibs ${PYTHON_VERSION_STRING} EXACT)
+    set(Python_Development_FOUND "${PYTHONLIBS_FOUND}")
+else ()
+    if (POLICY CMP0094)  # https://cmake.org/cmake/help/latest/policy/CMP0094.html
+        cmake_policy(SET CMP0094 NEW)  # FindPython should return the first matching Python on PATH
+    endif ()
```
astitcher, line 42: Given this checks if the policy exists I think you could hoist this out of the if block checking the cmake version - I think this would be a little tidier.

tools/cmake/Modules/FindPython.cmake
```diff
@@ -0,0 +1,49 @@
...
+
+    if (DEFINED PYTHON_EXECUTABLE)
+        set(Python_EXECUTABLE ${PYTHON_EXECUTABLE})
+    endif ()
```
astitcher, line 46: What if both PYTHON_EXECUTABLE and Python_EXECUTABLE are set? I know it seems unlikely, but if it happens I think what this does might be surprising - I'd expect Python_EXECUTABLE not to be overridden.

astitcher (approved):
This looks good - you should consider the comments I made about FindPython.cmake. But I don't feel strongly enough about them to stop you committing this to master.

So address (or not) my concerns, and feel free to rebase, squash and commit.

tools/cmake/Modules/FindPython.cmake
```diff
@@ -0,0 +1,53 @@
...
+if (CMAKE_VERSION VERSION_LESS "3.15.0")
+    if (DEFINED Python_EXECUTABLE)
+        set(PYTHON_EXECUTABLE ${Python_EXECUTABLE})
+    endif ()
+
+    find_package (PythonInterp REQUIRED)
+    # forward compatibility with FindPython
+    set(Python_EXECUTABLE "${PYTHON_EXECUTABLE}")
+    set(Python_LIBRARIES "${PYTHON_LIBRARIES}")
+    set(Python_INCLUDE_DIRS "${PYTHON_INCLUDE_PATH}")
```
jiridanek, line 34: hmm, PYTHON_INCLUDE_PATH won't be set until I find PythonLibs

tools/cmake/Modules/FindPython.cmake
jiridanek: @astitcher if both are set, I'd fail the build. WDYT?

astitcher (approved):
I like this. Good job!

jiridanek:
Windows GHA build is broken by this. I'll have to figure out what's wrong there.

astitcher:
> Windows GHA build is broken by this. I'll have to figure out what's wrong there.
For some reason the cmake configure is finding python 3.9.1 even though the build script explicitly asks for 3.6. I'm not sure that's exactly the cause, but it is something that should not be happening.

tools/cmake/Modules/FindPython.cmake
```diff
@@ -0,0 +1,66 @@
...
+    endif ()
+
+    # needed on GitHub Actions CI: actions/setup-python does not touch registry/frameworks on Windows/macOS
+    # this mirrors PythonInterp behavior which did not consult registry/frameworks first
+    if (NOT DEFINED Python_FIND_REGISTRY)
+        set(Python_FIND_REGISTRY "LAST")
+    endif ()
+    if (NOT DEFINED Python_FIND_FRAMEWORK)
+        set(Python_FIND_FRAMEWORK "LAST")
+    endif ()
```
jiridanek, line 63: @astitcher This should take care about finding the right Python on GitHub Actions. New CMake defaults to looking for Virtual environment (and in CI does not find any), then goes to look into Windows Registers. The default behavior consults PATH last, if everything else failed. It's in the docs, I somehow forgot I read about it before https://cmake.org/cmake/help/v3.15/module/FindPython.html#hints.

jiridanek:
`GitHub Actions has encountered an internal error when running your job. ` That's a first for me.

tools/cmake/Modules/FindPython.cmake
```diff
@@ -0,0 +1,66 @@
...
+    set(Python_Development_FOUND "${PYTHONLIBS_FOUND}")
+    set(Python_INCLUDE_DIRS "${PYTHON_INCLUDE_PATH}")
+    set(Python_LIBRARIES "${PYTHON_LIBRARIES}")
+else ()
+    if (DEFINED PYTHON_EXECUTABLE)
+        set(Python_EXECUTABLE ${PYTHON_EXECUTABLE})
+    endif ()
+
+    # needed on GitHub Actions CI: actions/setup-python does not touch registry/frameworks on Windows/macOS
+    # this mirrors PythonInterp behavior which did not consult registry/frameworks first
```
astitcher, line 57: II think this comment should just read:
`# This mirrors PythonInterp behavior on windows, which did not consult registry/frameworks first.`
IMO the specifics of github actions behaviour is not that important.

tools/cmake/Modules/FindPython.cmake
astitcher, line 63: Hmm, not sure this bit of code should go in the find module wrapper at all actually. As when we are able to only use cmake 3.15 everywhere (say in a decade or so!) we'll still need code like this for Windows/Mac.
Maybe we should have these lines:
```
set(Python_FIND_REGISTRY "LAST")
set(Python_FIND_FRAMEWORK "LAST")
```
before the `find_package(Python...)` line in `CMakeLists.txt`.
Possibly in both places?

jiridanek: > say in a decade or so!

Middle of this year, by my estimate! It is contingent on RHEL 8.4 being released and on getting rid of some of the older Ubuntu in CI.

I'm putting the `set`s in the top level CMake only. And want to keep the ifs around it, so that users can override it, if they really want to prioritize systemwide setting for python. I guess that CMake made that the default behavior for a good reason. It is only my fault that I cannot figure what reason that might be,

PR #295, opened 2021-02-05
fix invalid unsigned arithmetic.

ihsinme:
your check is incorrect. since the variables are unsigned, it is equivalent to store->hwm != store->lwm.
so I suggest a simple fix for the error.

### DIFF
--- a/c/src/messenger/store.c
+++ b/c/src/messenger/store.c
@@ -421,7 +421,7 @@ int pni_store_update(pni_store_t *store, pn_sequence_t id, pn_status_t status,
     }
   }
 
-  while (store->hwm - store->lwm > 0 &&
+  while (store->hwm > store->lwm &&
          !pn_hash_get(store->tracked, store->lwm)) {
     store->lwm++;
   }


ihsinme:
this is strange.
such a simple change affects the coverage so much.
I need time to understand why.

ihsinme:
it seems to me a decrease in coverage.
and the collection error is related to the time limit for the test.
not with my changes in the code.

jiridanek (approved):
lgtm

I think that the coverage changes are caused by 1) slight nondeterminism in the tests themselves (coverage normally fluctuates +-10 or so lines from run to run) 2) combination of the fact that last coverage job on master failed (too much output) and that new version of CodeCov gem was released 4th of February (https://rubygems.org/gems/codecov/versions/0.4.3). The new version behaves better. Our job uses the latest gems.

c/src/messenger/store.c
```diff
@@ -421,7 +421,7 @@ int pni_store_update(pni_store_t *store, pn_sequence_t id, pn_status_t status,
     }
   }
 
-  while (store->hwm - store->lwm > 0 &&
+  while (store->hwm > store->lwm &&
```
jiridanek, line 424: I'd feel safer replacing this with `store->hwm != store->lwm` just to keep the code behavior completely identical. I haven't managed to convince myself that this change is always going to work... Somebody else who actually understands this needs to review as well.

astitcher (commented):
Unfortunately the premise of this change is wrong!
This unsigned subtraction is entirely correct and your fix is incorrect.
In C (and C++) unsigned arithmetic is defined to wrap according to the usual 2 complement modulus rules. So as long as the difference between the high and low water marks is less than half the range of the integer (in this case 2^31) the subtraction gives the correct result irrespective of the high and low water marks wrapping around.

c/src/messenger/store.c
astitcher: See my comment - the change is incorrect.

ihsinme:
it seems to me a practical example that will help to understand why I proposed changes.
```cpp
unsigned long A;

// A= 70; (A - 11) = 59; true
// A= 11; (A - 11) = 0; false
// A= 1; (A - 11) = 18446744073709551607; true
// A= 0; (A - 11) = 18446744073709551606; true
if (A - 11 > 0)
```

astitcher:
> it seems to me a practical example that will help to understand why I proposed changes.
> 
> ```c++
> unsigned long A;
> 
> // A= 70; (A - 11) = 59; true
> // A= 11; (A - 11) = 0; false
> // A= 1; (A - 11) = 18446744073709551607; true
> // A= 0; (A - 11) = 18446744073709551606; true
> if (A - 11 > 0)
> ```

astitcher:
Apologies, you are correct that the test is useless and will always return true, due to the unsigned. But your fix is incorrect because it fails to understand the intent of the test!

The test is attempting to make sure that the hwm is ahead of the lwm, but if the lwm is (for a 16 bit example)at 0xffff and the hwm is at 0x0003 then the test should indeed be true (because of the wrapping nature of the buffer), but in your 'fix' it will be false.

I think the correct test is something like hwm-lwm < 0x8000 (in the 16 bit case).

ihsinme:
It’s more and more difficult for me to understand what confuses you.
we're talking about `typedef uint32_t  pn_sequence_t;`.
while it will work as written if `lwm = 0xffffffff` and `hwm = 0x3` then `lwm> hwm`.
if you are talking about the inaccuracy of the initial check, then explain it in more detail.
because `hwm-lwm <0x80000000` is generally dangerous in my opinion.

astitcher:
> It’s more and more difficult for me to understand what confuses you.
That's funny, because I feel similarly about your comments! I guess we're talking somewhat at cross purposes.

> we're talking about `typedef uint32_t pn_sequence_t;`.
> while it will work as written if `lwm = 0xffffffff` and `hwm = 0x3` then `lwm> hwm`.

If you are talking about the mathematics of unsigned ints then this is trivially true! However in the terms of the underlying buffer it is not true. The buffer is designed so that both hwm and lwm only ever increase and so eventually they will both wrap around from MAX_UINT to 0. This is fine and is exactly how unsigned ints are supposed to work. However it is an invariant of the buffer that lwm refers to a later buffer entry than hwm even when wrapped.

> if you are talking about the inaccuracy of the initial check, then explain it in more detail.
What inaccuracy? So I suppose that answer here is no - That isn't my concern.

> because `hwm-lwm <0x80000000` is generally dangerous in my opinion.

I don't understand this at all! And I think it's actually wrong! What this test  does is effectively check the most significant bit of the result of the subtraction. Since we've subtracted unsigned ints the answer can never be really negative, but the answer will still have the most significant bit set if the answer would have been negative if the result was signed.

So as long as the difference between hwm and lwm is always less than MAX_UINT/2 (which is the case here) the test I'm suggesting makes sure to run the loop until lwm equals hwm even if hwm is wrapped around and lwm is not. Whereas the one you are suggesting does not work in this case.

Actually the current test does in practice work correctly too but only because it ends up testing lwm==hwm so I think it's an accident given the way the code is written. And if for some reason the code was changed so that lwm was incremented by more than 1 it would no longer be correct.

ihsinme:
you definitely know better.
tell me why you think that the difference is guaranteed to be less than `MAX_UINT/2`?

additionally, I note that in my opinion `hwm-lwm>0&&hwm-lwm<0x8000` is equivalent to `hwm>lwm`

but in any case, you know better, so tell me what to do next with this PR.

astitcher:
> you definitely know better.
> tell me why you think that the difference is guaranteed to be less than `MAX_UINT/2`?

TBH, it's probably not absolutely guaranteed, but it would be an extremely large number of queued up messages (2 billion or so). Whereas the wrap-around will always happen after processing 4 billion or so messages. Do you see the difference? In other words the wrap around is guaranteed to happen if the system stays up long enough to process 4 billion messages, but we only require that the queued messages to be fewer than 2 billion or so - we probably don't have nearly enough memory to hold that many queud messages!

However in the case of this code it turns out that it is actually already correct even in the case where the difference is not guaranteed - thank you for making me think carefully about it. It's just unclearly written and probably correct by accident!

> 
> additionally, I note that in my opinion `hwm-lwm>0&&hwm-lwm<0x8000` is equivalent to `hwm>lwm`

Hmm, not really true (I don't see how it can be an opinion!): For example:

```
uint16_t lwm = 0xfffe;
uint16_t hwm = 0x0003;
```

`hwm>lwm == false`
`hwm-lwm>0&&hwm-lwm<0x8000 == true`

But in the context of our tests we want the second outcome not the first.

> 
> but in any case, you know better, so tell me what to do next with this PR.

Well, as we've seen there actually isn't a bug here in the first place. But, if you are desperate to change code and not to accept the code I suggest you could change it to simply: `lwm != hwm` which is the current effect of the code, and is correct although a little brittle if other code changes.

ihsinme:
> Well, as we've seen there actually isn't a bug here in the first place. But, if you are desperate to change code and not to accept the code I suggest you could change it to simply: `lwm != hwm` which is the current effect of the code, and is correct although a little brittle if other code changes.

absolutely not.
I wrote above what you know best.
my PR is just a suggestion.

> Hmm, not really true (I don't see how it can be an opinion!): For example:

forgive my statement is really stupid. sorry.

ChugR:
Dear ihsinme,

I'm very discouraged to hear your description of the statement you want forgiven in such terms and your apology for it. Please let me encourage you by saying that qpid-proton is a massive undertaking and that it it quite easy to misunderstand the purpose and intent of some specific lines of code. I personally on a routine basis attribute errors to qpid-proton when the error is in some other component, so I can say  from experience that "misunderstanding proton" is "normal". You have just joined the club with me!

When I first saw your issue I wondered if the logic in question could not have been replaced with TCP SEQ_CMP logic, which has survived the test of time as being correct in that problem space. But as usual I was busy and I never got back to replying over the last few days. I'm sorry about that.

Please stay on and challenge the qpid-proton code base wherever you see room for improvement. I am proud to see someone digging as deeply into it as you have.

Regards,
Chuck (chug, ChugR)

ihsinme:
thanks for your words.

PR #299, opened 2021-03-04
NO-JIRA (yet) Add -DPROACTOR=libuv jobs in GHA and TravisCI

jiridanek:
lets see this on travis


jiridanek:
I was hoping something will fail in a similar way macOS jobs do. That did not happen.

PR #302, opened 2021-03-31
PROTON-2355: Fix build with -DPROACTOR=none

jiridanek:
epoll proactor unconditionally uses pthread.h which will result in the
following build failure:

[  3%] Building C object c/CMakeFiles/qpid-proton-proactor-objects.dir/src/proactor/epoll.c.o
In file included from /nvme/rc-buildroot-test/scripts/instance-0/output-1/build/qpid-proton-0.33.0/c/src/proactor/epoll.c:60:
/nvme/rc-buildroot-test/scripts/instance-0/output-1/build/qpid-proton-0.33.0/c/src/proactor/epoll-internal.h:37:10: fatal error: pthread.h: No such file or directory
   37 | #include <pthread.h>
      |          ^~~~~~~~~~~

To fix this failure, the user could use -DPROACTOR=none but it also
fails on:

CMake Error at c/CMakeLists.txt:481 (add_library):
  Error evaluating generator expression:

    $<TARGET_OBJECTS:qpid-proton-proactor-objects>

  Objects of target "qpid-proton-proactor-objects" referenced but no such
  target exists.

### DIFF
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -464,7 +464,11 @@ set(qpid-proton-noncore-src
   ${qpid-proton-include-extra}
 )
 
-add_library (qpid-proton SHARED $<TARGET_OBJECTS:qpid-proton-core-objects> $<TARGET_OBJECTS:qpid-proton-platform-io-objects> $<TARGET_OBJECTS:qpid-proton-proactor-objects> ${qpid-proton-noncore-src})
+add_library (qpid-proton SHARED
+  $<TARGET_OBJECTS:qpid-proton-core-objects>
+  $<TARGET_OBJECTS:qpid-proton-platform-io-objects>
+  $<$<TARGET_EXISTS:qpid-proton-proactor-objects>:$<TARGET_OBJECTS:qpid-proton-proactor-objects>>
+  ${qpid-proton-noncore-src})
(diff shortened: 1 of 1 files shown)


c/CMakeLists.txt
```diff
@@ -464,7 +464,11 @@ set(qpid-proton-noncore-src
   ${qpid-proton-include-extra}
 )
 
-add_library (qpid-proton SHARED $<TARGET_OBJECTS:qpid-proton-core-objects> $<TARGET_OBJECTS:qpid-proton-platform-io-objects> $<TARGET_OBJECTS:qpid-proton-proactor-objects> ${qpid-proton-noncore-src})
+add_library (qpid-proton SHARED
+  $<TARGET_OBJECTS:qpid-proton-core-objects>
+  $<TARGET_OBJECTS:qpid-proton-platform-io-objects>
+  $<$<BOOL:${HAS_PROACTOR}>:$<TARGET_OBJECTS:qpid-proton-proactor-objects>>
```
astitcher, line 470: I suggest $<$<TARGET_EXISTS:qpid-proton-proactor-objects>:$<TARGET_OBJECTS:qpid-proton-proactor-objects>>
which would be more future proof.

astitcher (dismissed):

c/CMakeLists.txt
jiridanek: I considered that, but then I thought that `HAS_PROACTOR` is more directly derived from what options the user specified, whereas `$<TARGET_EXISTS` is a check for whatever the build ended up doing. So I felt that HAS_PROACTOR is safer.

I'll change it to TARGET_EXISTS.

PR #304, opened 2021-04-02
PROTON-2344: fix Python BlockingConnection resource leaks, memory and…

cliffjansen:
… socket fds

### DIFF
--- a/python/proton/_utils.py
+++ b/python/proton/_utils.py
@@ -440,11 +440,16 @@ class BlockingConnection(Handler):
                 self.conn.close()
                 self.wait(lambda: not (self.conn.state & Endpoint.REMOTE_ACTIVE),
                           msg="Closing connection")
+                if self.conn.transport:
+                    # Close tail to force transport cleanup without waiting/hanging for peer close frame.
+                    self.conn.transport.close_tail()
         finally:
             self.conn.free()
             # Nothing left to block on.  Allow reactor to clean up.
             self.run()
-            self.conn = None
+            if self.conn:
+                self.conn.handler = None  # break cyclical reference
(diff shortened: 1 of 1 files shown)


cliffjansen:
BlockingConnection.close() must ensure eventual cleanup of all resources by the time it returns and be resilient against exceptions.

The strategy under the old C reactor and IO code was to update connection properties to generate the close frame and when there was no remaining possibility of blocking operations to run the reactor to clean up all resources.  Some previous bug fix required stopping the reactor early and allowing the C object reference counting to promote the cleanup.

With the new Python IO, the strategy is the same but slightly different mechanisms are required.

This patch:

  - does a Transport.close_tail to prevent blocking waiting for the peer close frame (which may never be sent)

  - removes the early stop() so that Reactor cleanup can complete, especially Selectable and socket closing.

  - prevents mistaking a local close with an external disconnect.

python/proton/_utils.py
```diff
@@ -440,11 +440,16 @@ def close(self):
...
+                    self.conn.transport.close_tail()
         finally:
             self.conn.free()
             # Nothing left to block on.  Allow reactor to clean up.
             self.run()
-            self.conn = None
+            if self.conn:
+                self.conn.handler = None  # break cyclical reference
+                self.conn = None
             self.container.global_handler = None  # break circular ref: container to cadapter.on_error
```
astitcher, line 453: This is probably no longer necessary (according to the comment) as we aren't using cadapter.on_error any more.

astitcher (approved):

cliffjansen:
Fixes applied, including Andrew's suggestion.

PR #310, opened 2021-04-22
PROTON-2377: [cpp] Make tag_counter in sender.cpp atomic

DreamPearl:
Issue: https://issues.apache.org/jira/browse/PROTON-2377


cpp/include/proton/delivery.hpp
```diff
@@ -64,6 +65,9 @@ class delivery : public transfer {
     /// Settle with MODIFIED state.
     PN_CPP_EXTERN void modify();
 
+    /// Return the tag for this delivery.
+    PN_CPP_EXTERN binary tag() const;
```
jiridanek, line 69: You should rebase your PR branch on top of `main`, to get rid of this (already merged) code. Or you can just delete the commit from the branch; whichever you find simpler.

cpp/src/sender.cpp
```diff
@@ -60,14 +61,20 @@ class target sender::target() const {
 }
 
 namespace {
-// TODO: revisit if thread safety required
-uint64_t tag_counter = 0;
+std::atomic<uint64_t> tag_counter(0);
```
jiridanek, line 64: This would of course work. I was also thinking about maybe making the counter be an instance variable of a `sender`. Let's see what others think first.

cpp/include/proton/delivery.hpp
DreamPearl: Just did that. Thanks.

cpp/src/sender.cpp
gemmellr: I would probably lean toward a sender-specific value as well, though both ways have their own advantages.

@astitcher ? @ssorj ?

astitcher: It would make a lot of sense (from a performance stand point) to have the tag counter be scoped to the connection in some way (as this is the serialisation granule of proton).
Is it true that the delivery tags are only scoped to the sender then. So that different senders can use identical tags? if so we should just make this connection or more granularly scoped and just zero out the count when we create a new connection/session/sender/etc.

astitcher: To be clear if we can make the tag counter scoped to the connection or lower then we don't need any thread synchronisation at all. This is because it is an underlying assumption in proton that a connection can only be processed on a single thread at a time.

gemmellr: > Is it true that the delivery tags are only scoped to the sender then. So that different senders can use identical tags?

Yes, tags are only used in the scope of a particular link, so they only need to be unique for the outstanding set of unsettled deliveries on that link:

"The delivery-tag MUST be unique amongst all deliveries that could be considered unsettled by either end of the link. "
http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#doc-idp438000

gemmellr: >To be clear if we can make the tag counter scoped to the connection or lower then we don't need any thread synchronisation at all. This is because it is an underlying assumption in proton that a connection can only be processed on a single thread at a time.

Yep, thats what pompted my original reply on the other PR, when I thought the variable was sender-specific, before it was pointed out it was actually a global variable.

DreamPearl: I've made the changes. Can you please check if this is what you meant?

jiridanek: It's what I meant. Let's wait what Andrew thinks, as the cpp binding maintainer...

astitcher (approved):
I'm happy enough with this change - I was thinking that it might be better to have a connection specific prefix to the tag for the purposes of debugging. But on the whole this is very simple and correct so we should go with this. If we need a better debug experience then we can change the implementation.

PR #313, opened 2021-05-16
PROTON-2323 Propagate requested version from FindPython to FindPythonInterp

jiridanek:
Not immediately useful in Proton. This is to keep the module in sync with Dispatch code where it is duplicated.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,7 +42,7 @@ endif ()
 if (NOT DEFINED Python_FIND_FRAMEWORK)
   set(Python_FIND_FRAMEWORK "LAST")
 endif ()
-find_package(Python
+find_package(Python 3.6.8
         REQUIRED COMPONENTS Interpreter
         OPTIONAL_COMPONENTS Development)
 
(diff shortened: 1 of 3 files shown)


jiridanek:
Thinking about this, I think that Proton should require Python 3.6.8 in the CMakeLists. It's true that the bindings can be built just fine when only Python 2 is available, but 1) it's limiting --- some python scripts have to stay py2 compatible and it is necessary to keep track of which is which 2) it makes it easier for the users who want to use the Python binding; Proton will pick up the compatible python by default (on systems where both 2 and  3 are available).

CMakeLists.txt
```diff
@@ -42,7 +42,7 @@ endif ()
 if (NOT DEFINED Python_FIND_FRAMEWORK)
   set(Python_FIND_FRAMEWORK "LAST")
 endif ()
-find_package(Python
+find_package(Python 3.6.8
```
astitcher, line 45: Just 3.6 is consistent with are stated requirements

astitcher (changes requested):
I'm a little skeptical about the travis changes, because I'd really really prefer to use venvs everywhere. But lets just get this in for now - note my commnet about the required minimum being 3.6. But with that change just merge it.

CMakeLists.txt
jiridanek, line 45:
```suggestion
find_package(Python 3.6
```

PR #314, opened 2021-05-18
PROTON-2220 Regenerate test certificates with long expiry

jiridanek:


jiridanek:
I checked the generated cert and this did not work out as I intended. I will try again. At least, we have the ci result.

jiridanek:
Ok, now it should be there. Expiry in 100 years. Let the singularity robo machines that replace human civilization by that time deal with the maintenance then.

PR #315, opened 2021-05-24
PROTON-2389 Update Catch to Catch2

jiridanek:


c/tests/data_test.cpp
```diff
@@ -124,6 +124,7 @@ static void check_encode_decode(auto_free<pn_data_t, pn_data_free>& src) {
 	// Decode buf to data
 	int dec_size = pn_data_decode(data, buf, BUFSIZE - 1);
 	pn_error_t *dec_err = pn_data_error(data);
+	CHECK(pn_error_code(dec_err) == 0);
```
jiridanek, line 127: Unused variable error, there is multiple cases like this. I am surprised that compiler did not error on this before.

jiridanek:
@astitcher I wanted to have another look at this. i committed this right before going to sleep, actually. One thing that I'd want to understand are the previously unused variables. How come the tests compiled before just fine?

astitcher (approved):
This all looks good to me.
I wouldn't worry about too much about why the compiler didn't pick up on unused vars before - it could be a s simple as optimisation level.

c/tests/pn_test.hpp
```diff
@@ -129,7 +130,7 @@ struct handler {
 struct driver : public ::pn_connection_driver_t {
   struct handler &handler;
 
-  driver(struct handler &h);
+  explicit driver(struct handler &h);
```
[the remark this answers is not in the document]
jiridanek: I got it from clang-tidy and https://google.github.io/styleguide/cppguide.html#Implicit_Conversions. The widely accepted rule seems to put `explicit` everywhere, unless there is a very good reason not to. (And I can do this freely here, because I am not touching public API.)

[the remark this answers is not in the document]
astitcher: Ah, I'd forgotten the new iinitialisation syntax thing - that makes this much more compelling as a rule. It's definitely the better default anyway - but also makes conversion operators safer (even if it would make for widespread changes!).
Making this change to the exported include files would change the API, so we can't do it easily there (but actually not the ABI I think)

c/tests/pn_test.hpp
```diff
@@ -46,13 +46,14 @@ namespace pn_test {
...
+public:
+  auto_free(auto_free &x) = delete;
 
 public:
-  auto_free(T *p = 0) : ptr_(p) {}
+  explicit auto_free(T *p = 0) : ptr_(p) {}
   ~auto_free() { Free(ptr_); }
   T *get() const { return ptr_; }
-  operator T *() const { return ptr_; }
+  operator T *() const { return ptr_; } // not marking explicit for convenience
```
jiridanek, line 56: This class is something like an `std::unique_ptr` with a deleter. Apparently having explicit `operator*` was considered and dismissed, for reasons https://herbsutter.com/2012/06/21/reader-qa-why-dont-modern-smart-pointers-implicitly-convert-to/

astitcher: Actually it could be replaced with unique_ptr now - this version will work with C++03 which isn't needed any more. [std::unique_ptr has an optional deleter capability]

astitcher: Nut not having explicit here is the opposite of the advice from Herb Sutter I think.

jiridanek: Ok, I edited my comment too late, now the conversation does not make sense. Revert!

astitcher: Honestly I'm fine with putting this PR in as is and then moving it closer to good C++11 piece-by-piece

PR #317, opened 2021-05-26
PROTON-2254 Use more modern CMake install machinery for better config.cmake

jiridanek:
This supersedes #312 and allows building static Dispatch on CentOS 7.

* Build Proton with `-DCMAKE_POSITION_INDEPENDENT_CODE=ON -DBUILD_STATIC_LIBS=ON -DBUILD_BINDINGS=OFF -DCMAKE_INSTALL_PREFIX=install`
* Build Dispatch with `-DProton_DIR=qpid-proton/build/install/lib64/cmake/Proton -DProton_USE_STATIC_LIBS=ON`

This version will be hopefully upstreamable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -276,21 +276,17 @@ set (MAN_INSTALL_DIR share/man CACHE PATH "Manpage directory")
 
 mark_as_advanced (INCLUDE_INSTALL_DIR LIB_INSTALL_DIR SYSCONF_INSTALL_DIR SHARE_INSTALL_DIR MAN_INSTALL_DIR)
 
-macro (pn_absolute_install_dir NAME VALUE PREFIX)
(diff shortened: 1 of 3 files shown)


jiridanek:
With DISPATCH-2178 merged, Dispatch should no longer need to build position-independent static libraries here.

jiridanek:
This seems to work at this point. There is one little problem. In `version.h`, there is `#define PN_INSTALL_PREFIX` that is used for deriving path to the json connect config file. The path there ends up being absolute.

The most reasonable anchor point to make it relative is probably the location of proton .so library; however that would not work for static builds, and looking up the lib is quite messy. So I propose to ignore this one remaining instance of absolute paths in the install.

I got a review on this on https://github.com/apache/qpid-proton/pull/312 already, and I fixed there the objection (there is no longer any PIC present; but it can be still opted in on CMake command line if needed).

PR #318, opened 2021-06-04
PROTON-2095 Move away from SWIG to CFFI

jiridanek:


astitcher:
I'm really excited to see someone working to make the python bindings based on cffi rather than swig. I think this will bring significant benefits to Proton in build simplicity and dependencies.

Looking at these changes I have a few comments:

1. There seem to be many gratuitous spacing changes, which really detract from being able to see what is really changed to support the move to CFFI and things that are arguably a matter of taste or at least non functional. I'm not at all against these changes I just think it's important to separate them so that the actual functional changes are 'front and centre' and can be celarly reviewed.

2. I think the layering in this code is wrong:
I think it is a requirement to keep the existing `pn_..` interfaces unchanged or as unchanged as possible. I think that the cffi binding should be a drop in replacement for `import cproton` as there a number of places where this is used: Not least currently to test the c code itself. This means I expect minimal changes to the existing python code which imports cproton and uses it.
I definitely don't expect `ffi `types (`ffi.null` for example) to be present in the major python binding code.

jiridanek:
> I think that the cffi binding should be a drop in replacement for import cproton as there a number of places where this is used.

My thinking was that the `cproton` module is internal API that does not have to be preserved. The only external use of the API that I am aware of are these two tests in Qpid Dispatch, https://github.com/apache/qpid-dispatch/search?l=Python&q=cproton

It does sound plausible to me that your suggested approach would mean less work and less chance of error, though. I will discuss this with @SudhanAruna more.

jiridanek:
```
# cproton.py

import lib

import lib.pn_unchanged_function

def pn_str(a):
    return ffi.string(lib.pn_str(a)).decode()
    
__all__ = [
    'pn_unchanged_function',
    'pn_str',
]
```

for future reference

astitcher:
> > I think that the cffi binding should be a drop in replacement for import cproton as there a number of places where this is used.
> 
> My thinking was that the `cproton` module is internal API that does not have to be preserved. The only external use of the API that I am aware of are these two tests in Qpid Dispatch, https://github.com/apache/qpid-dispatch/search?l=Python&q=cproton

I think there is some confusion here about internal vs external use of the API. The testing I'm thinking of is indeed inside the python test suite, but these tests actually test the c code not the python code.
In any case, My point is about better layering for the python library. Which will affect the maintainability of the code more than anything else. So it'll be easier to debug/modify and generally evolve over time if it has better layering:
Keeping the details of ffi outside the top level python binding code will it seems to me make it much easier to evolve this code over time in many ways.

jiridanek:
Continued in #318.

PR #319, opened 2021-06-08
[DRAFT] [cpp] Add distributed tracing using OpenTelemetry

DreamPearl:
[PROTON-2487](https://issues.apache.org/jira/browse/PROTON-2487)

- Add the opentelemetry-cpp library dependencies
- Add opentelemetry-cpp install/build steps in GHA
- Add tracing.cpp file
  - Add tracing spans for messaging operations
  - Add basic attributes[delivery_tag, delivery_state, source_address, destination_address]
- Add a	tracing	demo file in examples
- Add a test file

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,6 +4,7 @@ on: [push, pull_request]
 
 jobs:
   build:
+    if: github.repository == 'DreamPearl/qpid-proton'
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
@@ -24,6 +25,12 @@ jobs:
       InstallPrefix: ${{github.workspace}}/INSTALL
       PKG_CONFIG_PATH: ${{matrix.pkg_config_path}}
       VCPKG_DEFAULT_TRIPLET: x64-windows
+      OTelDir: ${{github.workspace}}/opentelemetry-cpp
+      OTelBuildDir: ${{github.workspace}}/opentelemetry-cpp/BLD
+      OTelInstallPrefix: /usr
+      CurlDir: ${{github.workspace}}/curl
+      CurlBuildDir: ${{github.workspace}}/curl/BLD
+      CurlInstallPrefix: /usr
     steps:
     - uses: actions/checkout@v2
     - name: Create Build and Install directories
@@ -60,9 +67,59 @@ jobs:
       if: runner.os == 'macOS'
       run: |
         brew install libuv swig pkgconfig jsoncpp
+    - name: Checkout curl repo (for jaeger)
+      if: runner.os == 'Linux'
+      uses: actions/checkout@v2
+      with:
+        repository: curl/curl
+        path: ${{env.CurlDir}}
+        submodules: recursive
+    - name: Create Build directory for curl
+      run: mkdir -p "${CurlBuildDir}"
+      shell: bash
+    - name: curl cmake configure
+      if: runner.os == 'Linux'
+      working-directory: ${{env.CurlBuildDir}}
+      run: cmake "${{github.workspace}}/curl" "-DCMAKE_INSTALL_PREFIX=${CurlInstallPrefix}" "-DBUILD_SHARED_LIBS=ON" "-DCMAKE_POSITION_INDEPENDENT_CODE=ON" "-DBUILD_TESTING=OFF" "-DCMAKE_BUILD_TYPE=${BuildType}"
+      shell: bash
+    - name: curl cmake build/install
+      if: runner.os == 'Linux'
+      run: |
+        cmake --build "${CurlBuildDir}" --target all
+        sudo cmake --install "${CurlBuildDir}" --config ${BuildType}
+      shell: bash
+    - name: install thrift (for jaeger)
+      if: runner.os == 'Linux'
+      run: sudo ./ci/setup_thrift.sh
+      shell: bash
+    - name: Create directory for opentelemetry-cpp
+      run: mkdir -p "${OTelDir}"
+      shell: bash
+    - name: Checkout opentelemetry-cpp repo
+      if: runner.os == 'Linux'
+      uses: actions/checkout@v2
+      with:
+        repository: open-telemetry/opentelemetry-cpp
+        path: ${{env.OTelDir}}
+        submodules: recursive
+    - name: Create Build directory for opentelemetry-cpp
+      if: runner.os == 'Linux'
+      run: mkdir -p "${OTelBuildDir}"
+      shell: bash
+    - name: opentelemetry-cpp cmake configure
+      if: runner.os == 'Linux'
+      working-directory: ${{env.OTelBuildDir}}
+      run: cmake "${{github.workspace}}/opentelemetry-cpp" "-DCMAKE_INSTALL_PREFIX=${OTelInstallPrefix}" "-DBUILD_SHARED_LIBS=ON" "-DCMAKE_POSITION_INDEPENDENT_CODE=ON" "-DBUILD_TESTING=OFF" "-DCMAKE_BUILD_TYPE=${BuildType}" "-DWITH_JAEGER=ON"
+      shell: bash
+    - name: opentelemetry-cpp cmake build/install
+      if: runner.os == 'Linux'
+      run: |
+        cmake --build "${OTelBuildDir}" --target all
+        sudo cmake --install "${OTelBuildDir}" --config ${BuildType}
+      shell: bash
     - name: cmake configure
       working-directory: ${{env.BuildDir}}
-      run: cmake "${{github.workspace}}" "-DCMAKE_BUILD_TYPE=${BuildType}" "-DCMAKE_INSTALL_PREFIX=${InstallPrefix}" ${{matrix.cmake_extra}}
+      run: cmake "${{github.workspace}}" "-DCMAKE_BUILD_TYPE=${BuildType}" "-DCMAKE_INSTALL_PREFIX=${InstallPrefix}" ${{matrix.cmake_extra}} "-Dopentelemetry-cpp_DIR=${OTelInstallPrefix}/lib/cmake/opentelemetry-cpp"
       shell: bash
     - name: cmake build/install
       run: cmake --build "${BuildDir}" --config ${BuildType} -t install
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -19,10 +19,10 @@
 
 include(CMakeDependentOption)
 enable_language(CXX)
+include(CMakePrintHelpers)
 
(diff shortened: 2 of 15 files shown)


cpp/CMakeLists.txt
```diff
@@ -36,6 +36,14 @@ else()
   set(CONNECT_CONFIG_SRC src/connect_config_dummy.cpp)
 endif()
 
+# Check for OPENTELEMETRY-CPP for distributed tracing
+find_package(opentelemetry-cpp)
+cmake_dependent_option(ENABLE_OPENTELEMETRYCPP "Use opentelemetry for distributed tracing" ON "OPENTELEMETRY-CPP_FOUND" OFF)
+
+if (ENABLE_OPENTELEMETRYCPP)
+  include_directories(${OPENTELEMETRY_CPP_INCLUDE_DIRS})
```
jiridanek, line 44: This is "old-style" CMake. Nowadays, you should use `target_link_libraries` (https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#transitive-usage-requirements) later when you compile something that needs OpenTelemetry. Adding an OpenTelemetry-exported target this way will also add the include directories to the compilation.

If you need to adjust includes yourself (which should not be necessary here), you should use `target_include_directories` (https://cmake.org/cmake/help/v2.8.12/cmake.html#command:target_include_directories).

These things are available even in CMake 2.8.12 which we need to support in Proton.

# Docs

Here are some documents that introduce "Modern CMake". If these things can be used in CMake 2.8.12, then they should be used.

* https://cliutils.gitlab.io/modern-cmake/
* https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1

DreamPearl: ACK

.github/workflows/build.yml
```diff
@@ -60,6 +63,28 @@ jobs:
       if: runner.os == 'macOS'
       run: |
         brew install libuv swig pkgconfig jsoncpp
+    - uses: actions/checkout@v2
```
gemmellr, line 66: This looks redundant, duplicating the original proton checkout, with cleanup side effects.

.github/workflows/build.yml
```diff
@@ -60,6 +63,28 @@ jobs:
...
+    - name: Create directory for opentelemetry-cpp
+      run: mkdir -p "${OTelDir}"
+      shell: bash
+    - name: Checkout opentelemetry-cpp repo
+      uses: actions/checkout@v2
+      with:
+        repository: open-telemetry/opentelemetry-cpp
+        path: ${{env.OTelDir}}
+    - uses: actions/checkout@v2
```
gemmellr, line 75: This looks redundant, duplicating the original proton checkout, with cleanup side effects.

gemmellr: To be clearer, not sure why its highlighting all 4 lines...I just mean the last basic "- uses: actions/checkout@v2"

DreamPearl: Thanks, Done! Now it seems to be working. :)
 
And I just found out, that if we provide a directory beforehand then git clone will just dump data inside of it rather than creating a new directory. 
I manually created a directory for opentelemetry and thought git clone will also create a new subdirectory inside the one that I've created ( i.e. opentelemetry/opentelemetry-cpp). That's why it wasn't able to find CMakeLists file. :(

cpp/CMakeLists.txt
```diff
@@ -37,7 +37,7 @@ else()
 endif()
 
 # Check for OPENTELEMETRY-CPP for distributed tracing
-find_package(opentelemetry-cpp)
+find_package(opentelemetry-cpp CONFIG REQUIRED)
```
jiridanek, line 40: Is it going to be `REQUIRED`? I am all for it (less possible build configurations), but I imagine some users might want to build Proton without opentelemtry cpp.

gemmellr: I would doubt it generally, but especially since its not yet even in a released state. I had suggested on IRC shortly before your comment the next step would be to make it optional.

jiridanek: should be enough to delete the `REQUIRED`. There already is the `cmake_dependent_option` thing for this, for dealing with the optionality below in the file.

gemmellr: Yep, I noted that it mostly seemed optional already apart from the find.

DreamPearl: Done! I've deleted the `REQUIRED`. 
But this time I've changed the `cmake_dependent_option` to `option`. `option` will also work right?

gemmellr: I'll need to defer to the folks with cmake experience at this point to suggest on the specific approach, I dont actually know myself.

cpp/src/tracing.cpp
```diff
@@ -0,0 +1,240 @@
...
+    trace::Scope scope = get_tracer()->WithActiveSpan(span);
+
+    if (delivery_state == "unknown")
+    {
+        span->AddEvent("Delivery state is unknown");
+    }
+    if (delivery_state == "Received")
+    {
+        span->AddEvent("Delivery is received");
+    }
```
gemmellr, line 178: A delivery state is unlikely to ever be known at this point (creating the span, before firing on_message), the client or application will typically set it once they are given the message and finish processing it. That may be before on_message returns if they do it explicitly, or they may have the 'auto-accept' enabled (it is by default) and it is done by the client just after on_message returns, or they may disable that and do it manually later after some long processing completes.

For the JMS client the same selection of possibilities (and others) are true, and we cant easily access the state or span later anyway (and cant leave it open forever either, and need to create a new active span for the next message delivery). We also have cases there that the proton clients do not, where we specifically *dont* deliver the message to the application.

So what I did was, when onMessage returns (or the message wasnt given to onMessage at all) and I am completing the delivery span, I simply logged a summary event relating to the onMessage invocation, not the literal delivery state as it may well not be known yet, adding a description of what happened with e.g 'event = DELIVERED' (or EXPIRED, REDELIVERIES_EXCEEDED...or APPLICATION_ERROR if they throw an exception from onMessage).

cpp/src/tracing.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+    transfer tt(track);
+    std::string delivery_state = to_string(tt.state());
+
+    nostd::shared_ptr<trace::Span> span = get_tracer()->StartSpan(
+        "amqp-delivery-send",
+        {{"Delivery_tag", delivery_tag}, {"Destination_address", t_addr}},
+        options);
+
+    trace::Scope scope = proton::get_tracer()->WithActiveSpan(span);
```
gemmellr, line 221: I dont think the send span should be set as the 'active' span, as we dont want it to affect other things (e.g replacing an on_message active span), and/or the life of the scope then seems like it would be in play. We just want to create and retain a span and close it once the message is accepted/other later.

cpp/src/tracing.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+    if (delivery_state == "unknown") {
+        span->AddEvent("Delivery state is unknown");
+    }
```
gemmellr, line 225: The send state will essentially always be unknown at this point. It becomes known when the disposition arrives from peer (e.g broker) setting the applicable state (for the messages that are not sent pre-settled).

Wherever the message 'tracker' is being updated (perhaps [1]) upon disposition is the point the delivery state will become known for send, and where the send span should be ended.

For pre-settled sends that may be different, as they will never get updated state due to being settled in advance. Depending on handling of the events, their spans may even need completed within the send call itself.

[1] https://github.com/apache/qpid-proton/blob/0.35.0/cpp/src/messaging_adapter.cpp#L157-L177

cpp/src/messaging_adapter.cpp
```diff
@@ -48,6 +48,8 @@
 
 #include <assert.h>
 
+#include <proton/tracing.hpp>
```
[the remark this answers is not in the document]
DreamPearl: Done, now looks like the dummy file is linking correctly!

Now build is passing and I've also checked for when opentelemetry is not present by skipping otel build steps in gha build.

.github/workflows/build.yml
```diff
@@ -4,6 +4,7 @@ on: [push, pull_request]
 
 jobs:
   build:
+    if: github.repository == 'DreamPearl/qpid-proton'
```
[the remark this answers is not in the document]
DreamPearl: ack!

cpp/src/tracing.cpp
DreamPearl: @gemmellr  Ah, now it makes sense to me. I've made few changes, please let me know if I get it right.

Now I am ending the send_span where the message tracker is being updated and before ending; adding delivery state as an event (Now send_span delivery state is "accepted" for the demo file).

cpp/src/tracing.cpp
gemmellr: As discussed on IRC, it seems that in the opentelemetry-cpp case, the returned 'scope' being destroyed will reinstate the previously active span as the active span, so as that destruction seems to happen before send returns then it will essentially behave as I wanted in terms of not leaving the new span as the 'active'.

Setting the span as active in this case seems largely redundant since we have no other related spans for it to impact and simply inject the context and send the messaage, but if thats required then so be it. It was the knock on effect on the 'active' I was concerned with.

cpp/examples/tracing_demo.cpp
```diff
@@ -0,0 +1,101 @@
...
+        std::string conn_url = argc > 1 ? argv[1] : "//127.0.0.1:5672";
+        std::string addr = argc > 2 ? argv[2] : "examples";
+
+        opentelemetry::exporter::jaeger::JaegerExporterOptions opts;
+        auto exporter = std::unique_ptr<sdktrace::SpanExporter>(
+            new opentelemetry::exporter::jaeger::JaegerExporter(opts));
+        // auto exporter = std::unique_ptr<sdktrace::SpanExporter>(
+        //     new opentelemetry::exporter::trace::OStreamSpanExporter());
+
+        proton::initTracer(std::move(exporter));
```
DreamPearl, line 91: Should the exporter be user-dependent or should I hardcode it to JAEGER EXPORTER only?

gemmellr: For an example I think it would be ok to do it either way. The rest of the example seems to be aimed toward using Jaeger so I think using that would be fine.

@ssorj @astitcher thoughts ?

DreamPearl:
Continuing the changes in the new PR:  https://github.com/apache/qpid-proton/pull/355

DreamPearl:
PR https://github.com/apache/qpid-proton/pull/355 Merged. Closing this.

PR #322, opened 2021-06-24
PROTON-2399 Support Python 3.10

jiridanek:


gemmellr:
The first commit needs dropped now its effect was added to main, and the end brace is duplicated in setup.py.in, the one after 3.9 should be removed.

jiridanek:
Looks safe to merge to me, and Python 3.10 is well on its way to be released. I think it is reasonable to add this for 0.36.0.

PR #324, opened 2021-06-25
PROTON-2402: Clarify proton docs for on_sendable, prefetch for Messag…

NS-SlaFleur:
…ingHandler, send for Sender and thread-safety for Container.

https://issues.apache.org/jira/browse/PROTON-2402


NS-SlaFleur:
@NS-MNellen will take over this PR if there are any questions.

astitcher:
@ssorj Could you look at these doc changes and see if you think this is an improvement.

PR #328, opened 2021-08-06
Draft PR of my progress in replacing the frame codec with generated code

astitcher:
This does now pass all the in tree unit tests. On my fairly cursory testing with quiver this gives approx a x1.7 speed up.


astitcher:
Closing this draft in favour of #339 which has the nearly final set of changes

PR #332, opened 2021-08-14
PROTON-2407 [python] fix type-related documentation

jiridanek:


python/proton/_handlers.py
```diff
@@ -322,74 +322,64 @@ def __init__(self, peer_close_is_error: bool = False, delegate: Optional[Handler
...
         :const:`proton.Endpoint.LOCAL_ACTIVE`).
 
         :param endpoint: The local endpoint to be tested.
-        :type endpoint: Any child of :class:`proton.Endpoint`
         :return: ``True`` if local endpoint is in state
             :const:`proton.Endpoint.LOCAL_ACTIVE`, ``False`` otherwise.
-        :rtype: ``bool``
         """
-        return endpoint.state & Endpoint.LOCAL_ACTIVE
+        return bool(endpoint.state & Endpoint.LOCAL_ACTIVE)
```
jiridanek, line 334: The return type used to be `int`; I think in this instance it makes sense to adjust the code to match the docs, rather than the other way around.

python/proton/_endpoints.py
```diff
@@ -1203,11 +1201,8 @@ def send(self, obj: Union[bytes, 'Message'], tag: Optional[str] = None) -> Union
             # treat object as bytes
             return self.stream(obj)
 
-    def delivery_tag(self):
-        """
-        **HELP!** I have no idea what is going on with this!
-        :meth:`Link.next` returns a Link, not a Delivery.
-        """
```
jiridanek, line 1210: This is the `next` builtin, not Link.next.

python/proton/_data.py
```diff
@@ -1158,7 +1154,7 @@ def get_map(self) -> int:
         """
         return pn_data_get_map(self._data)
 
-    def get_array(self):
+    def get_array(self) -> Tuple[int, bool, Optional[int]]:
```
jiridanek, line 1157:
```suggestion
    def get_array(self) -> Optional[Tuple[int, bool, int]]:
```

hmm, possibly the correct type is actually this; in any case, what the doc used to say was wrong

jiridanek: actually, ^^^ cannot be right

PR #333, opened 2021-09-01
PROTON-2422: Proton will sometimes fail to send empty frame if the idle timeout ratio between peers is greater than 2. 

samacleese:
https://issues.apache.org/jira/browse/PROTON-2422


astitcher:
This is now obsoleted by the recent different fix to PROTON-2422

PR #334, opened 2021-09-17
PROTON-2254, PROTON-2433 Fixes to 'modern' CMake installed targets

jiridanek:


astitcher (approved):
LGTM (Assuming it fixes the problem!)
- also neat and tidy still!

jiridanek:
This is just something I was able to test now. For the real fix, I'd still want to remove the `INCLUDEDIR` cmake variable, so that I am not leaving unneeded things around. The 'nonmodern' `*Config.cmake` variables `*_INCLUDE_DIRS` should extract the paths directly from the modern target, the way it is done for `*_LIBRARIES`.

astitcher:
I think this may need configure_package_config_file(... PATHVARS INCLUDEDIR...) and using @PACKAGE_INCLUDEDIR@ inside the .in file to get the potentially relocated var. But this isn't exactly wht you meant - it should just give the correct result in all cases (according to the doc!)

But merge what you just did and I'll be on my merry way! Thanks.

jiridanek:
all right, then

astitcher:
Also afaict the LIBRARIES var is much easier because it just uses the modular name in the directly (eg Proton::core) and doesn't have to worry about actual file paths etc.

jiridanek:
> I think this may need configure_package_config_file(... PATHVARS INCLUDEDIR...) and using @PACKAGE_INCLUDEDIR@ inside the .in file to get the potentially relocated var.

This works for `cmake.in` files, but not for `pkgconfig.in`. The way cmake does the relocation is that it emits cmake code which sets helper variables, that are then used in the @PACKAGE_INCLUDEDIR@.

I have a reasonably good solution for this, but my last problem is that Irina's RPM specs pass in absolute paths for the libdir and includedir, which I need to relativize for use in pkgconfig, like this

Screenshot_20210919_102710

Hopefully string replacement (replace prefix, take care not to create absolute path by mistake) will take care of that.

https://indico.cern.ch/event/848215/contributions/3591953/attachments/1923018/3181752/HSFPackagingRelocation.pdf

astitcher:
> I have a reasonably good solution for this, but my last problem is that Irina's RPM specs pass in absolute paths for the libdir and includedir, which I need to relativize for use in pkgconfig, like this

Well you could just ask her to change the spec files!

PR #337, opened 2021-10-25
PROTON-2406: [cpp] Replace uses of deprecated reconnect_options::failover_urls with conn…

DreamPearl:
[PROTON-2406](https://issues.apache.org/jira/browse/PROTON-2406)

First, I wanted to reproduce the bug to make sure I understand the issue properly.
It would also help me to verify my changes made, now I can see the warnings are really gone.

Steps I took to reproduce the bug:

- Remove "-DPN_CPP_USE_DEPRECATED_API=1" from cpp/CMakeLists to actually deprecate the targeted lines.
- Add "remove_definitions(-Werror)" so that warnings are not treated as errors; to allow build not to fail in the middle.

### DIFF
--- a/cpp/examples/reconnect_client.cpp
+++ b/cpp/examples/reconnect_client.cpp
@@ -54,10 +54,8 @@ class reconnect_client : public proton::messaging_handler {
   private:
     void on_container_start(proton::container &c) override {
         proton::connection_options co;
-        proton::reconnect_options ro;
 
-        ro.failover_urls(failovers);
-        co.reconnect(ro);
+        co.failover_urls(failovers);
         c.connect(url, co);
     }
 
--- a/cpp/include/proton/connection_options.hpp
+++ b/cpp/include/proton/connection_options.hpp
@@ -197,6 +197,13 @@ class connection_options {
     /// If the connection to the primary connection url fails then try each of the fail-over
     /// connection urls in turn.
     ///
+    /// If this option is set then it will by default set the default reconnect timing options.
+    ///
+    /// If this option is changed using @ref connection::update_options before a reconnection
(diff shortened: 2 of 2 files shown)


cliffjansen:
As I read the documentation for reconnect_url() and failover_urls() I implicitly expect the same intention for:

  - default reconnect timing options
  - change via update_options()

even though it is only explicitly spelled out for the former and may not make sense in all permutations for the latter.  Perhaps the documentation should be made explicit for failover_urls().

Regardless, I think I would expect this example to have no reference to reconnect_options unless non-default values were being set and no call made to connection::reconnect().

The segfault appears to be from a missing argument where argc is one short to catch the expected number of input parameters.  However, it may also be a failing in the library to not robustly handle an empty vector properly.

DreamPearl:
> Regardless, I think I would expect this example to have no reference to reconnect_options unless non-default values were being set and no call made to connection::reconnect().

@cliffjansen Yep, fixed!
> The segfault appears to be from a missing argument where argc is one short to catch the expected number of input parameters. However, it may also be a failing in the library to not robustly handle an empty vector properly.

The segfault fix PR is now merged [PROTON-2441](https://github.com/apache/qpid-proton/pull/338). It was failing in the library, connection_options::failover_urls wasn't able to handle an empty vector.

jiridanek (approved):
lgtm!

DreamPearl:
Updated the documentation for failover urls regarding default reconnect timing options and change via update_options(). @astitcher @cliffjansen

astitcher:
This change looks good.
However the commit comment doesn't include the JIRA reference.
Please edit the commit message and I'll merge this.

DreamPearl:
> This change looks good. However the commit comment doesn't include the JIRA reference. Please edit the commit message and I'll merge this.

Done!

PR #338, opened 2021-10-29
PROTON-2441: [cpp] Fix connection_options failover urls segfault

DreamPearl:
Issue: [PROTON-2441](https://issues.apache.org/jira/browse/PROTON-2441)

### DIFF
--- a/cpp/src/connection_options.cpp
+++ b/cpp/src/connection_options.cpp
@@ -124,7 +124,7 @@ class connection_options::impl {
             cc.reconnect_url_ = reconnect_url.value;
             cc.reconnect_context_->current_url_ = -1;
         }
-        if (failover_urls.set) {
+        if (failover_urls.set && !failover_urls.value.empty()) {
             failover_urls.set = false;
             cc.failover_urls_ = failover_urls.value;
             cc.reconnect_context_->current_url_ = 0;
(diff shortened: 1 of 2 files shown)


DreamPearl:
@gemmellr @jiridanek Can you please take a look. 
The cause of segfault was that when we were passing empty vector to connection_options::failoverurls, the current_url was getting set to 0. And on the basis of the value of current_url i.e. 0, it was picking failover_urls instead of reconnect_urls. As it was trying to access failover_urls which was empty, it leads to segfault.
Deprecated code was able to handle the empty vector, non-deprecated wasn't.
Alternatively, I tried this: https://github.com/apache/qpid-proton/commit/50d8342c4932da698389dd991760c607de7f79f5

gemmellr:
The change itself seems reasonable to me, it looks similar to what the other bit a little above it is doing.

I know Jiri thought, and I agree, that ensuring a test of some kind covers this would be good since obviously nothing existing covered this except the example, and so still doesnt with this as-is. Plus that it might also be good to protect the spot where the segfault actually occurred, rather than just the outer segment where the option is passed through to it, so no other routes can hit similar issue in future.

DreamPearl:
@gemmellr 
> I know Jiri thought, and I agree, that ensuring a test of some kind covers this would be good since obviously nothing existing covered this except the example, and so still doesnt with this as-is.

I have written a test to capture the concerning bug if that exists.
> Plus that it might also be good to protect the spot where the segfault actually occurred, rather than just the outer segment where the option is passed through to it, so no other routes can hit similar issue in future.

Done.

astitcher:
I've merged this PR with a bug fix to an connection leak in the new reconnect test - this leak can be seen in the failing Travis CI job - I think it's unfortunate that the github actions didn't pick this up as well - are they running without valgrind?

astitcher:
@DreamPearl  also note that I subsequently modified your new test in commit 56520fbaa00ab1d64b9672b072f7b8b2e88c1916
to deduplicate the code and also to use more C++11 there now that we can. If you have questions about any of this just ask.

PR #340, opened 2021-11-03
PROTON-2396: Use random_device for seed initialization in uuid.cpp

DreamPearl:
[PROTON-2396](https://issues.apache.org/jira/browse/PROTON-2396)

### DIFF
--- a/cpp/src/uuid.cpp
+++ b/cpp/src/uuid.cpp
@@ -22,10 +22,15 @@
 #include "proton/uuid.hpp"
 #include "proton/types_fwd.hpp"
 
+#include <algorithm>
+#include <chrono>
+#include <climits>
 #include <cstdlib>
 #include <ctime>
-#include <sstream>
+#include <functional>
 #include <iomanip>
+#include <random>
+#include <sstream>
 
 #if defined(_WIN32)
 #include <process.h>
@@ -39,19 +44,27 @@ namespace proton {
 
 namespace {
 
-
-// Seed the random number generated once at startup.
 struct seed {
-    seed() {
-        // A hash of time and PID, time alone is a bad seed as programs started
-        // within the same second will get the same seed.
-        unsigned long secs = time(0);
-        unsigned long pid = GETPID();
-        std::srand(((secs*181)*((pid-83)*359))%104729);
-    }
-} seed_;
 
-}
+    std::independent_bits_engine<std::mt19937, CHAR_BIT, unsigned int> engine;
+
+    seed() {
+        // A hash of time, PID and random_device to produce an unsigned integer
+        // for seeding the engine.
+        unsigned long ticks =
+            std::chrono::system_clock::now().time_since_epoch().count();
+        unsigned long pid = GETPID();
+        unsigned int rd = std::random_device{}();
+
+        std::seed_seq seq{ticks, pid, (unsigned long)rd};
+        engine =
+            std::independent_bits_engine<std::mt19937, CHAR_BIT, unsigned int>(
+                seq);
+    }
+
+} thread_local seed_;
+
+} // namespace
 
 uuid uuid::copy() {
     uuid u;
@@ -70,12 +83,7 @@ uuid uuid::copy(const char* bytes) {
 
 uuid uuid::random() {
     uuid bytes;
-    int r = std::rand();
-    for (size_t i = 0; i < bytes.size(); ++i ) {
-        bytes[i] = r & 0xFF;
-        r >>= 8;
-        if (!r) r = std::rand();
-    }
+    std::generate(bytes.begin(), bytes.end(), std::ref(seed_.engine));
 
     // From RFC4122, the version bits are set to 0100
     bytes[6] = (bytes[6] & 0x0F) | 0x40;


cpp/src/uuid.cpp
```diff
@@ -70,11 +74,11 @@ uuid uuid::copy(const char* bytes) {
 
 uuid uuid::random() {
     uuid bytes;
-    int r = std::rand();
+    int r = seed_.dist(seed_.engine);
     for (size_t i = 0; i < bytes.size(); ++i ) {
         bytes[i] = r & 0xFF;
         r >>= 8;
-        if (!r) r = std::rand();
+        if (!r) r = seed_.dist(seed_.engine);
```
jiridanek, line 81: Thinking about the original algorithm, this line means that if there is a zero byte at the "end" of the random number, it will not be put into the UUID and a new number will be generated instead. For example, if the `seed_.dist` before the loop generated a 0, then only the first byte will be put into the uuid and then immediately a new random number will be generated. I'm trying to say that the random data in the UUID will be IMO somewhat biased against `\x00` bytes.

Maybe best to use something like `std::independent_bits_engine` suggested in the first answer on https://stackoverflow.com/questions/25298585/efficiently-generating-random-bytes-of-data-in-c11-14

cpp/src/uuid.cpp
```diff
@@ -42,15 +44,17 @@ namespace {
 
 // Seed the random number generated once at startup.
 struct seed {
```
jiridanek, line 46: Threadsafety, I almost forgot! The std:: random stuff is not threadsafe, so maybe it should be declared thread_local as that would be easiest.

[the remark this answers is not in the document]
jiridanek: Instead of having the seed_, the variables could be private static in the uuid class, btw.

jiridanek: (assuming that there is not something else which prevents concurrent use; I think there isn't)

jiridanek:
Eh, I wanted to be putting comments on individual lines of code. They do show correctly at https://github.com/apache/qpid-proton/commit/a3c7f46498baaf7e2a4005dbbd2d02201121e745, but they do not show with the lines here on the PR. Sorry. Anyways, yeah, I think I have pretty much nothing to complain about here.

cpp/src/uuid.cpp
DreamPearl: Done, Thanks!

cpp/src/uuid.cpp
DreamPearl: I use thread_local here, this is my first time using it, and I couldn't find many examples that include both cpp and hpp files. Can you please take a look?

jiridanek (commented):
I'll leave the final reviews to others, I think. In my judgement, I'd structure the variables differently, in what gets to be static, and what is thread_local. (I'd make ticks static, initialize it only once at the start of the program...) But. At this point it's probably best that you finalize this with a single person, who's got to be Andrew, because there is nobody else for the C++ binding, instead of multiple people throwing at you (possibly) conflicting suggestions.

DreamPearl:
Now the build is passing. @astitcher Can you please take a look?

cpp/include/proton/uuid.hpp
```diff
@@ -35,13 +38,20 @@ namespace proton {
...
+    thread_local static std::seed_seq seed;
+
   public:
     /// Make a copy.
     PN_CPP_EXTERN static uuid copy();
 
     /// Return a uuid copied from bytes.  Bytes must point to at least
     /// 16 bytes.  If `bytes == 0` the UUID is zero-initialized.
-    PN_CPP_EXTERN static uuid copy(const char* bytes);
+    PN_CPP_EXTERN static uuid copy(const char *bytes);
```
astitcher, line 54: FWIW the convention used in our C++ is in line with usual C++ conventions and has the '*' in the type side not the name side. Our C code is in line with more usual C conventions and is the other way round. This is probably not written down anywhere - sorry!

cpp/include/proton/uuid.hpp
```diff
@@ -35,13 +38,20 @@ namespace proton {
 
 /// A 16-byte universally unique identifier.
 class uuid : public byte_array<16> {
+
+  private:
+    thread_local static std::independent_bits_engine<std::mt19937, CHAR_BIT,
+                                                     unsigned int>
+        engine;
+    thread_local static std::seed_seq seed;
+
```
astitcher, line 47: Please don't add these private members to the class. It changes the ABI (also the API but in a more subtle way). 
Since these are thread_locals anyway they don't need to be per instance variables so can just be static to the implementation file itself and not be in the class at all.

cpp/src/uuid.cpp
```diff
@@ -38,20 +43,17 @@
...
+// started within the same second will get the same seed.
+thread_local unsigned long ticks =
+    std::chrono::system_clock::now().time_since_epoch().count();
+unsigned long pid = GETPID();
+unsigned int rd = std::random_device{}();
+} // namespace
+
+thread_local std::seed_seq uuid::seed{ticks, pid, (unsigned long)rd};
+thread_local std::independent_bits_engine<std::mt19937, CHAR_BIT, unsigned int>
+    uuid::engine(seed);
```
astitcher, line 56: I like the struct seed construct to initialise the seed once. You can shift your code into there and make it thread_local I think which will be a neater implementation and a smaller change.

cpp/src/uuid.cpp
```diff
@@ -38,20 +43,17 @@
...
-        // within the same second will get the same seed.
-        unsigned long secs = time(0);
-        unsigned long pid = GETPID();
-        std::srand(((secs*181)*((pid-83)*359))%104729);
-    }
-} seed_;
-
-}
+// A hash of time, PID and random_device, time alone is a bad seed as programs
+// started within the same second will get the same seed.
```
astitcher, line 47: I think this comment maybe now incorrect as std::chrono::system_clock::now() isn't in seconds any more (is it?)

cpp/src/uuid.cpp
```diff
@@ -70,12 +72,7 @@ uuid uuid::copy(const char* bytes) {
 
 uuid uuid::random() {
     uuid bytes;
-    int r = std::rand();
-    for (size_t i = 0; i < bytes.size(); ++i ) {
-        bytes[i] = r & 0xFF;
-        r >>= 8;
-        if (!r) r = std::rand();
-    }
+    std::generate(bytes.begin(), bytes.end(), std::ref(engine));
```
astitcher, line 75: in my way of doing this engine would be seed_::engine - maybe seed_ should have a better name?

astitcher (changes requested):

cpp/include/proton/uuid.hpp
jiridanek: @DreamPearl Are you still using that clang-format config? You could change the setting there and commit that... some time in the future.

cpp/include/proton/uuid.hpp
jiridanek: @DreamPearl Right, ABI. Sorry I forgot to think about that before.

astitcher: Actually thinking a bit more. It probably doesn't actually change the ABI. But it does unnecessarily leak implementation details of the uuid class. IMO it's better to just make these static thread_local in the implementation file without putting them in the class at all. This will generate pretty much exactly the same code but is better encapsulated (again IMO)

cpp/include/proton/uuid.hpp
DreamPearl: Yep, I am still using git clang-format. Updated the local configuration to adapt this.
Thanks!

cpp/src/uuid.cpp
DreamPearl: I have modified the code to initialize the seed once using the struct seed construct.

cpp/include/proton/uuid.hpp
DreamPearl: Moved them out of the class and put them all in struct construct.

cpp/src/uuid.cpp
DreamPearl: I use seed_.engine.

cpp/src/uuid.cpp
DreamPearl: Yeah, my bad! It was in nanoseconds. Updated the comment.

astitcher:
Commited with a minor change to simplify the initialization/use of the the random generator

PR #341, opened 2021-11-15
Change the input buffer growth logic to do something rather than nothing

ssorj:
I discovered by print debugging that, while the output buffer grows on demand, the input buffer does not.  The current logic to grow the input buffer is inside pn_transport_capacity (surprise!).  I moved it to pn_transport_process.  I'm not sure if the condition I'm testing for is the best one, but it does make the growth logic operate.  This improves performance by about 100k messages/second or roughly 15% for my standard test.

### DIFF
--- a/c/src/core/transport.c
+++ b/c/src/core/transport.c
@@ -2922,25 +2922,7 @@ ssize_t pn_transport_capacity(pn_transport_t *transport)  /* <0 == done */
   if (transport->tail_closed) return PN_EOS;
   //if (pn_error_code(transport->error)) return pn_error_code(transport->error);
 
-  ssize_t capacity = transport->input_size - transport->input_pending;
-  if ( capacity<=0 ) {
-    // can we expand the size of the input buffer?
-    int more = 0;
-    if (!transport->local_max_frame) {  // no limit (ha!)
-      more = transport->input_size;
-    } else if (transport->local_max_frame > transport->input_size) {
-      more = pn_min(transport->input_size, transport->local_max_frame - transport->input_size);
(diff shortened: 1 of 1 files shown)


astitcher:
Hmm, I think this may actually be the wrong approach. This bug was caused when we moved to the proactor and connection_driver (ok - so quite a long time ago now!) in the proactor code (epoll implementation) we have these lines (epoll.c:1201):

```
pn_rwbytes_t rbuf = pn_connection_driver_read_buffer(&pc->driver);
if (rbuf.size > 0 && !pc->read_blocked) {
  ssize_t n = read(pc->psocket.epoll_io.fd, rbuf.start, rbuf.size);
  if (n > 0) {
    pn_connection_driver_read_done(&pc->driver, n);
    ...
  }
  ...
}
```
`pn_connection_driver_read_buffer()` calls `pn_transport_capacity()` to get the the transport buffer so that we can find the space to read into. however at this point we don't know how many bytes we can read so there is no way to try to resize the buffer. `pn_connection_driver_read_done()` calls `pn_transport_process()` to actually do the protocol processing. it looks to me that the resizing that this change does will effectively try to make sure that the input buffer is sized big enough for 2 of whatever buffer was just read. So the speed up seems to come from making the buffer bigger so that fewer reads can happen.

I wonder if we should just make the buffer resizing logic much more explicitly in the proactor layer (or perhaps in the connection_driver layer). This might make it a bit easier to understand the buffering strategy and to perhaps alter it in the future,

astitcher:
Closed in favour of #349

PR #345, opened 2021-11-29
PROTON-2471 Run raw connection examples during proton-c examples test

jiridanek:


jiridanek:
The CI fails on Windows and macOS are legit. Raw feature is not implemented in other than epoll proactor. Test has to be skipped.

jiridanek:
There is no good way to propagate the info about proactor availability to the `testme` file in examples that I can think of. My best bet so far seems to be to skip building the `raw_` examples if epoll proactor is not selected, then skip the test if the examples binaries are not present.

PR #346, opened 2021-12-09
PROTON-2308: Add support for setting Dynamic Node Properties

DreamPearl:
[PROTON-2308](https://issues.apache.org/jira/browse/PROTON-2308)

### DIFF
--- a/cpp/include/proton/source_options.hpp
+++ b/cpp/include/proton/source_options.hpp
@@ -27,6 +27,7 @@
 #include "./duration.hpp"
 #include "./source.hpp"
 
+#include <map>
 #include <string>
 
 /// @file
@@ -91,6 +92,9 @@ class source_options {
 
     /// Extension capabilities that are supported/requested
     PN_CPP_EXTERN source_options& capabilities(const std::vector<symbol>&);
+
+    /// Set the dynamic node properties.
+    PN_CPP_EXTERN source_options& dynamic_properties(const std::map<symbol, value>&);
 
   private:
     void apply(source&) const;
--- a/cpp/include/proton/terminus.hpp
+++ b/cpp/include/proton/terminus.hpp
@@ -27,6 +27,7 @@
 
 #include <proton/terminus.h>
 
+#include <map>
 #include <string>
 #include <vector>
 
@@ -102,6 +103,10 @@ class terminus {
 
     /// Extension capabilities that are supported/requested
     PN_CPP_EXTERN std::vector<symbol> capabilities() const;
+
+    /// Obtain the AMQP dynamic node properties for the
+    /// terminus as a standard map.
+    PN_CPP_EXTERN std::map<symbol, value> dynamic_properties() const;
 
   protected:
     pn_terminus_t *pn_object() const { return object_; }
--- a/cpp/src/link_test.cpp
+++ b/cpp/src/link_test.cpp
@@ -20,11 +20,120 @@
 
 #include "test_bits.hpp"
 
-#include <proton/sender_options.hpp>
-#include <proton/receiver_options.hpp>
+#include <proton/connection.hpp>
+#include <proton/connection_options.hpp>
 #include <proton/container.hpp>
+#include <proton/delivery.hpp>
+#include <proton/listen_handler.hpp>
+#include <proton/listener.hpp>
+#include <proton/message.hpp>
+#include <proton/messaging_handler.hpp>
+#include <proton/receiver_options.hpp>
+#include <proton/sender_options.hpp>
+#include <proton/source_options.hpp>
+#include <proton/target_options.hpp>
+#include <proton/types.hpp>
 
 #include <iostream>
+
+#include <condition_variable>
+#include <map>
+#include <mutex>
+#include <thread>
+
+namespace {
+std::mutex m;
+std::condition_variable cv;
+bool listener_ready = false;
+int listener_port;
+const std::string DYNAMIC_ADDRESS = "test_dynamic_address";
+} // namespace
+
+class test_recv : public proton::messaging_handler {
+  private:
+    class listener_ready_handler : public proton::listen_handler {
+        void on_open(proton::listener& l) override {
+            {
+                std::lock_guard<std::mutex> lk(m);
+                listener_port = l.port();
+                listener_ready = true;
+            }
+            cv.notify_one();
+        }
+    };
+
+    std::string url;
+    proton::listener listener;
+    listener_ready_handler listen_handler;
+
+  public:
+    test_recv(const std::string& s) : url(s) {}
+
+    void on_container_start(proton::container& c) override {
+        listener = c.listen(url, listen_handler);
+    }
+
+    void on_receiver_open(proton::receiver& r) override {
+        std::map<proton::symbol, proton::value> m_sender({{proton::symbol("supported-dist-modes"), proton::symbol("move")}});
+        std::map<proton::symbol, proton::value> props = r.target().dynamic_properties();
+
+        ASSERT(r.target().dynamic());
+        ASSERT(r.target().address().empty());
+        ASSERT_EQUAL(m_sender, props);
+
+        proton::target_options opts;
+        std::map<proton::symbol, proton::value> m({{proton::symbol("supported-dist-modes"), proton::symbol("copy")}});
+        opts.address(DYNAMIC_ADDRESS);
+        opts.dynamic_properties(m);
+        r.open(proton::receiver_options().target(opts));
+        listener.stop();
+    }
+};
+
+class test_send : public proton::messaging_handler {
+  private:
+    std::string url;
+
+  public:
+    test_send(const std::string& s) : url(s) {}
+
+    void on_container_start(proton::container& c) override {
+        proton::target_options opts;
+        std::map<proton::symbol, proton::value> m({{proton::symbol("supported-dist-modes"), proton::symbol("move")}});
+        opts.dynamic(true);
+        opts.dynamic_properties(m);
+        c.open_sender(url, proton::sender_options().target(opts));
+    }
+
+    void on_sender_open(proton::sender& s) override {
+        std::map<proton::symbol, proton::value> m({{proton::symbol("supported-dist-modes"), proton::symbol("copy")}});
+        std::map<proton::symbol, proton::value> props = s.target().dynamic_properties();
+
+        ASSERT_EQUAL(DYNAMIC_ADDRESS, s.target().address());
+        ASSERT_EQUAL(m, props);
+
+        s.connection().close();
+    }
+};
+
+int test_dynamic_node_properties() {
+
+    std::string recv_address("127.0.0.1:0/test");
+    test_recv recv(recv_address);
+    proton::container c(recv);
+    std::thread thread_recv([&c]() -> void { c.run(); });
+
+    // wait until listener is ready
+    std::unique_lock<std::mutex> lk(m);
+    cv.wait(lk, [] { return listener_ready; });
+
+    std::string send_address = "127.0.0.1:" + std::to_string(listener_port) + "/test";
+    test_send send(send_address);
+    proton::container(send).run();
+    thread_recv.join();
+
+    return 0;
+}
 
 int test_link_name()
 {
@@ -48,5 +157,6 @@ int test_link_name()
 int main(int argc, char** argv) {
     int failed = 0;
     RUN_ARGV_TEST(failed, test_link_name());
+    RUN_ARGV_TEST(failed, test_dynamic_node_properties());
     return failed;
 }
--- a/cpp/src/node_options.cpp
+++ b/cpp/src/node_options.cpp
@@ -20,14 +20,17 @@
  */
 
 #include "proton/codec/vector.hpp"
+#include "proton/map.hpp"
 #include "proton/source.hpp"
 #include "proton/source_options.hpp"
 #include "proton/target.hpp"
 #include "proton/target_options.hpp"
+#include "proton/types.hpp"
 
 #include "proton_bits.hpp"
 
 #include <limits>
+#include <map>
 
 namespace proton {
 
@@ -97,6 +100,7 @@ class source_options::impl {
     option<enum source::distribution_mode> distribution_mode;
     option<source::filter_map> filters;
     option<std::vector<symbol> > capabilities;
+    option<std::map<symbol, value>> dynamic_properties;
 
     void apply(source& s) {
         node_address(s, address, dynamic, anonymous);
@@ -110,6 +114,11 @@ class source_options::impl {
         }
         if (capabilities.set) {
             value(pn_terminus_capabilities(unwrap(s))) = capabilities.value;
+        }
+        if (dynamic_properties.set) {
+            map<symbol, value> source_map;
+            get(dynamic_properties.value, source_map);
+            value(pn_terminus_properties(unwrap(s))) = source_map;
         }
     }
 };
@@ -134,6 +143,10 @@ source_options& source_options::expiry_policy(enum source::expiry_policy m) { im
 source_options& source_options::distribution_mode(enum source::distribution_mode m) { impl_->distribution_mode = m; return *this; }
 source_options& source_options::filters(const source::filter_map &map) { impl_->filters = map; return *this; }
 source_options& source_options::capabilities(const std::vector<symbol>& c) { impl_->capabilities = c; return *this; }
+source_options& source_options::dynamic_properties(const std::map<symbol, value>& c) {
+    impl_->dynamic_properties = c;
+    return *this;
+}
 
 void source_options::apply(source& s) const { impl_->apply(s); }
 
@@ -148,6 +161,7 @@ class target_options::impl {
     option<duration> timeout;
     option<enum target::expiry_policy> expiry_policy;
     option<std::vector<symbol> > capabilities;
+    option<std::map<symbol, value>> dynamic_properties;
 
     void apply(target& t) {
         node_address(t, address, dynamic, anonymous);
@@ -155,6 +169,11 @@ class target_options::impl {
         node_expiry(t, expiry_policy, timeout);
         if (capabilities.set) {
             value(pn_terminus_capabilities(unwrap(t))) = capabilities.value;
+        }
+        if (dynamic_properties.set) {
+            map<symbol, value> target_map;
+            get(dynamic_properties.value, target_map);
+            value(pn_terminus_properties(unwrap(t))) = target_map;
         }
     }
 };
@@ -177,6 +196,10 @@ target_options& target_options::durability_mode(enum target::durability_mode m) 
 target_options& target_options::timeout(duration d) { impl_->timeout = d; return *this; }
 target_options& target_options::expiry_policy(enum target::expiry_policy m) { impl_->expiry_policy = m; return *this; }
 target_options& target_options::capabilities(const std::vector<symbol>& c) { impl_->capabilities = c; return *this; }
+target_options& target_options::dynamic_properties(const std::map<symbol, value>& c) {
+    impl_->dynamic_properties = c;
+    return *this;
+}
 
 void target_options::apply(target& s) const { impl_->apply(s); }
 
--- a/cpp/src/terminus.cpp
+++ b/cpp/src/terminus.cpp
@@ -19,10 +19,10 @@
  *
  */
 
-#include "proton/duration.hpp"
 #include "proton/terminus.hpp"
+#include "proton/duration.hpp"
+#include "proton/types.hpp"
 #include "proton/value.hpp"
-#include "proton/codec/vector.hpp"
 
 #include "proton_bits.hpp"
 
@@ -61,4 +61,12 @@ std::vector<symbol> terminus::capabilities() const {
     return caps.empty() ? std::vector<symbol>() : caps.get<std::vector<symbol> >();
 }
 
+std::map<symbol, value> terminus::dynamic_properties() const {
+    value props(pn_terminus_properties(object_));
+    std::map<symbol, value> properties;
+    if (!props.empty()) {
+        get(props, properties);
+    }
+    return properties;
+}
 }
--- a/cpp/include/proton/target_options.hpp
+++ b/cpp/include/proton/target_options.hpp
@@ -27,6 +27,7 @@
 #include "./duration.hpp"
 #include "./target.hpp"
 
+#include <map>
 #include <string>
 
 /// @file
@@ -82,6 +83,9 @@ class target_options {
 
     /// Extension capabilities that are supported/requested
     PN_CPP_EXTERN target_options& capabilities(const std::vector<symbol>&);
+
+    /// Set the dynamic node properties.
+    PN_CPP_EXTERN target_options& dynamic_properties(const std::map<symbol, value>&);
 
   private:
     void apply(target&) const;


DreamPearl:
@astitcher Updated the PR. In middle of writing tests.

cpp/include/proton/terminus.hpp
```diff
@@ -103,6 +104,10 @@ class terminus {
     /// Extension capabilities that are supported/requested
     PN_CPP_EXTERN std::vector<symbol> capabilities() const;
 
+    /// Obtain a reference to the AMQP dynamic node properties for the
+    /// terminus as a standard map.
+    PN_CPP_EXTERN std::map<symbol, value> dynamic_node_properties();
```
jiridanek, line 109: Comment says reference, but function returns a copy?

cpp/src/link_test.cpp
```diff
@@ -23,9 +23,114 @@
...
+
+    void on_sendable(proton::sender &s) override {
+        proton::message msg;
+        msg.body("message");
+        proton::tracker t = s.send(msg);
+        s.connection().close();
+    }
+};
+
+int test_dynamic_node_properties() {
```
jiridanek, line 114: This test sets properties on a sender and then reads them out from the receiver's target. I am wondering if the other direction is also needed to be tested: set properties on receiver and read them from sender. Possibly not.

cpp/src/link_test.cpp
```diff
@@ -23,9 +23,114 @@
...
+    std::unique_lock<std::mutex> lk(m);
+    cv.wait(lk, [] { return listener_ready; });
+
+    std::string send_address =
+        "127.0.0.1:" + std::to_string(listener_port) + "/test";
+    test_send send(send_address);
+    proton::container(send).run();
+    thread_recv.join();
+
+    return 0;
```
jiridanek, line 131: there is a lot of tests that have this basic structure, but differ in details (depending on what function of API is currently being tested). I am wondering if they could be somehow refactored. But that is not here nor there for this PR.

cpp/src/link_test.cpp
DreamPearl: Well, I am also wondering if the other direction needs to be tested as well or not. I gave it a try (set the properties on the receiver and try to read them from the sender in the same test).

- If I am setting dynamic node properties on the receiver like this `c.receiver_options(proton::receiver_options().source(opts));` and reading it from the sender then the test is getting terminated. And I can not see them getting set in the `attach frame`, plus the SOURCE in the third and fourth attach frame is NULL, which might be the reason for termination.

- But if I am setting it like this`c.receiver_options(proton::receiver_options().target(opts));` , then all is good.

The question is which one is correct or none of them is correct.

Meanwhile, I am looking for resources to get the idea.

Below are the attach frames:
```
[0x1fef6d0]: AMQP:FRAME:0 -> @attach(18) [name="3d0a3e34-a9fd-42b8-9cfd-0b1bbcc9c910", handle=0x0,
role=false, snd-settle-mode=0x2, rcv-settle-mode=0x0, source=@source(40) [address=null, durable=0x0, expiry-
policy=null, timeout=0x0, dynamic=false, dynamic-node-properties=null, distribution-mode=null, filter=null, default-
outcome=null, outcomes=null, capabilities=null], target=@target(41) [address="test", durable=0x0, expiry-policy=null,
 timeout=0x0, dynamic=false, dynamic-node-properties={:symbol="value"}, capabilities=null], unsettled=null, incomplete-
unsettled=null, initial-delivery-count=0x0, max-message-size=0x0, offered-capabilities=null, desired-capabilities=null,
 properties=null]
```
```
[0x7f7ffc002de0]: AMQP:FRAME:0 <- @attach(18) [name="3d0a3e34-a9fd-42b8-9cfd-0b1bbcc9c910", handle=0x0, 
role=false, snd-settle-mode=0x2, rcv-settle-mode=0x0, source=@source(40) [address=null, durable=0x0, expiry-
policy=null, timeout=0x0, dynamic=false, dynamic-node-properties=null, distribution-mode=null, filter=null, default-
outcome=null, outcomes=null, capabilities=null], target=@target(41) [address="test", durable=0x0, expiry-policy=null,
timeout=0x0, dynamic=false, dynamic-node-properties={:symbol="value"}, capabilities=null], unsettled=null, incomplete-
unsettled=null, initial-delivery-count=0x0, max-message-size=0x0, offered-capabilities=null, desired-capabilities=null,
properties=null]
```
```
[0x7f7ffc002de0]: AMQP:FRAME:0 -> @attach(18) [name="3d0a3e34-a9fd-42b8-9cfd-0b1bbcc9c910", handle=0x0, 
role=true, snd-settle-mode=0x2, rcv-settle-mode=0x0, source=null, target=@target(41) [address="test", durable=0x0,
expiry-policy=null, timeout=0x0, dynamic=false, dynamic-node-properties={:symbol2="value2"}, capabilities=null],
unsettled=null, incomplete-unsettled=null, initial-delivery-count=0x0, max-message-size=0x0, offered-capabilities=null, 
desired-capabilities=null, properties=null]
```
```
[0x1fef6d0]: AMQP:FRAME:0 <- @attach(18) [name="3d0a3e34-a9fd-42b8-9cfd-0b1bbcc9c910", handle=0x0, role=true,
 snd-settle-mode=0x2, rcv-settle-mode=0x0, source=null, target=@target(41) [address="test", durable=0x0, expiry-
policy=null, timeout=0x0, dynamic=false, dynamic-node-properties={:symbol2="value2"}, capabilities=null],
unsettled=null, incomplete-unsettled=null, initial-delivery-count=0x0, max-message-size=0x0, offered-capabilities=null, 
desired-capabilities=null, properties=null]
```
[Complete logs](https://gist.github.com/DreamPearl/f35d3431a28c0432ad04352c4b0bffa5)

gemmellr: When being the 'requesting client' party asking for a consumer link to attach and have a dynamic node be created by setting the dynamic flag [and optionally setting dynamic node properties] that is done via the _source_, as it is requesting somewhere to consume from is dynamically created for it. The other side would read the source details requested by the requesting peer and action things, creating a node, giving it an address, and responding with its details. When it is sending its 'reply' attach it sets the live details that now exist also via the source (which from its perspective is now the local side, which it is authoritative on). The original peer can then read those details to establish the address and properties of the created node.

For a producer, it is similar but the _target_ is where the details will be set, as the original peer is requesting that somewhere to send to is dynamically created for it. The other peer will action this and respond with details via the _target_ to indicate what actually exists.

gemmellr: A dynamic consumer is by far the primary use case for dynamic, so if only having one test it would be that (essentially the opposite of this test). Having one test for a consumer, and another for a producer (from the client perspective) wouldnt be a bad idea, they are separate things, but dynamic senders are rare.

The test doesnt actually do all the steps I'd say it should. Its possible another test is doing some of them somewhere, or that its just not tested for the cpp binding. It should use a 'dynamic' source/target and verify the behaviour in full, right now it isnt, but is instead attaching to a known address. The 'dynamic' handling involves the original peer not setting an address, and the other peer generating an address for the dynamic node and supplying its detail back to the original side in its reply attach, which the requesting peer can then read to determine the name of the dynamic node created for it. The dynamic-node-properties adds to that mechanism by letting the requester also ask for certain node details to be applied. The generating peer can always use that same field to indicate properties actually in effect for the dynamic node it created.

The test only looks to verify that the node properties were sent [illegally due to not setting dynamic] from requestor to the other side. It doesnt check it can handle reading the details that come back, i.e to verify any dynamic address was given, or that any dynamic-node-properties arrived.

As an example, here is a Java test verifying the dynamic flag and address propagation aspects (it doesnt test the dynamic node properties) for a dynamic consumer:
https://github.com/vert-x3/vertx-proton/blob/4.2.2/src/test/java/io/vertx/proton/ProtonClientTest.java#L468-L545

I dont know if the 'built in' handshaking bits in proton-cpp that are opening the link in your test will inspect the dynamic flag and generate an address by themselves for the attach. If they do the test could just at least verify there is a generated value. If not, or better still, the test could set a generated address at the server side and check the client side can pick it up when its told the link opened. It could do the same for the dynamic node properties, e.g have the server apply something other than originally asked for, verify they are picked up too at the client once the link opens.

DreamPearl: For now, I have the server apply dynamic node properties other than the desired properties asked by the client. And verifying that they are picked up by the client.

pn_log:  https://gist.github.com/DreamPearl/129d9d7d4f22b237269f6a5beefb227f

gemmellr: That bit sounds/looks good. There are peripheral issues as mentioned before that should also be addressed. 

The trace log shows the test still uses an illegal target definition in the clients sender attach, where it sets dynamic-node-properties, but doesnt set dynamic=true, and sets an address "test" for the target (which the server naively copies and so reflects back). Thats then compounded when the test/client expects a 'response' attach with dynamic=true, something which shouldn't occur in this scenario.

Right now it verifies there are [different] dynamic node properties on the the servers 'response' attach, and the dynamic flag is set on the target, but not that there is a generated address.  It would be good for the test to verify the whole of the 'dynamic' interation, i.e that the client 'requests' dynamic=true and doesnt set an address, and that the client can pick up the address value set by the server once the link opens (since the dynamic node would be useless without an address).

I think it would be better if the server did its work in the 'on_receiver_open' (or per next comment, sender open) method rather than using the options in the container init, since you wont/cant generally use the options in that situation as a server as not everything is dynamic, and each dynamic node needs its own address etc. When the link attaches the server could verify it got a proper/expected request for a dynamic node, and respond setting the appropriate details on the target directly (or probably rather the source, per next comment), rather than using the options and also might as well open the link itself and not bothering with calling messaging_handler::on_receiver_open(r);).

Main other comment would be that this is still only trying to test a 'dynamic sender' from the clients perspective, when almost noone will ever do that, so it should either use a 'dynamic receiver' or test both.

cpp/src/link_test.cpp
```diff
@@ -23,9 +23,114 @@
...
+#include <thread>
+
+namespace {
+std::mutex m;
+std::condition_variable cv;
+bool listener_ready = false;
+int listener_port;
+const std::string DNP_SYMBOL = "DNP_SYMBOL";
+const std::string DNP_VALUE = "DNP_VALUE";
+} // namespace
```
astitcher, line 51: If you are trying to replicate a realistic scenario as @gemmellr is suggesting maybe you want to use real property values as defined in the section 3.5.9 of the AMQP standard?
If I read the standard correctly the 2 defined properties are the symbols "lifetime-policy" and "supported-dist-modes":

The allowed values of lifetime policy seem to be described lists with values '@delete-on-closed(43)[]', '@delete-on-nolinks(44)[]', '@delete-on-no-messages{45)[]', '@delete-on-no-links-or-messages(46)[]' - these all have specific meanings as to when the creator of the dynamic node should delete the node. As a side note it seems unusual but these property values seem to be described lists which should always be empty as there is no useful thing to put in them. 

The allowed values for 'supported-dist-modes' are the symbols 'move' and 'copy' and the list '['move', copy'] (or in the other order!). Or even single element lists with just symbols 'copy' or 'move' I think.

Of course we have no code in the C++ binding that actually cares what the values are so this doesn't matter a whole lot, but it might be nice to generate realistic dynamic property values.

cpp/src/link_test.cpp
astitcher: Although @gemmellr is correct in his analysis of what is going on in a real AMQP messaging interaction, this is probably not all that important as the C++ binding (at least currently) has no logic at either sender or receiver to do anything special for dynamic nodes - it just passes the info on to the application which is expected to validate and act on that information.
So from the point of view of this test it is only important to check that the properties get correctly carried in both directions. Unless the tester itself generates the requested nodes the API is not going to so there really isn't anything to validate in the test.
In a future change to the C++ binding it would definitely make sense to validate that you can't set address and set dynamic if you are sender/target or receiver/source. And also that if dynamic node properties are set the dynamic must also be set for the node, but that isn't done by the binding (or proton-c) currently so testing for it makes no sense. Producing realistic protocol interactions makes some sense but only to make the test "more realistic".

cpp/src/node_options.cpp
```diff
@@ -19,15 +19,17 @@
  *
  */
 
-#include "proton/codec/vector.hpp"
```
astitcher, line 22: Not sure why this has been removed. It doesn't seem related to your changes. If it's not needed it would be better (and less distracting) to remove it in a separate change (it could be NO-JIRA if you're really sure it's not needed - effectively that is what this change is!)

cpp/src/terminus.cpp
```diff
@@ -61,4 +61,12 @@ std::vector<symbol> terminus::capabilities() const {
     return caps.empty() ? std::vector<symbol>() : caps.get<std::vector<symbol> >();
 }
 
+std::map<symbol, value> terminus::dynamic_properties() {
```
astitcher, line 64: As per my comment in the header file this method should be ```const```

astitcher (changes requested):

cpp/src/link_test.cpp
gemmellr: The supported-dist-modes would be the simplest real noe property (as anything legal is allwoed I didnt see much need to use specific values, but it would be a representative test). The supported-dist-modes value is noted as having a 'multiple' definition, which means it can takes a single symbol or an array (not list) of symbols, with null or an empty array both meaning not-specified (per section 1.4 Composite Type Representation in the spec).

gemmellr:
> Although @gemmellr is correct in his analysis of what is going on in a real AMQP messaging interaction, this is probably not all that important as the C++ binding (at least currently) has no logic at either sender or receiver to do anything special for dynamic nodes - it just passes the info on to the application which is expected to validate and act on that information.
So from the point of view of this test it is only important to check that the properties get correctly carried in both directions. 

I very much disagree. The test should verify functionality actually works the way it needs to be used, and thus help ensure someone doesn't go breaking it later. A test of 'dynamic node [properties]' usage that doesnt actually use the handling/options for dynamic nodes in the required way is simply not a good test. It means it has never been verified to fully work in the situation it needs to, and its never being fully verified it isnt broken later. Assuming things work a certain way is how so many bugs come up when it later turns out it doesnt.

If there is a test of the 'no node properties' dynamic usage elsewhere then that would at least mean the 2 bits were somewhat tested in isolation even if the node-props bit was still not actually in full (which would still be poor but not quite as much). I might expect that other testing to be in this same test class, but it doesnt appear so. Is that bit actually tested right now anywhere?

>Unless the tester itself generates the requested nodes the API is not going to so there really isn't anything to validate in the test.
In a future change to the C++ binding it would definitely make sense to validate that you can't set address and set dynamic if you are sender/target or receiver/source. And also that if dynamic node properties are set the dynamic must also be set for the node, but that isn't done by the binding (or proton-c) currently so testing for it makes no sense. Producing realistic protocol interactions makes some sense but only to make the test "more realistic".

Youd simply check the request, set an address value to pass back, mark it dynamic, and then verify the receiver got that address and flag. That would in fact verify that the client options to set dynamic work as expected, that the client is able to read back the provided dynamic address as it needs to be able to, and inspect the dynamic flag was set appropriately. Thats not 'nothing to validate', and is not about just being 'more realistic', but about actually testing what needs to work to use this functionality.

I didnt suggest adding negative tests thats you cant set the things in the illegal way you can currently, only that the test should cover testing it actually does the things it should do when used the way it should be, which appears untested.

astitcher:
> > Although @gemmellr is correct in his analysis of what is going on in a real AMQP messaging interaction, this is probably not all that important as the C++ binding (at least currently) has no logic at either sender or receiver to do anything special for dynamic nodes - it just passes the info on to the application which is expected to validate and act on that information.
> > So from the point of view of this test it is only important to check that the properties get correctly carried in both directions.
> 
> I very much disagree. The test should verify functionality actually works the way it needs to be used, and thus help ensure someone doesn't go breaking it later. A test of 'dynamic node [properties]' usage that doesnt actually use the handling/options for dynamic nodes in the required way is simply not a good test. It means it has never been verified to fully work in the situation it needs to, and its never being fully verified it isnt broken later. Assuming things work a certain way is how so many bugs come up when it later turns out it doesnt.

I'm not sure we're actually disagreeing very much here! I agreee it makes the most sense to test the API in the way that it would be actually used - my major point is simply that the binding code has no magic in it to ensure correct AMQP semantics and so this cannot  be tested here. A minor point we do disagree on is that I think testing that the values get transferred correctly is the most important point and I think @gemmellr is saying that is not much use unless it actually tests what might be used - I don't think this disagreement matters much! Doing it the @gemmellr way satisfies me too!

> 
> If there is a test of the 'no node properties' dynamic usage elsewhere then that would at least mean the 2 bits were somewhat tested in isolation even if the node-props bit was still not actually in full (which would still be poor but not quite as much). I might expect that other testing to be in this same test class, but it doesnt appear so. Is that bit actually tested right now anywhere?
> 

Yes it would be really good to have that test here too! But currently we have no C++ test for this at all.

astitcher:
As a side note making properties that represent the documented AMQP possibilities is not actually all that easy as I said in my previous comment. We maybe need something predefined to allow the auto delete behaviours to be set more conveniently.

cpp/src/link_test.cpp
astitcher: Oh, yes - I forgot that multiple means array not list!  A painful (but semantically correct!) decision being the only place in the standard that arrays are required!

cpp/src/link_test.cpp
```diff
@@ -23,9 +23,121 @@
...
+
+  public:
+    test_recv(const std::string &s) : url(s) {}
+
+    void on_container_start(proton::container &c) override {
+        listener = c.listen(url, listen_handler);
+    }
+
+    void on_receiver_open(proton::receiver& r) override {
+        std::map<proton::symbol,proton::value> m_sender({{proton::symbol("supported-dist-modes"), proton::value("move")}});
```
astitcher, line 77: Per the standard I think 'move' needs to be a symbol too. I think this will create an AMQP string value of 'move' which is not what the standard requests.

cpp/src/link_test.cpp
```diff
@@ -23,9 +23,121 @@
...
+        std::map<proton::symbol,proton::value> props = r.target().dynamic_properties();
+
+        ASSERT(r.target().dynamic());
+        ASSERT(r.target().address().empty());
+        ASSERT_EQUAL(m_sender, props);
+
+        proton::target_options opts;
+        std::map<proton::symbol, proton::value> m({{proton::symbol("supported-dist-modes"), proton::value("copy")}});
```
astitcher, line 85: As above I think this should be ```proton::value(proton::symbol("copy"))```

cpp/src/link_test.cpp
```diff
@@ -23,9 +23,121 @@
...
+class test_send : public proton::messaging_handler {
+  private:
+    std::string url;
+
+  public:
+    test_send(const std::string& s) : url(s) {}
+
+    void on_container_start(proton::container& c) override {
+        proton::target_options opts;
+        std::map<proton::symbol,proton::value> m({{proton::symbol("supported-dist-modes"), proton::value("move")}});
```
astitcher, line 104: As above - maybe it would be better to have constants for move and copy symbols that can be used in this code to avoid repeating ```proton::value(proton::symbol("move"))``` etc. actually perhaps ```proton::symbol("move")``` may be enough as I think it should automatically convert to ```proton::value``

astitcher (changes requested):
@DreamPearl  I'm now pretty happy with this - and will merge it with some small changes to the tests as long as @gemmellr thinks that the resulting tests are sufficiently representative of real protocol use.

cpp/src/link_test.cpp
DreamPearl: Got it, done!

cpp/src/link_test.cpp
DreamPearl: I tried to make them 'constant' locally but I found the current version to be more readable. However, if you think adding constants will add more value, I can update it.

DreamPearl:
@gemmellr I've made a few changes. Can you please take a look?

cpp/include/proton/source_options.hpp
```diff
@@ -27,6 +27,7 @@
 #include "./duration.hpp"
 #include "./source.hpp"
 
+#include <map>
 #include <string>
```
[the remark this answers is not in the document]
DreamPearl: Got it. I will raise a small PR for this.

cpp/include/proton/terminus.hpp
```diff
@@ -103,6 +104,10 @@ class terminus {
     /// Extension capabilities that are supported/requested
     PN_CPP_EXTERN std::vector<symbol> capabilities() const;
 
+    /// Obtain the AMQP dynamic node properties for the
+    /// terminus as a standard map.
+    PN_CPP_EXTERN std::map<symbol, value> dynamic_properties();
```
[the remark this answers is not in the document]
DreamPearl: I actually removed it temporarily and forgot to add it back. :(

cpp/src/link_test.cpp
```diff
@@ -23,9 +23,121 @@
...
+    void on_container_start(proton::container &c) override {
+        listener = c.listen(url, listen_handler);
+    }
+
+    void on_receiver_open(proton::receiver& r) override {
+        std::map<proton::symbol,proton::value> m_sender({{proton::symbol("supported-dist-modes"), proton::symbol("move")}});
+        std::map<proton::symbol,proton::value> props = r.target().dynamic_properties();
+
+        ASSERT(r.target().dynamic());
+        ASSERT(r.target().address().empty());
```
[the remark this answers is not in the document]
DreamPearl: Internally the address is NULL but while retrieving address() as std::string, the NULL will get converted to empty string as per [target::address()](https://github.com/apache/qpid-proton/blob/main/cpp/src/target.cpp#L41)

Trace: https://gist.github.com/DreamPearl/229a5878a3935e43b6a0dfca9490226a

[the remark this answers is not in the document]
gemmellr: Ah I see, I guess that will need to do if that is the API unfortunately.

(Another good example where a test peer that validated wire behaviour without the API would be useful)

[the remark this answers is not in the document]
DreamPearl:
```
void on_receiver_open(proton::receiver& r) override {
       ...
        ASSERT(r.target().address().empty());
        const char *address = pn_terminus_get_address(unwrap(r.target()));
        ASSERT(NULL == address);
        ASSERT_EQUAL(m_sender, props);
       ...
    }
};
```
If we don't use the address method, then I guess we have to do something like the above. That means we will be using the C function in the C++ binding test, is that something we can/should do? As I don't see it being done in any other C++ test.

Also, I am getting an error in the above snippet.
```
...
[100%] Linking CXX executable link_test
/usr/bin/ld: CMakeFiles/link_test.dir/src/link_test.cpp.o: undefined reference to symbol 'pn_terminus_get_address'
...
```
Alternatively, we may test this behavior under C tests (Maybe this is what you meant above).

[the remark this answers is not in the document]
gemmellr: If the API is intended to return an empty string then I think that its ok if that is all the test verifies. I dont know about adding use of the C function in the C++ tests, not my area of expertise at all, though if it was easy and not frowned upon then I personally would consider verifying the value fully to be worthwhile. But if isnt simple or is frowned upon to do such things then thats fair enough.

I wasnt suggesting you add C tests (though I do think they _should_ already exist)

cpp/src/link_test.cpp
```diff
@@ -23,9 +23,121 @@
...
+        ASSERT_EQUAL(m_sender, props);
+
+        proton::target_options opts;
+        std::map<proton::symbol, proton::value> m({{proton::symbol("supported-dist-modes"), proton::symbol("copy")}});
+        opts.address(DYNAMIC_ADDRESS);
+        opts.dynamic_properties(m);
+        if (r.uninitialized()) {
```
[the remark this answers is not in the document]
DreamPearl: Ah, it was not needed as you mentioned. Removed it. Thanks!

PR #347, opened 2021-12-16
Extend frame logging to optionally decode transfer payloads as AMQP

astitcher:
You can turn on this extra frame logging by setting the 'frame+' log level. This can usually be done by setting the PN_LOG environment variable to 'frame+'

This PR also contains changes to the recent AMQP value dump code so that it uses a fixed size string output buffer instead of using the pn_string_t type which will grow indefinitely.

I'd be very interested in people using this extra frame trace to see whether they think it's useful.


astitcher:
In response to some feedback I've made the full frame logging the default (and only) behaviour when frame tracing is turned on (PN_LOG=frame). The extra control wasn't very useful, at least for this proof of concept.

Now that null values are not printed in frame traces the traces become significantly less cluttered and more useful than before.

astitcher:
My intention is to split out the fixed string work from this branch as it isn't really the focus of this WIP, but an independently useful change which gos to bounding the size of the logging buffers that be received by a logger sink.

PR #348, opened 2021-12-17
Document ttl=0 behavior

ktdreyer:
Update the API documentation to explain that the default message `ttl` value is `0`.

Explain that Proton does not send any `ttl` header on the wire when a message's ttl is `0`.

### DIFF
--- a/c/include/proton/message.h
+++ b/c/include/proton/message.h
@@ -203,6 +203,9 @@ PN_EXTERN pn_millis_t    pn_message_get_ttl               (pn_message_t *msg);
  *
  * See ::pn_message_get_ttl() for a detailed description of message ttl.
  *
+ * The default ttl value for a new message is 0. If this value is 0, Proton
+ * will send no ttl message header.
+ *
  * @param[in] msg a message object
  * @param[in] ttl the new value for the message ttl
  * @return zero on success or an error code on failure
--- a/python/proton/_message.py
+++ b/python/proton/_message.py
@@ -225,6 +225,9 @@ class Message(object):
         """The time to live of the message measured in seconds. Expired messages
         may be dropped.
 
+        The default ttl value for a new message is 0. If this value is 0,
+        Proton will send no ttl message header.
+
         :raise: :exc:`MessageException` if there is any Proton error when using the setter.
         """
         return millis2secs(pn_message_get_ttl(self._msg))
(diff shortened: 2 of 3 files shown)


jiridanek:
@ktdreyer The behavior of not sending the default values is something that Proton does in general, not just for the TTL, IIRC. Is there some particular gotcha you encountered, where this behavior is problematic?

ktdreyer:
Yes, https://pagure.io/koji/issue/3120 has the background. There is some confusion around what a `0` TTL means. Does it mean "send a header with a zero value", or "send no header" ?

Python has a `NoneType`, so a programmer might assume that `0` sends a zero header, while `None` means send no header at all. My doc update in this PR implicitly clarifies why Proton uses `0` instead of `None` to mean "no ttl header".

astitcher:
Having checked the AMQP spec there is no default value for TTL. I think that proton-c just reuses 0 to mean no ttl header because sending a ttl of 0 makes no sense as it would mean expire this message immediately.

astitcher:
> Yes, https://pagure.io/koji/issue/3120 has the background. There is some confusion around what a `0` TTL means. Does it mean "send a header with a zero value", or "send no header" ?

It seems that for proton-c it means send no header - this is an artifact of the C API taking an integer with no distinguished 'None' value. A newly created message will have this value anyway - in that sense it is the default value!
> 
> Python has a `NoneType`, so a programmer might assume that `0` sends a zero header, while `None` means send no header at all. My doc update in this PR implicitly clarifies why Proton uses `0` instead of `None` to mean "no ttl header".

That is sensible but the the Python binding sits on top of the proton-c so cannot set the value to 0 because the c code uses this to mean don't send the value. I don't think it matters because send a TTL of 0 makes very little sense in any case - meaning this message has already timed out.

It definitely makes sense for the Python binding to translate a None value to 0 into the C API.

astitcher:
> Yes, https://pagure.io/koji/issue/3120 has the background. There is some confusion around what a `0` TTL means. Does it mean "send a header with a zero value", or "send no header" ?

Incidentally setting a ttl header value of 0 would be (mostly) useless because it means the message has 0ms before it expires, in other words it is already expired and can be immediately discarded - so any ActiveMQ behaviour that takes this to mean keep indefinitely is wrong.

However I expect that what is actually happening is that the ttl value is not being sent at all and the broker is just not configured to auto delete messages at all.

PR #349, opened 2021-12-17
A different approach to resizing the input buffer

astitcher:
This is a proposed change for the PR #341


astitcher:
This change improves the performance from ~640k msgs/s to ~ 690k msgs/s on my laptop over a loopback connection. YMMV!

astitcher:
Change merged

PR #351, opened 2022-01-04
NO-JIRA Remove header duplicacy in node_options.cpp and target.cpp

DreamPearl:
- node_options.cpp includes both `#include "proton/codec/vector.hpp"` and  `#include "proton/types.hpp"` 
  - `#include "proton/types.hpp"` includes proton types used to represent AMQP types and thus already consist `#include "proton/codec/vector.hpp"`
  - Note: node_options.cpp also needs `#include "proton/codec/map.hpp"` which is being covered by types.hpp file.

- target.cpp includes `#include "proton_bits.hpp"` twice.

### DIFF
--- a/cpp/src/node_options.cpp
+++ b/cpp/src/node_options.cpp
@@ -19,7 +19,6 @@
  *
  */
 
-#include "proton/codec/vector.hpp"
 #include "proton/map.hpp"
 #include "proton/source.hpp"
 #include "proton/source_options.hpp"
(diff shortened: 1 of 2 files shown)


jiridanek:
I'd put the PR comment ("node_options.cpp includes both...") into the commit message. It is not always easy to navigate from commit in project history to the PR.

DreamPearl:
> I'd put the PR comment ("node_options.cpp includes both...") into the commit message. It is not always easy to navigate from commit in project history to the PR.

Got it. Done! Thanks.

PR #352, opened 2022-01-07
PROTON-2308: Change dynamic_node API to be more like existing APIs

astitcher:
Also mark new APIs as unsettled

### DIFF
--- a/cpp/include/proton/terminus.hpp
+++ b/cpp/include/proton/terminus.hpp
@@ -22,6 +22,7 @@
  *
  */
 
+#include "./map.hpp"
 #include "./types_fwd.hpp"
 #include "./internal/export.hpp"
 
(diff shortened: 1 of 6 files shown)


cpp/include/proton/terminus.hpp
```diff
@@ -99,14 +102,15 @@ class terminus {
...
+    /// @deprecated in favor of dynamic_properties()??
     PN_CPP_EXTERN value node_properties() const;
 
-    /// Extension capabilities that are supported/requested
+    /// **Unsettled API** Extension capabilities that are supported/requested
     PN_CPP_EXTERN std::vector<symbol> capabilities() const;
 
-    /// Obtain the AMQP dynamic node properties for the
+    /// **Unsettled API** Obtain the AMQP dynamic node properties for the
     /// terminus as a standard map.
```
DreamPearl, line 112: @astitcher The documentation still says "standard map".

astitcher: Good catch

PR #353, opened 2022-01-10
PROTON-2481: [cpp] Improve constructor syntax for maps

DreamPearl:
[PROTON-2481](https://issues.apache.org/jira/browse/PROTON-2481)
- Add a copy constructor to copy from std::map
- Add an initializer_list constructor

### DIFF
--- a/cpp/include/proton/map.hpp
+++ b/cpp/include/proton/map.hpp
@@ -26,6 +26,8 @@
 #include "./internal/pn_unique_ptr.hpp"
 
 #include <cstddef>
+#include <initializer_list>
+#include <map>
 
 /// @file
 /// @copybrief proton::map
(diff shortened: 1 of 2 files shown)


cpp/include/proton/map.hpp
```diff
@@ -26,6 +26,8 @@
 #include "./internal/pn_unique_ptr.hpp"
 
 #include <cstddef>
+#include <initializer_list>
+#include <map>
```
astitcher, line 30: I'm not sure this is correct given my comment below - you may need pair though as that is used in the initializer_list constructor

cpp/include/proton/map.hpp
```diff
@@ -79,6 +81,12 @@ class PN_CPP_CLASS_EXTERN map {
     /// Copy a map.
     PN_CPP_EXTERN map& operator=(const map&);
 
+    /// Copy a std::map.
+    PN_CPP_EXTERN map(const std::map<K, T>&);
```
astitcher, line 85: I think this is too specific: The assigment operator uses a template to take any argument that is assignable to a proton::value: as a map - this constructor should do the same.
The assignment operator uses enable_if to ensure that it is the template type will work - I'm not sure you can use it in exactly the same way for a constructor. Actually I think we should replace the custom enable_if with std::enable_if now that we can use C++ 11

astitcher (changes requested):

jiridanek:
There aren't any usages for the new constructors. I'd like to see some test that calls them.

astitcher:
Merged

PR #354, opened 2022-01-14
PROTON-2060: Proposed doc changes to clarify auto-settle usage

kpvdr:
I have added text to the Python, C++, C and Ruby docs.

The Python, C++ and Ruby docs talk about the use of auto-settle. The C docs simply state the conditions for sender settle. The text used is based on the text of PROTON-2060.

### DIFF
--- a/c/include/proton/delivery.h
+++ b/c/include/proton/delivery.h
@@ -287,6 +287,12 @@ PN_EXTERN void pn_delivery_abort(pn_delivery_t *delivery);
  * @note If pn_delivery_current(delivery) is true before the call then
  * pn_link_advance(pn_delivery_link(deliver)) is called automatically.
  *
+ * @note The sender **should not** settle after only receiving a terminal
+ * status disposition with no settle flag, as then there would then be no way
+ * to receive any further events for that delivery (such as the subsequent
+ * on_settle message that might be expected when the receiver finally settles
+ * the message).
+ *
  * @param[in] delivery a delivery object
  */
 PN_EXTERN void pn_delivery_settle(pn_delivery_t *delivery);
--- a/c/include/proton/delivery.h
+++ b/c/include/proton/delivery.h
@@ -287,12 +287,6 @@ PN_EXTERN void pn_delivery_abort(pn_delivery_t *delivery);
  * @note If pn_delivery_current(delivery) is true before the call then
  * pn_link_advance(pn_delivery_link(deliver)) is called automatically.
  *
- * @note The sender **should not** settle after only receiving a terminal
- * status disposition with no settle flag, as then there would then be no way
- * to receive any further events for that delivery (such as the subsequent
- * on_settle message that might be expected when the receiver finally settles
- * the message).
- *
  * @param[in] delivery a delivery object
  */
 PN_EXTERN void pn_delivery_settle(pn_delivery_t *delivery);
--- a/cpp/include/proton/sender_options.hpp
+++ b/cpp/include/proton/sender_options.hpp
@@ -82,7 +82,16 @@ class sender_options {
     /// Set the delivery mode on the sender.
     PN_CPP_EXTERN sender_options& delivery_mode(delivery_mode);
 
-    /// Automatically settle messages (default is true).
+    /**
+     * Automatically settle messages (default is true).
+     *
+     * \note
+     * Sender **auto-settlement** only occurs for a delivery after the
+     * sender receives a settled disposition for that delivery. Otherwise,
+     * there would be no way to receive any further events for that delivery
+     * (such as the subsequent on_settle message that might be expected when
+     * the receiver finally settles the message).
+     */
     PN_CPP_EXTERN sender_options& auto_settle(bool);
 
     /// Options for the source node of the sender.
--- a/cpp/include/proton/sender_options.hpp
+++ b/cpp/include/proton/sender_options.hpp
@@ -83,14 +83,8 @@ class sender_options {
     PN_CPP_EXTERN sender_options& delivery_mode(delivery_mode);
 
     /**
-     * Automatically settle messages (default is true).
-     *
-     * \note
-     * Sender **auto-settlement** only occurs for a delivery after the
-     * sender receives a settled disposition for that delivery. Otherwise,
-     * there would be no way to receive any further events for that delivery
-     * (such as the subsequent on_settle message that might be expected when
-     * the receiver finally settles the message).
+     * If ``True`` (default), automatically settle messages upon receiving a settled
+     * disposition for that delivery. Otherwise messages must be explicitly settled.
      */
     PN_CPP_EXTERN sender_options& auto_settle(bool);
 
--- a/python/proton/_handlers.py
+++ b/python/proton/_handlers.py
@@ -48,12 +48,9 @@ class OutgoingMessageHandler(Handler):
     A utility for simpler and more intuitive handling of delivery
     events related to outgoing i.e. sent messages.
 
-    :param auto_settle: If ``True``, settle all messages (default). Otherwise
+    :param auto_settle: If ``True`` (default), automatically settle messages
+        upon receiving a settled disposition for that delivery. Otherwise
         messages must be explicitly settled.
-
-        .. note:: Sender auto-settlement only occurs for a delivery after the
-            sender receives a settled disposition for that delivery.
-
     :type auto_settle: ``bool``
     :param delegate: A client handler for the endpoint event
     """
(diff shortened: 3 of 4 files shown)


python/proton/_handlers.py
```diff
@@ -50,6 +50,10 @@ class OutgoingMessageHandler(Handler):
 
     :param auto_settle: If ``True``, settle all messages (default). Otherwise
         messages must be explicitly settled.
+
+        .. note:: Sender auto-settlement only occurs for a delivery after the
```
astitcher, line 54: Delete this as a separate para add the comment into the basic description of auto settle behaviour as in:
If  ``True`` (default), automatically settle messages upon receiving a settled disposition for that delivery...

python/proton/_handlers.py
```diff
@@ -663,11 +667,19 @@ class MessagingHandler(Handler, Acking):
     simpler to deal with and/or avoids repetitive tasks for common use
     cases.
 
+    .. note:: Sender **auto-settlement** only occurs for a delivery after the
+        sender receives a settled disposition for that delivery. Otherwise,
+        there would be no way to receive any further events for that delivery
+        (such as the subsequent on_settle message that might be expected when
+        the receiver finally settles the message).
+
```
astitcher, line 675: Putting this comment here seems unmotivated. It's just not connected narratively with the comment before.

python/proton/_handlers.py
```diff
@@ -950,11 +962,19 @@ class TransactionalClientHandler(MessagingHandler, TransactionHandler):
     and provides a convenience method :meth:`accept` for performing
     a transactional acceptance of received messages.
 
+    .. note:: Sender **auto-settlement** only occurs for a delivery after the
+        sender receives a settled disposition for that delivery. Otherwise,
+        there would be no way to receive any further events for that delivery
+        (such as the subsequent on_settle message that might be expected when
+        the receiver finally settles the message).
+
```
astitcher, line 970: As above delete this para we cover it below under auto settlement.
If we want to explain this point we need to attach the explanation to the explanation of settlement as a concept,.

python/proton/_handlers.py
```diff
@@ -950,11 +962,19 @@ class TransactionalClientHandler(MessagingHandler, TransactionHandler):
...
     :param prefetch: Initial flow credit for receiving messages, defaults to 10.
-    :param auto_accept: If ``True``, accept all messages (default). Otherwise messages
-        must be individually accepted or rejected.
+    :param auto_accept: If ``True``, accept all messages (default). Otherwise
+        messages must be individually accepted or rejected.
     :param auto_settle: If ``True``, settle all messages (default). Otherwise
-        messages must be explicitly settled.
+        messages must be explicitly settled. Sender auto-settlement only occurs
+        for a delivery after the sender receives a settled disposition for that
+        delivery.
```
astitcher, line 977: Again use the previous suggestion.

ruby/lib/core/sender.rb
```diff
@@ -29,10 +29,17 @@ class Sender < Link
 
     # Open the {Sender} link
     #
+    # @note Sender **auto-settlement** only occurs for a delivery after the
+    #     sender receives a settled disposition for that delivery. Otherwise,
+    #     there would be no way to receive any further events for that delivery
+    #     (such as the subsequent on_settle message that might be expected when
+    #     the receiver finally settles the message).
```
astitcher, line 36: I honestly don't understand why auto-settlement is called out specially here (or above). There's nothing special about it. Its described below under its option.

ruby/lib/core/sender.rb
```diff
@@ -29,10 +29,17 @@ class Sender < Link
...
     # @overload open_sender(address)
     #   @param address [String] address of the target to send to
     # @overload open_sender(opts)
-    #   @option opts [Boolean] :auto_settle (true) if true, automatically settle transfers
+    #   @option opts [Boolean] :auto_settle (true) if true, automatically settle transfers.
+    #       **Note:** Sender auto-settlement only occurs for a delivery after the sender receives
+    #       a settled disposition for that delivery.
```
astitcher, line 42: As above we're just describing the basic behaviour, do inline as I suggested for python.

python/proton/_handlers.py
```diff
@@ -663,11 +667,19 @@ class MessagingHandler(Handler, Acking):
...
+        the receiver finally settles the message).
+
     :param prefetch: Initial flow credit for receiving messages, defaults to 10.
     :param auto_accept: If ``True``, accept all messages (default). Otherwise
         messages must be individually accepted or rejected.
     :param auto_settle: If ``True``, settle all messages (default). Otherwise
-        messages must be explicitly settled.
+        messages must be explicitly settled. Sender auto-settlement only occurs
+        for a delivery after the sender receives a settled disposition for that
+        delivery.
```
astitcher, line 682: Use the suggested wording as above. We are describing the basic behaviour here and there is no need for a confusing qualification IMO.

c/include/proton/delivery.h
```diff
@@ -287,6 +287,12 @@ PN_EXTERN void pn_delivery_abort(pn_delivery_t *delivery);
  * @note If pn_delivery_current(delivery) is true before the call then
  * pn_link_advance(pn_delivery_link(deliver)) is called automatically.
  *
+ * @note The sender **should not** settle after only receiving a terminal
+ * status disposition with no settle flag, as then there would then be no way
+ * to receive any further events for that delivery (such as the subsequent
+ * on_settle message that might be expected when the receiver finally settles
+ * the message).
+ *
```
astitcher, line 295: What does settlement have to do with delivery abort? If there is a connection it needs to be called out explicitly to explain why this comment is here. Afaict If the sender aborts a delivery then it's likely they actually *do* want to settle the delivery because the delivery is now effectively dead and why would they want any more events about it? If you are making a comment relevant to this point then say that explicitly.
Essentially I think **should not** is way too strong here - if the application has no need for the receive settlement after aborting the message, and it seems to me likely they don't, then settling the delivery at the sender end first should be fine.

cpp/include/proton/sender_options.hpp
```diff
@@ -82,7 +82,16 @@ class sender_options {
...
+    /**
+     * Automatically settle messages (default is true).
+     *
+     * \note
+     * Sender **auto-settlement** only occurs for a delivery after the
+     * sender receives a settled disposition for that delivery. Otherwise,
+     * there would be no way to receive any further events for that delivery
+     * (such as the subsequent on_settle message that might be expected when
+     * the receiver finally settles the message).
+     */
```
astitcher, line 94: I think the suggested wording I made for the python API is good here too (see below). I don't see the need to call out auto-settlement in so many words, just explain the basic behaviour as below.

In the API docs that explain settlement we should explain that you can receive no more events after settling a delivery, but that understanding is fundamental to what *settlement is* and so shouldn't need explaining every place we talk about settlement.

astitcher (changes requested):
I think there are way too many words here. We can clarify the meaning of auto-settlement with fewer, better chosen words.
However this needs to be reviewed by someone with actual doc expertise to come at it from the user perspective rather than a technical expert.

python/proton/_handlers.py
kpvdr: Done.

python/proton/_handlers.py
kpvdr: Removed.

python/proton/_handlers.py
kpvdr: Done.

python/proton/_handlers.py
kpvdr: Deleted.

python/proton/_handlers.py
kpvdr: Done.

ruby/lib/core/sender.rb
kpvdr: Deleted.

ruby/lib/core/sender.rb
kpvdr: Done.

cpp/include/proton/sender_options.hpp
kpvdr: Done.

c/include/proton/delivery.h
kpvdr: I deleted this note.

astitcher (approved):
lgtm

[Do be sure to squash and rebase before merging though]

kpvdr:
Squashed and merged.

PR #355, opened 2022-01-19
PROTON-2487: [cpp] Implement distributed tracing using OpenTelemetry

DreamPearl:
[PROTON-2487](https://issues.apache.org/jira/browse/PROTON-2487)

- Add the opentelemetry-cpp library dependencies
- Add opentelemetry-cpp install/build steps in GHA
- Add tracing.cpp file
  - Add tracing spans for messaging operations
  - Add basic attributes[delivery_tag, delivery_state, source_address, destination_address]
- Add a tracing demo file in examples
- Add a test file

Previous discussion: https://github.com/apache/qpid-proton/pull/319

### DIFF
--- a/ci/otel.sh
+++ b/ci/otel.sh
@@ -1,0 +1,43 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+set -e
+
+sudo apt-get update
+
+# OTel dependencies
+
+sudo apt-get install libcurl4-openssl-dev
+sudo apt-get install libboost-locale-dev
+sudo apt-get install libthrift-dev
+
+# Clone OpenTelemetry-cpp
+
+git clone -b v1.6.1 --recursive https://github.com/open-telemetry/opentelemetry-cpp
+
+# Build/Install OpenTelemetry-cpp
+
+cd opentelemetry-cpp
+mkdir build
+cd build
+
+cmake .. -DCMAKE_INSTALL_PREFIX=/usr -DBUILD_SHARED_LIBS=ON -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DBUILD_TESTING=OFF -DCMAKE_BUILD_TYPE=RelWithDebInfo -DWITH_JAEGER=ON
+cmake --build . --target all
+sudo cmake --install . --config RelWithDebInfo
+cd ../..
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -36,6 +36,17 @@ else()
   set(CONNECT_CONFIG_SRC src/connect_config_dummy.cpp)
 endif()
 
+# Check for OPENTELEMETRY-CPP for distributed tracing
+find_package(opentelemetry-cpp)
+option(ENABLE_OPENTELEMETRYCPP "Use opentelemetry for distributed tracing" ${OPENTELEMETRY-CPP_FOUND})
+if (ENABLE_OPENTELEMETRYCPP)
+  include_directories(${OPENTELEMETRY_CPP_INCLUDE_DIRS})
+  set(TRACING_SRC src/tracing_opentelemetry.cpp src/tracing_stub.cpp)
+  set(TRACING_LIBS opentelemetry-cpp::trace)
+else()
+  set(TRACING_SRC src/tracing_stub.cpp src/init_tracer_stub.cpp)
+endif()
+
 list(APPEND PLATFORM_LIBS Threads::Threads)
 
 set(CXX_EXAMPLE_FLAGS "${CXX_WARNING_FLAGS} ${CXX_STANDARD}")
@@ -105,6 +116,7 @@ set(qpid-proton-cpp-source
   src/value.cpp
   src/work_queue.cpp
   ${CONNECT_CONFIG_SRC}
+  ${TRACING_SRC}
   )
 
 add_library(qpid-proton-cpp SHARED ${qpid-proton-cpp-source})
@@ -119,7 +131,7 @@ if(BUILD_STATIC_LIBS)
     EXPORT_NAME cpp)
 endif(BUILD_STATIC_LIBS)
 
-target_link_libraries (qpid-proton-cpp LINK_PRIVATE ${PLATFORM_LIBS} qpid-proton-core qpid-proton-proactor ${CONNECT_CONFIG_LIBS})
+target_link_libraries (qpid-proton-cpp LINK_PRIVATE ${PLATFORM_LIBS} qpid-proton-core qpid-proton-proactor ${CONNECT_CONFIG_LIBS} ${TRACING_LIBS})
 
 set_target_properties (
   qpid-proton-cpp
--- a/cpp/examples/CMakeLists.txt
+++ b/cpp/examples/CMakeLists.txt
@@ -82,6 +82,16 @@ foreach(example
   target_link_libraries(${example} Proton::cpp)
 endforeach()
 
+# Tracing examples
+if (ENABLE_OPENTELEMETRYCPP)
+foreach(example
+    tracing_client
+    tracing_server)
+  add_executable(${example} ${example}.cpp)
+  target_link_libraries(${example} Proton::cpp opentelemetry-cpp::trace opentelemetry-cpp::jaeger_trace_exporter opentelemetry-cpp::ostream_span_exporter)
+  endforeach()
+endif()
+
 # Examples that use threads directly
 foreach(example
     multithreaded_client
--- a/cpp/examples/tracing.dox
+++ b/cpp/examples/tracing.dox
@@ -1,0 +1,107 @@
+// -*-markdown-*-
+// NOTE: doxygen can include markdown pages directly but there seems to be a bug
+// that shows messed-up line numbers in \skip \until code extracts. This file
+// is markdown wrapped in a doxygen comment - which works. The file is best viewed/edited
+// as markdown.
+
+/**
+
+@page tracing Tracing
+
+The tracing support is added using OpenTelemetry-cpp.
+
+There are two types of spans:
+1. <b>Library generated spans</b>:
+
+    Send span <b>("amqp-message-send”)</b> gets created just before sending a message and ends when that message gets settled.
+
+    Receive span <b>("amqp-message-received")</b> gets created just before receiving a message and ends as soon as the message is received.
+
+2. <b>Application generated spans</b>:
+
+    Users can also create their own spans and those spans will get linked with the library generated spans.
+
+## Usage:
+
+### 1. Initialize the exporter
+
+       opentelemetry::exporter::jaeger::JaegerExporterOptions opts;
+       std::unique_ptr<opentelemetry::sdk::trace::SpanExporter> exporter = std::unique_ptr<opentelemetry::sdk::trace::SpanExporter>(
+            new opentelemetry::exporter::jaeger::JaegerExporter(opts));
+
+Above is an example of initializing a jaeger exporter. Apart from jaeger exporter, there are many other exporters supported by OpenTelelemetry.
+
+### 2. Set the Tracer Provider
+
+       auto processor = std::unique_ptr<opentelemetry::sdk::trace::SpanProcessor>(
+            new opentelemetry::sdk::trace::SimpleSpanProcessor(std::move(exporter)));
+       auto provider = opentelemetry::nostd::shared_ptr<opentelemetry::trace::TracerProvider>(
+            new opentelemetry::sdk::trace::TracerProvider(std::move(processor));
+       opentelemetry::trace::Provider::SetTracerProvider(provider);
+
+### 3. Enable tracing in Proton C++
+
+       proton::initOpenTelemetryTracer();
+
+This will generates the library spans ("amqp-message-send" and "amqp-message-received").
+
+In addition, users can create their own spans in the application. The @ref tracing_client and @ref tracing_server examples demonstrates how to do that with jaeger exporter.
+
+## Steps to run tracing examples: 
+
+    A working example of distributed tracing using broker, tracing_server and tracing_client examples.
+
+### Start Jaeger, for example: 
+
+docker run -d --name jaeger \
+  -e COLLECTOR_ZIPKIN_HOST_PORT=:9411 \
+  -p 5775:5775/udp \
+  -p 6831:6831/udp \
+  -p 6832:6832/udp \
+  -p 5778:5778 \
+  -p 16686:16686 \
+  -p 14268:14268 \
+  -p 14250:14250 \
+  -p 9411:9411 \
+  jaegertracing/all-in-one:1.25 
+
+### Build/Install Opentelemetry-cpp: 
+
+Note: Thrift is jaeger exporter dependency.
+
+1. Clone opentelemetry-cpp  
+https://github.com/open-telemetry/opentelemetry-cpp
+
+2. cd opentelemetry-cpp
+
+3. mkdir bld 
+4. cd bld
+
+5. cmake .. -DBUILD_TESTING=OFF -DBUILD_SHARED_LIBS=ON -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DWITH_JAEGER=ON 
+6. make 
+7. sudo make install 
+
+### Demo with broker.
+
+
+WORKDIR cpp/examples  
+
+1. @ref broker.cpp   
+<b>In one window run:</b>  
+`./broker`  
+
+2. @ref tracing_server.cpp  
+<b>In another window run:</b>   
+`./tracing_server`
+
+3. @ref tracing_client.cpp  
+<b>In the final window:</b>  
+`./tracing_client`
+
+### Look in the Jaeger UI: 
+
+Browse to http://localhost:16686/. This should open up a console for the Jaeger tracing system.  
+Select the Service dropdown at the top of the Search options (if not already selected). 
+Hit Find Traces. 
+
+*/
--- a/cpp/examples/tracing_client.cpp
+++ b/cpp/examples/tracing_client.cpp
@@ -1,0 +1,194 @@
+/*
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+#include "options.hpp"
+#include <proton/connection.hpp>
+#include <proton/container.hpp>
+#include <proton/delivery.hpp>
+#include <proton/message.hpp>
+#include <proton/messaging_handler.hpp>
+#include <proton/receiver_options.hpp>
+#include <proton/source_options.hpp>
+#include <proton/tracing.hpp>
+#include <proton/tracker.hpp>
+#include <proton/message_id.hpp>
+
+#include <bits/stdc++.h>
+#include <iostream>
+#include <string>
+#include <vector>
+#include <map>
+
+// Include opentelemetry header files
+#include <opentelemetry/sdk/trace/simple_processor.h>
+#include <opentelemetry/sdk/trace/tracer_provider.h>
+#include <opentelemetry/trace/provider.h>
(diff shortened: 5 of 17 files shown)


cpp/src/tracing_dummy.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+namespace proton
+{
+void on_message_start_span(message &message, delivery &d) {}
+
+void on_message_end_span(delivery &d) {}
+
+void on_settled_span(delivery &d, tracker &track) {}
+
+void send_span(const message &message, binary &tag, tracker &track) {}
```
astitcher, line 34: These functions should be changed into an abstract class to make switching between the tracing and non tracing case easier.

cpp/src/sender.cpp
```diff
@@ -70,6 +71,10 @@ tracker sender::send(const message &message, const binary &tag) {
     pn_delivery_t *dlv = pn_delivery(
         pn_object(),
         pn_dtag((reinterpret_cast<const char *>(&tag[0])), tag.size()));
+
+    proton::binary tag_cp = tag;
+    tracker track = make_wrapper<tracker>(dlv);
+    send_span(message, tag_cp, track);
```
astitcher, line 77: Here you might want to use a pointer to message to allow you to copy the message when tracing is enabled and not copy the message when it isn't.
Something like
`proton::Message* message_to_send = &message;`

astitcher (changes requested):

cpp/src/sender.cpp
```diff
@@ -67,11 +68,20 @@ tracker sender::send(const message &message) {
...
+    tracker track = make_wrapper<tracker>(dlv);
     std::vector<char> buf;
-    message.encode(buf);
+    proton::message message_cp;
+    if (isTracingEnable()) {
+        message_cp = message;
+        send_span(message_cp, tag_cp, track);
+        message_to_send = &message_cp;
+    }
+    message_to_send->encode(buf);
```
DreamPearl, line 84: Also tried without using a message pointer.
```
if (isTracingEnable()) {
    proton::message message_cp = message;
    send_span(message_cp, tag_cp, track);
    message_cp.encode(buf);
    }
 else
    message.encode(buf);
```

cpp/src/sender.cpp
DreamPearl: Done, thanks! 

Also, we need to type cast it into a non-const pointer .
`proton::message* message_to_send = (proton::message*)&message;`

cpp/src/tracing_dummy.cpp
DreamPearl: Done!

cpp/src/sender.cpp
DreamPearl: This `const proton::message* message_to_send = &message;` also worked, no need to const cast. Thanks!

cpp/src/sender.cpp
```diff
@@ -71,12 +72,17 @@ tracker sender::send(const message &message, const binary &tag) {
...
+    tracker track = make_wrapper<tracker>(dlv);
+    ot.message_encode(message, buf, tag, track);
+    
     assert(!buf.empty());
     pn_link_send(pn_object(), &buf[0], buf.size());
     pn_link_advance(pn_object());
-    if (pn_link_snd_settle_mode(pn_object()) == PN_SND_SETTLED)
+    if (pn_link_snd_settle_mode(pn_object()) == PN_SND_SETTLED) {
         pn_delivery_settle(dlv);
+    }
```
DreamPearl, line 85: Removed braces.

cpp/examples/tracing_demo.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+    void on_message(proton::delivery &d, proton::message &m) override {
+        std::cout << "Message received: " << m.body() << std::endl;
+        if (m.message_annotations().empty()) {
+            std::cout << "Message annotations is empty " << std::endl;
+        } else {
+            std::cout << "Message annotations is not empty " << std::endl;
+            std::cout << "Receive message_annotations: "
+                      << m.message_annotations() << std::endl;
+        }
```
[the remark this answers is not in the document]
DreamPearl: Done!

cpp/src/tracing_private.hpp
```diff
@@ -0,0 +1,54 @@
...
+    /// By default returns the object of DummyTracing.
+    static Tracing& getTracing();
+
+    /// Initialize the Tracing object with the OpentelemetryTracing object.
+    static void activate(Tracing& r);
+
+    virtual void message_encode(const message &m, std::vector<char> &buf, const binary &tag, const tracker &track) = 0;
+    virtual void on_message_handler(messaging_handler& h, delivery& d, message& message) = 0;
+    virtual void on_settled_span(delivery& d, tracker& track) = 0;
+    virtual void send_span(const message& message, const binary& tag, const tracker& track) = 0;
```
astitcher, line 48: This abstract method is now not needed.
What it does is now part of the OpentelemetryTracing implementation only and so should just be part of Opentelemetry::message_encode.

cpp/src/tracing_dummy.cpp
```diff
@@ -0,0 +1,54 @@
...
+        m.encode(buf);
+    }
+
+    void on_message_handler(messaging_handler& h, delivery& d, message& message) override {
+        h.on_message(d, message);
+    }
+
+    void on_settled_span(delivery& d, tracker& track) override {}
+
+    void send_span(const message& message, const binary& tag, const tracker& track) override {}
```
astitcher, line 44: shouldn't be needed - remove from parent class.

cpp/src/tracing.cpp
```diff
@@ -0,0 +1,255 @@
+/*
```
astitcher, line 1: This file should be renemaed tracing_opentelemetry.cpp as it is the Opentelemetry implementation of tracing - in theory there could be others.

cpp/src/tracing_dummy.cpp
```diff
@@ -0,0 +1,54 @@
+/*
```
astitcher, line 1: This file should be renamed as tracing_stub.cpp to be in line with naming in other places.

cpp/src/tracing_dummy.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include <proton/annotation_key.hpp>
+#include <proton/message.hpp>
+#include <proton/messaging_handler.hpp>
+
+#include "tracing_private.hpp"
+
+namespace proton
+{
+
+class DummyTracing : public Tracing {
```
astitcher, line 32: Probably should rename this to StubTracing in line with new proposed filename.

cpp/src/tracing.cpp
```diff
@@ -0,0 +1,255 @@
...
+        binary tag = d.tag();
+        nostd::shared_ptr<opentelemetry::trace::Span> span = tag_span[tag];
+        transfer tt(track);
+        std::string delivery_state = to_string(tt.state());
+        span->AddEvent("delivery state: " + delivery_state);
+
+        span->End();
+    }
+
+    void send_span(const message& message, const binary& tag, const tracker& track) override {
```
astitcher, line 172: This should be folded into message_encode and removed.

cpp/src/tracing.cpp
```diff
@@ -0,0 +1,255 @@
...
+    auto provider = opentelemetry::trace::Provider::GetTracerProvider();
+    nostd::shared_ptr<opentelemetry::trace::Tracer> tracer = provider->GetTracer("qpid-tracer");
+    return tracer;
+}
+
+class OpentelemetryTracing : public Tracing {
+  public:
+    void message_encode(const message& m, std::vector<char>& buf, const binary& tag, const tracker& track) override {
+        proton::message message_cp = m;
+        Tracing& ot = Tracing::getTracing();
```
astitcher, line 115: You never need to use getTracing from within a tracing member function as you use this there or in this case just access the member function directly.

As in `send_span(...)`

However I suggest removing send span completely.

Or if you really want to leave it then it (which would be fine) it doesn't need to be virtual or to override the parent class as it is only used here in this class and no one else needs to call it (so probably should be private as well).

cpp/src/tracing_private.hpp
```diff
@@ -0,0 +1,54 @@
...
+    /// Returns an object for the Tracing. If Tracing is enable, returns the object of OpentelemetryTracing.
+    /// By default returns the object of DummyTracing.
+    static Tracing& getTracing();
+
+    /// Initialize the Tracing object with the OpentelemetryTracing object.
+    static void activate(Tracing& r);
+
+    virtual void message_encode(const message &m, std::vector<char> &buf, const binary &tag, const tracker &track) = 0;
+    virtual void on_message_handler(messaging_handler& h, delivery& d, message& message) = 0;
+    virtual void on_settled_span(delivery& d, tracker& track) = 0;
```
astitcher, line 47: I think this may have the wrong signature you shouldn't need a delivery here at all in fact it's even really wrong as the send side only has trackers not deliveries. You should be able to get the tag from the tracker - if there is no C++ API then use the C API directly.

cpp/src/tracing.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+        opentelemetry::trace::Scope scope = get_tracer()->WithActiveSpan(span);
+
+        h.on_message(d, message);
+
+        span->End();
+    }
+
+    void on_settled_span(delivery& d, tracker& track) override {
+
```
astitcher, line 162: Shouldn't need to pass in a delivery only a tracker as this is sender side.

cpp/src/tracing.cpp
```diff
@@ -0,0 +1,255 @@
...
+        binary tag = d.tag();
+        nostd::shared_ptr<opentelemetry::trace::Span> span = tag_span[tag];
+        transfer tt(track);
+        std::string delivery_state = to_string(tt.state());
```
astitcher, line 166: I don't understand why we need a new copy of the transfer (`tt`) just to get a string version of the delivery state, we should be able to get that directly from the original `track` without copying.

cpp/examples/CMakeLists.txt
```diff
@@ -89,3 +89,10 @@ foreach(example
   add_executable(${example} ${example}.cpp)
   target_link_libraries(${example} Proton::cpp Threads::Threads)
 endforeach()
+
+if (ENABLE_OPENTELEMETRYCPP)
+  set(example tracing_demo)
+  add_executable(${example} ${example}.cpp)
+  include_directories(${CMAKE_SOURCE_DIR}/exporters/jaeger/include)
+  target_link_libraries(${example} Proton::cpp opentelemetry_trace opentelemetry_exporter_jaeger_trace)
+endif()
```
astitcher, line 98: As far as I can tell the tracing example should compile without any special attention to whether Opentelemetry is compiled into Proton or not. Obviously if there is no real tracing library then this example will do no tracing but I don't think any of the stuff in the conditional should be necessary - if it is necessary then we need to debug why and fix that.(excluding the `add_executable(..)` which I'm suggesting could just go directly into the single threaded examples foreach above.

cpp/examples/README.dox
```diff
@@ -137,3 +137,55 @@ A multithreaded sender and receiver enhanced for flow control.
...
+
+# Look in the Jaeger UI: 
+
+Browse to http://localhost:16686 
+Select the Service dropdown at the top of the Search options (if not already selected). 
+Hit Find Traces. 
+
+__Requires C++11 opentelemetry-cpp__
+
+*/
```
astitcher, line 191: I think we should consider putting this whole section of the examples readme into its own file - the instructions for running the tracing infrastructure are intricate and might perhaps confuse people who aren't interested in tracing?

cpp/examples/tracing_demo.cpp
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+};
+
+int main(int argc, char **argv) {
+    try {
+        std::string conn_url = argc > 1 ? argv[1] : "//127.0.0.1:5672";
+        std::string addr = argc > 2 ? argv[2] : "examples";
+
+        proton::initTracer(jaeger);
```
astitcher, line 102: I think it would be worthwhile calling out that this is the only thing that is necessary to make this example a tracing example.

astitcher (changes requested):
@DreamPearl More changes! As always get back to me if any of this doesn't make sense.

cpp/src/tracing.cpp
DreamPearl: Make sense. And I went with removing the send span completely.

cpp/src/tracing_private.hpp
DreamPearl: This reminds me, I worked on adding an accessor for the delivery tag on the tracker later after writing this code. So yeah now we do have C++ API for that. Thanks!

cpp/src/sender.cpp
```diff
@@ -71,7 +72,11 @@ tracker sender::send(const message &message, const binary &tag) {
         pn_object(),
         pn_dtag((reinterpret_cast<const char *>(&tag[0])), tag.size()));
     std::vector<char> buf;
-    message.encode(buf);
+
+    Tracing& ot = Tracing::getTracing();
+    tracker track = make_wrapper<tracker>(dlv);
```
astitcher, line 77: This is now duplicated below in the return line (82). There you now should just `return tracker;`

astitcher (changes requested):
A few more things to consider - also I'm wondering about thread safety and thinking that maybe there should be a tracer for each proton::container rather than there being a single global.

DreamPearl:
@astitcher  Tracing demo spans on Jaeger UI:
https://gist.github.com/DreamPearl/25935c37d97284f22639de454c623717

cpp/examples/tracing_client.cpp
```diff
@@ -0,0 +1,108 @@
...
+        proton::initOpentelemetryTracer(proton::jaeger);
+
+        client c(conn_url, addr, msg);
+
+        proton::container(c).run();
+
+        return 0;
+    } catch (const std::exception& e) {
+        std::cerr << e.what() << "\n";
+        return 1;
```
DreamPearl, line 104: Will remove this in the next PR update. (Removed locally)

astitcher:
That jaeger display looks good

cpp/examples/tracing.dox
```diff
@@ -0,0 +1,74 @@
...
+## Usage:
+
+Minimum requirements to enable tracing in proton:
+
+1. Include `<proton/tracing.hpp>`  
+2. Call `proton::initOpenTelemetryTracer()` with a service-name as an argument.
+
+
+A working example of distributed tracing using broker, server and client examples.
+
```
DreamPearl, line 20: Documentation for tracing is here having instructions for running the examples.
This will be shown on a separate page named Tracing.

cpp/src/tracing_opentelemetry.cpp
```diff
@@ -0,0 +1,273 @@
...
+namespace nostd = opentelemetry::nostd;
+namespace sdktrace = opentelemetry::sdk::trace;
+
+const std::string kContextKey = "x-opt-qpid-tracestate";
+
+// TODO: Have a delivery context to do the work, instead of having a map to associate the spans with the delivery tags.
+
+// A map to associate the spans with the delivery tags, needed for ending the spans on message settled.
+// Deleting the map entries after the span is ended to avoid the memory leakage in future.
+std::unordered_map<binary, nostd::shared_ptr<opentelemetry::trace::Span>> tag_span;
```
DreamPearl, line 79: Replaced the std::map with the std::unordered_map.

cpp/src/tracing_opentelemetry.cpp
```diff
@@ -0,0 +1,273 @@
...
+#include <string>
+#include <unordered_map>
+
+// Custom specialization of std::hash injected in namespace std for proton::binary as a key in tag_span i.e. an unordered_map.
+template <> struct std::hash<proton::binary> {
+    std::size_t operator()(const proton::binary& k) const {
+        std::string s(k[0], k.size());
+        return std::hash<std::string>{}(s);
+    }
+};
```
DreamPearl, line 66: Custom specialization of std::hash for the unordered_map having proton::binary as key.

DreamPearl:
Spans on Jaeger UI: https://gist.github.com/DreamPearl/6ccaa6b64587ec78ab93a0c8c71cbc70

cpp/examples/tracing_client.cpp
```diff
@@ -0,0 +1,166 @@
...
+        provider = opentelemetry::nostd::shared_ptr<opentelemetry::trace::TracerProvider>(
+            new opentelemetry::sdk::trace::TracerProvider(std::move(processor), resource));
+
+        // Set the global trace provider
+        opentelemetry::trace::Provider::SetTracerProvider(provider);
+
+        // Enable tracing in proton cpp
+        if (t == "tracing")
+            proton::initOpenTelemetryTracer();
+
```
DreamPearl, line 148: The new approach to enable tracing.

cpp/examples/tracing_server.cpp
```diff
@@ -0,0 +1,143 @@
...
+        auto processor = std::unique_ptr<opentelemetry::sdk::trace::SpanProcessor>(
+            new opentelemetry::sdk::trace::SimpleSpanProcessor(std::move(exporter)));
+        provider = opentelemetry::nostd::shared_ptr<opentelemetry::trace::TracerProvider>(
+            new opentelemetry::sdk::trace::TracerProvider(std::move(processor), resource));
+
+        // Set the global trace provider
+        opentelemetry::trace::Provider::SetTracerProvider(provider);
+    
+        // Enable tracing in proton cpp
+        if (t == "tracing")
```
DreamPearl, line 131: Note: No longer need this line as this example is solely for demonstrating tracing in proton. Same for the tracing_client.

DreamPearl:
In middle of improving the tracing examples and documentation.

cpp/CMakeLists.txt
```diff
@@ -36,6 +36,19 @@ else()
...
 
+# Check for OPENTELEMETRY-CPP for distributed tracing
+find_package(opentelemetry-cpp)
+option(ENABLE_OPENTELEMETRYCPP "Use opentelemetry for distributed tracing" ${OPENTELEMETRY-CPP_FOUND})
+
+if (ENABLE_OPENTELEMETRYCPP)
+  include_directories(${OPENTELEMETRY_CPP_INCLUDE_DIRS})
+  set(TRACING_SRC src/tracing_opentelemetry.cpp src/tracing_stub.cpp)
+  set(TRACING_LIBS opentelemetry-cpp::trace opentelemetry-cpp::jaeger_trace_exporter opentelemetry-cpp::ostream_span_exporter)
+  find_package(CURL REQUIRED)
```
astitcher, line 47: Not needed now with otel 1.5.1

cpp/CMakeLists.txt
```diff
@@ -36,6 +36,19 @@ else()
...
 endif()
 
+# Check for OPENTELEMETRY-CPP for distributed tracing
+find_package(opentelemetry-cpp)
+option(ENABLE_OPENTELEMETRYCPP "Use opentelemetry for distributed tracing" ${OPENTELEMETRY-CPP_FOUND})
+
+if (ENABLE_OPENTELEMETRYCPP)
+  include_directories(${OPENTELEMETRY_CPP_INCLUDE_DIRS})
+  set(TRACING_SRC src/tracing_opentelemetry.cpp src/tracing_stub.cpp)
+  set(TRACING_LIBS opentelemetry-cpp::trace opentelemetry-cpp::jaeger_trace_exporter opentelemetry-cpp::ostream_span_exporter)
```
astitcher, line 46: might not need any of the exporters here

cpp/examples/tracing_client.cpp
```diff
@@ -0,0 +1,179 @@
...
+        opentelemetry::exporter::jaeger::JaegerExporterOptions opts;
+
+        // Initialize Jaeger Exporter
+        std::unique_ptr<opentelemetry::sdk::trace::SpanExporter> exporter = std::unique_ptr<opentelemetry::sdk::trace::SpanExporter>(
+            new opentelemetry::exporter::jaeger::JaegerExporter(opts));
+
+        // Set service-name
+        auto resource_attributes = opentelemetry::sdk::resource::ResourceAttributes
+        {
+            {"service.name", "qpid-client"}
```
astitcher, line 146: probably use qpid-example-client as service name

cpp/examples/tracing_server.cpp
```diff
@@ -0,0 +1,148 @@
...
+        // Initialize Jaeger Exporter
+        opentelemetry::exporter::jaeger::JaegerExporterOptions opts;
+        std::unique_ptr<opentelemetry::sdk::trace::SpanExporter> exporter = std::unique_ptr<opentelemetry::sdk::trace::SpanExporter>(
+            new opentelemetry::exporter::jaeger::JaegerExporter(opts));
+
+        // Set service-name
+        auto resource_attributes = opentelemetry::sdk::resource::ResourceAttributes
+        {
+            {"service.name", "qpid-server"}
+        };
```
astitcher, line 123: qpid-example-server

astitcher (commented):
looks good - some small nits

DreamPearl:
@astitcher Made modifications in the examples (tracing_client and tracing_server).

Spans on Jaeger UI: https://gist.github.com/DreamPearl/de4c6d317fefe8e35a0f551bc4785b3b

cpp/examples/tracing_client.cpp
```diff
@@ -0,0 +1,204 @@
...
+        opentelemetry::nostd::shared_ptr<opentelemetry::trace::Span> span = tracer->StartSpan("send_request",
+            {{"reply_to", req.reply_to()}, {"message", to_string(req.body())}},
+            options);
+        opentelemetry::trace::Scope scope = tracer->WithActiveSpan(span);
+
+        // Storing the 'scope' in a map to keep it alive and erasing it when a response is received.
+        scope_map[req.id()] = std::make_shared<opentelemetry::trace::Scope>(std::move(scope));
+        id_counter++;
+
+        sender.send(req);
```
DreamPearl, line 99: This `send_request` span will show the complete request and response cycle.

ci/otel.sh
```diff
@@ -0,0 +1,43 @@
...
+
+# OTel dependencies
+
+sudo apt-get install libcurl4-openssl-dev
+sudo apt-get install libboost-locale-dev
+sudo apt-get install libthrift-dev
+
+# Clone OpenTelemetry-cpp
+
+git clone -b v1.6.1 --recursive https://github.com/open-telemetry/opentelemetry-cpp
```
DreamPearl, line 32: @astitcher With v1.6.1 tests are passing.

astitcher (approved):
Just realised the docs are now wrong!
I ok with merging the code - but it would be good to have more correct docs!

astitcher:
Merged

PR #360, opened 2022-03-02
PROTON-2511: Fix Fatal Python error: deallocating None

tanzhiying:
Fatal Python error: deallocating None

### DIFF
--- a/python/cproton.i
+++ b/python/cproton.i
@@ -238,6 +238,7 @@ PN_HANDLE(PNI_PYTRACER);
 %typemap(out) pn_msgid_t {
   switch ($1.type) {
     case PN_NULL:
+      Py_INCREF(Py_None);
       $result = Py_None;
       break;
     case PN_ULONG:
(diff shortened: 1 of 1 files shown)


jiridanek:
@astitcher I think that this is useful fix and I saw a similar problem in Dispatch as well, https://issues.apache.org/jira/browse/DISPATCH-2316. I am not sure if 0.37.0 should be released without this having been fixed. I haven't reproduced this for myself to see how likely it is to arise in practice, but my feeling is that it is likely and very inconvenient when it happens.

In Dispatch, this happened during router shutdown. I think it will be a similar case here, it will crash Python when client is exiting and Python runs a final GC.

tanzhiying:
你好，我已收到您的邮箱，尽快回复                                      -----谭志颖

astitcher:
Oops, this is a potentially nasty bug! I should check that we don't make this mistake anywhere else.

astitcher (approved):

PR #362, opened 2022-03-14
PROTON-2517: rewind all pn_data about to be used in emitters.h

cliffjansen:
Intercepts all pn_data_t emitter functions and does rewind.

In array case, also points to first node so that pn_data_type and pn_data_get_array work properly.

These are consistent with the old codec.

Not done: save and restore of pn_data_t state at entry and exit on the assumption that the callers of the emmiter functions do not expect this state preserved.

### DIFF
--- a/c/src/core/emitters.h
+++ b/c/src/core/emitters.h
@@ -534,12 +534,23 @@ static inline void emit_copy(pni_emitter_t* emitter, pni_compound_context* compo
   if (!data || pn_data_size(data) == 0) {
     pni_emitter_writef8(emitter, PNE_NULL);
   } else {
+    pn_handle_t point = pn_data_point(data);
+    pn_data_rewind(data);
     pni_emitter_data(emitter, data);
+    pn_data_restore(data, point);
   }
   compound->count++;
 }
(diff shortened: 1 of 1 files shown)


astitcher:
The save and restore state is needed here because (at least some of) the pn_data_t values used here come from the user and are not internal to the proton implementation. And the user left the pn_data_t in one state and we can't leave it silently in another.
Note that this is true even if the pn_data_t is say an array of symbols representing a capability list - The current proton API just gives the user the pn_data_t to modify and the user the user modifies this pn_data_t but at no point 'gives it back' to proton. This is a really crappy API, but there we are. The user can assume that it can just keep on modifying this pn_data_t, so we do have to maintain the cursor state.
The area of where pn_data_t is exposed for this kind of thing is on my near term list of things to fix!

astitcher (changes requested):
As noted in the other comment I think that saving and restoring the cursor is needed.

astitcher (approved):
Yep, I like this version of the changes.
It doesn't interfere with my thoughts on null omission either.

PR #366, opened 2022-04-07
PROTON-2529 CMakeLists.txt: fix build without a working C++ compiler

ffontaine:
Don't call `check_language(CXX)` if `BUILD_CPP` is not set to avoid the following if a non working C++ compiler is installed on the system:

```
CMake Error at /nvmedata/autobuild/instance-25/output-1/host/share/cmake-3.18/Modules/CMakeTestCXXCompiler.cmake:59 (message):
  The C++ compiler

    "/usr/bin/c++"

  is not able to compile a simple test program.
```

Fixes:
 - http://autobuild.buildroot.org/results/7bf05fb0208fa09543f841ec6e7adc6e4d984b0c

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,18 +78,20 @@ set (C_STANDARD_FLAGS ${C_STANDARD_${CMAKE_C_COMPILER_ID}})
 set (C_EXTENDED_FLAGS ${C_EXTENDED_${CMAKE_C_COMPILER_ID}})
 
 ## C++
-check_language (CXX)
-if (CMAKE_CXX_COMPILER)
-  enable_language(CXX)
+if (BUILD_CPP)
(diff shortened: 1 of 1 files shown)


ffontaine:
Unfortunately, as spotted by CI, this doesn't work as the default value of `BUILD_CPP` is set through:

```
if (CMAKE_CXX_COMPILER)
  set (DEFAULT_CPP ON)
endif()
```

I'll have to think about it perhaps I'll just had a C++ dependency to qpid-proton in buildroot.

astitcher:
This change won't work - you actually do need a C++ compiler to run the tests so the CMake build configuration is correct. So yes you actually do need to make a build dependency on c++.

PR #368, opened 2022-04-21
Object/remove reify

astitcher:
PR in case you're interested in some preliminary work to simplify and hopefully speed up the object system in proton


astitcher:
Closed in favor of #369 which is the continuation of this work.

PR #369, opened 2022-04-29
Object system speedups and simplifications

astitcher:
This PR incorporates the work in #368 and adds some more simplification of the object system to speed up my benchmarks by 10-15% compared to the main branch commit it is based on.

The speed ups are due to:
* Removing unnecessary reify operations
* Inlining a lot of the object mechanism
* Devirtualizing most of the object function lookups.
* Something else...?

### DIFF
--- a/c/src/core/object/object.c
+++ b/c/src/core/object/object.c
@@ -26,13 +26,14 @@
 #include <stdlib.h>
 #include <assert.h>
 
-#define pn_object_initialize NULL
-#define pn_object_finalize NULL
-#define pn_object_inspect NULL
-#define pn_object_hashcode NULL
-#define pn_object_compare NULL
+#define CID_pn_default CID_pn_object
+#define pn_default_initialize NULL
+#define pn_default_finalize NULL
+#define pn_default_inspect NULL
+#define pn_default_hashcode NULL
+#define pn_default_compare NULL
 
-const pn_class_t PN_OBJECT[] = {PN_CLASS(pn_object)};
+const pn_class_t PN_DEFAULT[] = {PN_CLASS(pn_default)};
 
 void *pn_void_new(const pn_class_t *clazz, size_t size) { return pni_mem_allocate(clazz, size); }
 #define pn_void_initialize NULL
@@ -334,7 +335,7 @@ const pn_class_t *pn_class(void *object)
   if (object) {
     return pni_head(object)->clazz;
   } else {
-    return PN_OBJECT;
+    return PN_DEFAULT;
   }
 }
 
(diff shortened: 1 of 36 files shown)


astitcher:
I'm pretty happy with the state of this PR. I think the commits need to be somewhat changed for coherence and comprehension. But it is essentially ready to merge

c/src/core/object/object.c
```diff
@@ -58,50 +67,130 @@ pn_cid_t pn_class_id(const pn_class_t *clazz)
...
+  if (head != NULL) {
+    object = head + 1;
+    head->clazz = clazz;
+    head->refcount = 1;
+  }
+  return object;
+}
+
+static inline void pni_object_incref(void *object) {
+  if (object) {
```
ssorj, line 83: Should this assert instead of null check?  _refcount and _decref are (_decref needs an explicit assert, I think).  Maybe the null check in pn_object_incref, not pni_object_incref?

In general, I'd prefer all the refcounting methods required a non-null object, but that does require fixing up a number of places.

c/src/core/object/object.c
```diff
@@ -58,50 +67,130 @@ pn_cid_t pn_class_id(const pn_class_t *clazz)
...
+}
+
+static inline void pni_class_incref(const pn_class_t *clazz, void *object) {
+  if (clazz->incref) {
+    clazz->incref(object);
+  } else {
+    pni_object_incref(object);
+  }
+}
+
```
ssorj, line 124: Can we have both of these use the if/else form?  I find it easier to read.

ssorj: Same for pni_class_refcount below.

c/src/core/object/object.c
```diff
@@ -58,50 +67,130 @@ pn_cid_t pn_class_id(const pn_class_t *clazz)
...
+static inline int pni_class_refcount(const pn_class_t *clazz, void *object) {
+  return clazz->refcount
+  ? clazz->refcount(object)
+  : pni_object_refcount(object);
+}
+
+static inline void pni_class_free(const pn_class_t *clazz, void *object) {
```
ssorj, line 139: A bigger question about naming conventions.  These are static in a library file.  Should they use pni_?  In some other places, they have no prefix.  (I've long been confused about this.)

c/src/core/object/object.c
```diff
@@ -58,50 +67,130 @@ pn_cid_t pn_class_id(const pn_class_t *clazz)
...
   return object;
 }
 
 void *pn_class_incref(const pn_class_t *clazz, void *object)
 {
-  assert(clazz);
   if (object) {
-    clazz = clazz->reify(object);
-    clazz->incref(object);
+    if (clazz==PN_OBJECT) {
```
ssorj, line 160: Here and elsewhere, I would prefer space around ==.

ssorj (approved):
I get a very nice ~12 % speedup from this patch.  It also simplifies the object code quite a bit, which I love.

I have some clerical, naming, and contract-related notes.

c/src/core/object/object.c
astitcher: pn_class_incref/decref are actually allowed to be called with a NULL object so they really do need to check for null and do nothing if they get that - the other inlined functions may not need the check.

c/src/core/object/object.c
astitcher: I personally find the ternary easier to read! But I'm happy to change it.

c/src/core/object/object.c
astitcher: These lines have gone anyway now.

c/src/proactor/win_iocp.cpp
```diff
@@ -1514,7 +1514,7 @@ void pni_iocp_initialize(void *obj)
   pni_shared_pool_create(iocp);
   iocp->completion_port = CreateIoCompletionPort(INVALID_HANDLE_VALUE, NULL, 0, 0);
   assert(iocp->completion_port != NULL);
-  iocp->zombie_list = pn_list(PN_OBJECT, 0);
+  iocp->zombie_list = pn_list(PN_CLASSCLASS(pni_iocpdesc), 0);
```
ssorj, line 1517: I noticed that this one is not like the others, so checking if it's as it should be.  The others pass the address of PN_CLASSCLASS.

astitcher: Good spot - that's probably my windows build failure right there - thanks!

PR #370, opened 2022-05-04
PROTON-2540: [cpp] Provide a way to query proton::connection for the url

DreamPearl:
Right now, there are only two ways:

1. Save the URL where it will always guarantee to return the URL for the current connection.
2. Reconstruct the url using host, port, etc, but right now lib does not store all the URL parts.

So, either store the remaining parts of the URL for reconstructing the whole URL or save the entire URL itself.

Went with the first approach in this PR.

### DIFF
--- a/cpp/src/connection.cpp
+++ b/cpp/src/connection.cpp
@@ -93,6 +93,12 @@ session_range connection::sessions() const {
     return session_range(session_iterator(make_wrapper(pn_session_head(pn_object(), 0))));
 }
 
+std::string connection::url() const {
+    connection_context& cc = connection_context::get(pn_object());
+    if (!active()) throw proton::error("No active connection");
+    return cc.active_url_;
+}
+
 receiver_range connection::receivers() const {
   pn_link_t *lnk = pn_link_head(pn_object(), 0);
   while (lnk) {
--- a/cpp/src/reconnect_test.cpp
+++ b/cpp/src/reconnect_test.cpp
@@ -41,6 +41,10 @@
 
 namespace {
 
+// The actual port used for making a url.
+// Gets updated whenever a connection is open on server side.
+int listening_port_;
+
 // Wait for N things to be done.
 class waiter {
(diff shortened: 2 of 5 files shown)


cpp/src/connection.cpp
```diff
@@ -93,6 +93,12 @@ session_range connection::sessions() const {
     return session_range(session_iterator(make_wrapper(pn_session_head(pn_object(), 0))));
 }
 
+std::string connection::url() const {
+    connection_context& cc = connection_context::get(pn_object());
+    if(!active()) throw proton::error("No active connection");
```
astitcher, line 98: type: No space after 'if'

astitcher, line 98: I think this is an open question - what to do when the connection is not active: Could throw an exception as this code does, or somehow return nullptr maybe but then would need to change the return type and returning an pointer isn't a good idea overall. Or maybe return an option, but we don't do this anywhere else/

The issue here is whether it will be a bad user experience to receive an exception in the non active case.

astitcher (commented):
This looks basically good

cpp/src/reconnect_test.cpp
```diff
@@ -129,10 +134,13 @@ class server_connection_handler : public proton::messaging_handler {
 
 class tester_base: public proton::messaging_handler {
   void on_connection_open(proton::connection& c) override {
+    std::string want_url = "amqp://localhost:" + std::to_string(listening_port_);
+
     if (!c.reconnected()) {
       start_count++;
       c.open_sender("messages");
     }
+    ASSERT_EQUAL(c.url(), want_url);
```
DreamPearl, line 143: @astitcher A minor change from our last discussion. CMIIW

Later I realized that c.active() will always be true in on_connection_open(), so no need of adding an assertion for an exception here for the case when the connection is not active.
Hence, here just checking for the URL.

astitcher:
Merged

PR #377, opened 2022-07-29
Update Travis CI to stop using Xenial and use Focal instead

astitcher:
Xenial is over 5 years old now - I don't think we need to make sure we build on it any longer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,81 +17,59 @@
 # under the License
 #
 
+os: linux
+dist: focal
+language: cpp
+env:
+  global:
+    - PYTHON=python3
 jobs:
   include:
-  - os: linux
-    dist: xenial
-    language: cpp
-    compiler: gcc
-    env:
-    - OPENSSL_ia32cap='0x00000000'
-  - os: linux
-    dist: xenial
-    language: cpp
-    compiler: clang
-    env:
-    - OPENSSL_ia32cap='0x00000000'
+  - compiler: gcc
+  - compiler: clang
     # c-threaderciser test hangs on older clang
+    env:
     - QPID_PROTON_CMAKE_ARGS='-DENABLE_LINKTIME_OPTIMIZATION=OFF -DTHREADERCISER=ON'
     - QPID_PROTON_CTEST_ARGS='--exclude-regex c-threaderciser'
   - name: static libs
-    os: linux
-    dist: focal
-    language: cpp
     compiler: gcc
     env:
-    - PYTHON=python3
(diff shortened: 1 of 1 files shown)


astitcher:
I was hoping to get Travis to build this PR but having force pushed a change it looks like it's not doing another build! :-(

.travis.yml
```diff
@@ -17,81 +17,57 @@
...
-    - OPENSSL_ia32cap='0x00000000'
-  - os: linux
-    dist: xenial
-    language: cpp
-    compiler: clang
-    env:
-    - OPENSSL_ia32cap='0x00000000'
+  - compiler: gcc
+  - compiler: clang
     # c-threaderciser test hangs on older clang
```
jiridanek, line 29: > I was hoping to get Travis to build this PR but having force pushed a change it looks like it's not doing another build! :-(

Could be due to syntax error, you are missing `env:` here!

Also, I have my doubts about few other things in this file. There should be a travis syntax checker somewhere! https://github.com/travis-ci/travis.rb#lint maybe, but I hoped for something online, where I can paste file and it will give me report.

astitcher: Thanks for that syntax checking! It seems to have launched a travis job now.

astitcher:
@jiridanek Did you have anything else you wanted to change in the travis job?

jiridanek:
> @jiridanek Did you have anything else you wanted to change in the travis job?

Nothing urgent. Upgrade clang version in that job where it explicitly installs clang from apt repo, maybe. When that was added, it installed the latest clang, but now there were few clang releases in between. But that is something to be done in the future, not in this PR.

astitcher:
@jiridanek so it still fails because it seems that the specific matrix entries override the entire global env instead of adding to it so python3 can't be common using an env var if different jobs have env vars.

istm that there should be a way to just add env vars for matrix jobs - any ideas?

jiridanek:
> @jiridanek so it still fails because it seems that the specific matrix entries override the entire global env instead of adding to it so python3 can't be common using an env var if different jobs have env vars.
> 
> istm that there should be a way to just add env vars for matrix jobs - any ideas?

If there is shared `script` block or something, maybe you can do `export PYTHON=python3`? Otherwise I think you have to repeat the variable in every row of the matrix, such as the PROTON_VERSION=main in https://github.com/jiridanek/skupper-router/blob/main/.travis.yml#L31

astitcher:
Figured it out from the docs: you can add a global stanza to the env block

PR #379, opened 2022-09-21
PROTON-2438: [cpp] Cancellable tasks

DreamPearl:
[PROTON-2438](https://issues.apache.org/jira/browse/PROTON-2438)

### DIFF
--- a/cpp/src/container_test.cpp
+++ b/cpp/src/container_test.cpp
@@ -40,6 +40,7 @@
 #include <thread>
 #include <mutex>
 #include <condition_variable>
+#include <chrono>
 
 
 namespace {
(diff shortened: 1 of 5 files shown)


DreamPearl:
@astitcher

cpp/src/container_test.cpp
```diff
@@ -574,6 +574,45 @@ void test_container_mt_close_race() {
...
+    }
+
+public:
+    schedule_cancel(): w1_state(0), w2_state(0) {}
+
+    int w1_state;
+    int w2_state;
+};
+
+int test_schedule_cancel() {
```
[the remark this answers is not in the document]
astitcher: The tests could be improved:
* Add in some negative testing:
**  test canceling invalid work handle
**  test cancelling a cancelled work item
**  test cancelling an work item that has already executed
* Cross thread cancelling

astitcher (changes requested):
lgtm except for the testing changes

cpp/src/container_test.cpp
```diff
@@ -574,6 +575,119 @@ void test_container_mt_close_race() {
...
+
+    void on_connection_close(proton::connection &c) override {
+        // Cross-thread cancelling
+
+        ASSERT(w4_state==0);
+        // Cancel the fourth task before its execution.
+        c.container().cancel(w4_handle);
+
+        // To make sure fifth task has been executed by now.
+        std::this_thread::sleep_for(std::chrono::milliseconds(50));
```
astitcher, line 656: You can't sleep in a event callback - it messes everything up!!
Instead you need to schedule an event and then wake up the connection in the event and handle the rest of this code in that event.

astitcher (changes requested):
Need to fix sleep in on_connection_close callback

cpp/src/container_test.cpp
```diff
@@ -574,6 +575,116 @@ void test_container_mt_close_race() {
...
+        w2_handle = c.schedule(proton::duration(260), proton::make_work(&schedule_cancel::change_w2_state, this, &c));
+
+        // We will not cancel this scheduled task.
+        w3_handle = c.schedule(proton::duration(35), proton::make_work(&schedule_cancel::change_w3_state, this, &c));
+
+        // We will try to cancel this task before its execution from different thread i.e connection thread.
+        w4_handle = c.schedule(proton::duration(270), proton::make_work(&schedule_cancel::change_w4_state, this, &c));
+
+        // We will try to cancel this task after its execution from different thread i.e. connection thread.
+        w5_handle = c.schedule(proton::duration(0), proton::make_work(&schedule_cancel::change_w5_state, this, &c));
```
DreamPearl, line 629: @astitcher Scheduling the task to execute after 0 ms.

cpp/src/container_test.cpp
```diff
@@ -40,6 +40,7 @@
 #include <thread>
 #include <mutex>
 #include <condition_variable>
+#include <chrono>
```
DreamPearl, line 43: TODO: Don't need this now.  -_-

astitcher: I'm confused - This still seems to be in the PR though

gemmellr:
Can you close the PR, the change looks to have been landed as 9a89ff4ea, minus one removal that was commented on above.

DreamPearl:
PR merged as https://github.com/apache/qpid-proton/commit/9a89ff4eadee92ce014616492675b1d7195db4e2.

cpp/src/proactor_container_impl.cpp
```diff
@@ -532,7 +540,13 @@ void container::impl::run_timer_jobs() {
     // We've now taken the tasks to run from the deferred tasks
     // so we can run them unlocked
     // NB. We copied the due tasks in reverse order so execute from end
-    for (int i = tasks.size()-1; i>=0; --i) tasks[i].task();
+
+    for (int i = tasks.size()-1; i>=0; --i) {
+        if(is_active_.count(tasks[i].w_handle)) {
+            tasks[i].task();
+            is_active_.erase(tasks[i].w_handle);
```
zlomek, line 547: @DreamPearl 
I believe, there is a race condition in reading / writing `is_active_` here while modifying it in `schedule()` or `cancel()` above.
Could you please check?

zlomek: Maybe, `is_active_` could be checked / modified while generating `tasks` vector in the critical section under `deferred_lock_` locked, and `tasks` then would contain only the active tasks?

zlomek:
```
void container::impl::run_timer_jobs() {
    std::vector<work> tasks;

    // We first extract all the runnable tasks and then run them -  this is to avoid having tasks
    // injected as we are running them (which could potentially never end)
    {
        GUARD(deferred_lock_);

        // Figure out how many tasks we need to execute and pop them to the back of the queue
        for (timestamp now = timestamp::now(); deferred_.size() > 0; deferred_.pop_back()) {
            // Is the next task in the future?
            timestamp next_time = deferred_.front().time;
            if ( next_time>now ) {
                pn_proactor_set_timeout(proactor_, (next_time-now).milliseconds());
                break;
            }

            std::pop_heap(deferred_.begin(), deferred_.end());

            if (is_active_.erase(deferred_.back().w_handle))
                tasks.push_back(std::move(deferred_.back().task));
        }
    }

    // We've now taken the tasks to run from the deferred tasks
    // so we can run them unlocked
    for (work task : tasks) task();
}
```

DreamPearl: @zlomek Thank you for raising this!
Would you like to raise a PR having a fix if you are interested? Otherwise, I can.

zlomek: https://github.com/apache/qpid-proton/pull/421

PR #381, opened 2022-11-01
NO-JIRA: add workflow_dispatch trigger, and update various actions versions to resolve deprecation warnings

jiridanek:


gemmellr:
Since I actually made most of the changes originally (but using v3 instead of v4 setup-python as I did avoids the MacOS failure to find OpenSSL I hit,  though still leaves some of the deprecations in play) this seems good for now. We can look at how to resolve the remaining deprecations separately. Merged.

PR #382, opened 2022-11-25
PROTON-2654: [cpp] Tests for newly added map constructors

DreamPearl:
[PROTON-2654](https://issues.apache.org/jira/browse/PROTON-2654)

Tests for https://github.com/apache/qpid-proton/pull/353

### DIFF
--- a/cpp/src/map_test.cpp
+++ b/cpp/src/map_test.cpp
@@ -114,6 +114,20 @@ void test_value() {
     ASSERT_THROWS(conversion_error, m.value(bad));
 }
 
+void test_copy_constructor() {
+    proton::map<string, scalar> m = std::map<string, scalar>{{"a", 2}, {"b", 3}};
+    ASSERT_EQUAL(scalar(2), m.get("a"));
+    ASSERT_EQUAL(scalar(3), m.get("b"));
+    ASSERT_EQUAL(2U, m.size());
+}
+
+void test_initializer_list_constructor() {
+    proton::map<string, scalar> m({{"a", 2}, {"b", 3}});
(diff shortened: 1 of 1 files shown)


DreamPearl:
@astitcher @gemmellr

cpp/src/map_test.cpp
```diff
@@ -114,6 +114,23 @@ void test_value() {
     ASSERT_THROWS(conversion_error, m.value(bad));
 }
 
+void test_copy_constructor() {
+    std::map<string, scalar> sm;
+    sm["a"] = 2;
+    sm["b"] = 3;
```
astitcher, line 120: Wouldn't this be clearer as:
`std::map<string, scalar> m = {{"a", 2}, {"b", 3}};`

astitcher: or even (syntax might not be 100% correct):
`auto m = std::map<string,scalar>{{"a", 2}, {"b", 3}};`

cpp/src/map_test.cpp
```diff
@@ -114,6 +114,23 @@ void test_value() {
...
+    proton::map<string, scalar> m(sm);
+    ASSERT_EQUAL(scalar(2), m.get("a"));
+    ASSERT_EQUAL(scalar(3), m.get("b"));
+    ASSERT_EQUAL(2U, m.size());
+}
+
+void test_initializer_list_constructor() {
+    proton::map<string, scalar> m = {{"a", 2}, {"b", 3}};
```
astitcher, line 128: as above this could be:
`auto m = proton::map<string,scalar>{{"a", 2}, {"b", 3}};`
Not sure this is definitely better, but perhaps more idiomatic post c++11.

PR #383, opened 2022-12-08
PROTON-2653: [cpp] Move tracing.dox to docs folder

DreamPearl:
[PROTON-2653](https://issues.apache.org/jira/browse/PROTON-2653)

Tracing documentation(example/tracing.dox) wasn't getting picked up by the qpid-site scripts as per already defined paths. Hence, it was not visible on the qpid website.
https://github.com/apache/qpid-site/blob/asf-site/scripts/gen-proton-release-docs#L58-L61


gemmellr:
In this case as the doc is as much about the example, having it alongside the example seems to make sense to me, so updating the site scripting instead of moving the file may be the way to go. I have actually just done that now, so I could fix things for the published site for 0.38.0, if you give it a reload youll find the 0.38.0 tracing doc is now present.

Lets see what others think about where the file should go in the next release.

astitcher:
I think that this particular bit of doc should probably be in the top level doc directory. But there isn't an explanation of these examples in the README.dox file in the examples and this should be added so as to explain what these files are.

DreamPearl:
Merged as https://github.com/apache/qpid-proton/commit/21280a08a9c673393fb7dd01dfd2f05f36378722.

PR #384, opened 2022-12-21
PROTON-2663 Add a constructor to ssl_client_options class

markohrastovec:
PROTON-2663

Please consider to incorporate this patch to Qpid Proton Cpp. Without this patch, Qpid Proton Cpp is a blocker for our project.


astitcher:
Thank you for this PR - I've now merged it into the main branch and it should be in the 0.39 release

PR #386, opened 2023-01-18
Use CFFI for python bindings

astitcher:
This PR changes the python binding to use CFFI instead of SWIG.

It passes all the proton python tests and runs all of the python examples


astitcher:
This version of the CFFI work has been tested against the in tree proton tests, the skupper-router tests and runs all the python examples.
There does seem to be some issue with building on appveyor under VS 2015 and VS 2017 that seems a little odd and does appear to be related to changes in the python build.

astitcher:
There is also a fairly consistent issues with the memory leak test in the python-integration tests running under Ubuntu on CI which I don't see in my own testing - this may be to do with the extra load in a typical CI environment.

astitcher:
Merged

PR #387, opened 2023-02-22
PROTON-2183: [cpp] Fix aborted messages exception.

DreamPearl:
[PROTON-2183](https://issues.apache.org/jira/browse/PROTON-2183)
Handled the aborted messages.

### DIFF
--- a/cpp/src/messaging_adapter.cpp
+++ b/cpp/src/messaging_adapter.cpp
@@ -118,7 +118,12 @@ void on_delivery(messaging_handler& handler, pn_event_t* event) {
 
     if (pn_link_is_receiver(lnk)) {
         delivery d(make_wrapper<delivery>(dlv));
-        if (!pn_delivery_partial(dlv) && pn_delivery_readable(dlv)) {
+        if (pn_delivery_aborted(dlv)) {
+            pn_delivery_settle(dlv);
+            pn_link_flow(lnk, 1);
+            pn_link_advance(lnk);
+        }
+        else if (!pn_delivery_partial(dlv) && pn_delivery_readable(dlv)) {
             // generate on_message
             pn_connection_t *pnc = pn_session_connection(pn_link_session(lnk));
             connection_context& ctx = connection_context::get(pnc);


jiridanek (commented):
In https://issues.apache.org/jira/browse/PROTON-2183, I have the following note as a comment

> Triage note: missing feature, cpp does not handle aborted messages (it would have to drop the message and update credit, unless draining).

So, I am wondering if there is a special case needed for the draining state.

Also, I would be nice to have a test for this.

astitcher:
> In https://issues.apache.org/jira/browse/PROTON-2183, I have the following note as a comment
> 
> > Triage note: missing feature, cpp does not handle aborted messages (it would have to drop the message and update credit, unless draining).
> 
> So, I am wondering if there is a special case needed for the draining state.

I think the same would apply in general wouldn't it? and there is nothing special for normal deliveries.

> 
> Also, I would be nice to have a test for this.

I agree with this - there is code in the C examples to generate aborted messages and you'd need to use something like that to test this as there is no way to abort messages in C++

gemmellr:
> 
> So, I am wondering if there is a special case needed for the draining state.
> 

Yep, just dont grant the replacement credit when draining.

astitcher:
> > So, I am wondering if there is a special case needed for the draining state.
> 
> Yep, just dont grant the replacement credit when draining.

That seems to make sense except I can't see anywhere else in the C++ or python code that we do this check, perhaps the C code takes care of it? Or maybe it's a bug everywhere (that's quite possible!)

gemmellr:
If the C core is already taking care of the credit handling for CPP abort handling (I dont know..?) then that would be fine, and perhaps suggest the C++ bit shouldn't grant the replacement credit at any time. But we dont grant it while draining as the whole point of the drain is to use credit.

DreamPearl:
PR merged as https://github.com/apache/qpid-proton/commit/058e2928dad9d016fe14a8b6690547e39cdfd0fd.

TODO: Add test (in-progress)

PR #388, opened 2023-03-01
NO-JIRA: [Python] Allow configuration of container ID

Jmennius:
This allows simple external configuration of container ID.
Without it the only way to overwrite or provide custom container ID is to inherit the class.

### DIFF
--- a/python/proton/_reactor.py
+++ b/python/proton/_reactor.py
@@ -1192,7 +1192,11 @@ class Container(Reactor):
     for creating connections and sender- or receiver- links.
     """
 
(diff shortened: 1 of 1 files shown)


astitcher:
@Jmennius I've merged a neater version of this change with a test to make sure it works. Note you'll likely get a faster response at least from me) if you include a test or two for your changes.

Jmennius:
> @Jmennius I've merged a neater version of this change with a test to make sure it works. Note you'll likely get a faster response at least from me) if you include a test or two for your changes.

Thanks! will consider fort the future.

PR #323, opened 2021-06-25
PROTON-2401: on_link_closed should be closed when the link is remotel…

NS-SlaFleur:
…y closed (closing state) and then locally closed (closed state).

https://issues.apache.org/jira/browse/PROTON-2401

PR #391, opened 2023-04-05
PROTON-2401: [Python] on_xxx_closed should be called on locally closed after remote close

astitcher:
When links/sessions/connections are remotely closed then locally closed the on_xxx_closed event should still fire. The existing code only fires this event for local then remote close.

This is the continuation of #323

### DIFF
--- a/python/proton/_handlers.py
+++ b/python/proton/_handlers.py
@@ -406,6 +406,10 @@ class EndpointStateHandler(Handler):
             self.on_link_closing(event)
         event.link.close()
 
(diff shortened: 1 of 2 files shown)


astitcher:
I've modified the original PR:
* The same logic should apply to the connection and session events too
* No need to call xxx_close as you'll only get the _close event after the xxx_xlose API is called already

I'd like to add some use of these event/some tests to be sure this change is good before merging it.

astitcher:
I've added a test for the connection closed event.

PR #392, opened 2023-04-06
Fixing a leak problem.

ozdemirda:
If the delivery object is not settled on the sender side, the delivery object does not get returned to the pool of the link. I believe that this issue was missed, because the delivery objects in this specific example are not completely lost and they get freed when the connection is closed (Valgrind happily reports that all is good). Without the proposed fix, this code will mislead people and the code written based on this example will cause memory exhaustion in the long run.


astitcher:
Merged

PR #394, opened 2023-04-19
PROTON-2716: replace `include_directories()` call with opentelemetry-cpp provided targets

jiridanek:
MSVC apparently tried something to exclude 3rd party libs from strict compilation warning settings, https://devblogs.microsoft.com/cppblog/broken-warnings-theory/, and gcc/cmake should have this too, all IMPORTED targets include directories with -isystem, which should suppress excessive warnings https://stackoverflow.com/a/73549820/1047788..

[embarrassingly many minutes later] Aand, I think I found it, we have if (ENABLE_OPENTELEMETRYCPP)
  include_directories(${OPENTELEMETRY_CPP_INCLUDE_DIRS})

so we are preempting what I described above by pulling these dirs into the project on our own without CMake mediation! When this line (include_directories) is deleted, then the Clang build starts working.

This change sillences compile errors such as

```
In file included from /home/jdanek/repos/qpid/qpid-proton/cmake-build-debug-clang/vcpkg_installed/x64-linux/include/opentelemetry/nostd/./internal/absl/types/../utility/../base/internal/invoke.h:42:
/home/jdanek/repos/qpid/qpid-proton/cmake-build-debug-clang/vcpkg_installed/x64-linux/include/opentelemetry/nostd/./internal/absl/types/../utility/../base/internal/../../meta/type_traits.h:299:36: error: builtin __has_trivial_destructor is deprecated; use __is_trivially_destructible instead [-Werror,-Wdeprecated-builtins]
    : std::integral_constant<bool, __has_trivial_destructor(T) &&
```


astitcher (approved):
Lgtm 
This is clearly the correct cmake!
Looks like a mistake when we converted to using the new style of cmake

astitcher (changes requested):
Having approved - please change the commit comment to be much shorter! and just discuss the actual change rather than the process which you found it (however amusing!)

jiridanek:
@gemmellr @astitcher Do I need a jira for this?

gemmellr:
It sounds important enough to have one due to the fire and 'breaks build' aspect hehe. It doesnt need to be a very elaborate JIRA.

PR #395, opened 2023-04-19
PROTON-2715: remove unused for-cycle counter variable

jiridanek:
Some for loops just silently scream to be let out as while

```
[ 83%] Building CXX object cpp/examples/CMakeFiles/tracing_server.dir/tracing_server.cpp.o
/home/jdanek/repos/qpid/qpid-proton/cpp/examples/flow_control.cpp:65:18: error: variable 'i' set but not used [-Werror,-Wunused-but-set-variable]
        for (int i = sequence; available && s.credit() > 0; i++) {
                 ^
1 error generated.
```

### DIFF
--- a/cpp/examples/flow_control.cpp
+++ b/cpp/examples/flow_control.cpp
@@ -62,7 +62,7 @@ class flow_sender : public proton::messaging_handler {
     flow_sender() : available(0), sequence(0) {}
 
     void send_available_messages(proton::sender &s) {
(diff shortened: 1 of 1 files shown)


astitcher (approved):
Amazing the compiler stayed silent so long!

jiridanek:
> Amazing the compiler stayed silent so long!

It's the power of switching to clang! Maybe Red Hat should consider doing that ;DDD

PR #396, opened 2023-04-19
PROTON-xxxx: suppress deprecation warnings when using Jaeger in traci…

jiridanek:
…ng example

```
/home/jenkins/workspace/rh-qpid-proton-dist-el9-main/build/BUILD/qpid-proton-0.39.0/cpp/examples/tracing_client.cpp:157:50:
error: ‘JaegerExporter’ is deprecated
[-Werror=deprecated-declarations]
  157 |             new opentelemetry::exporter::jaeger::JaegerExporter(opts));
      |                                                  ^~~~~~~~~~~~~~
In file included from
```

### DIFF
--- a/cpp/examples/tracing_client.cpp
+++ b/cpp/examples/tracing_client.cpp
@@ -150,11 +150,16 @@ int main(int argc, char **argv) {
         // 2. Set the global trace provider.
         // 3. Call proton::initOpenTelemetryTracer().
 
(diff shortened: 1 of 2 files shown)


astitcher:
I think we should fix the example to use a non deprecated exporter rather than suppress the warning - the deprecation is going to happen pretty soon after the release (July 2023) so using the oltp exported instead makes more sense.

PR #397, opened 2023-04-19
PROTON-2714: add annotations for printf-like format strings and fix linux compiler warnings

jiridanek:


c/src/core/connection_driver.c
```diff
@@ -43,7 +43,7 @@ static pn_event_t *batch_next(pn_connection_driver_t *d) {
   /* Log the next event that will be processed */
   pn_event_t *next = pn_collector_next(d->collector);
   if (next && PN_SHOULD_LOG(&d->transport->logger, PN_SUBSYSTEM_EVENT, PN_LEVEL_DEBUG)) {
-    pni_logger_log_msg_inspect(&d->transport->logger, PN_SUBSYSTEM_EVENT, PN_LEVEL_DEBUG, next, "");
+    pni_logger_log_msg_inspect(&d->transport->logger, PN_SUBSYSTEM_EVENT, PN_LEVEL_DEBUG, next, "%s", "");
```
jiridanek, line 46: there is only this one line where the annotations lead to siliness needed not to generate a warning (empty formatting string on gcc generates warning)

c/src/ssl/openssl.c
```diff
@@ -1034,7 +1037,7 @@ static ssize_t process_input_ssl( pn_transport_t *transport, unsigned int layer,
     transport->present_layers |= LAYER_SSL;
   }
 
-  ssl_log( transport, PN_LEVEL_TRACE, "process_input_ssl( data size=%d )",available );
+  ssl_log( transport, PN_LEVEL_TRACE, "process_input_ssl( data size=%zu )",available );
```
jiridanek, line 1040: this may rather need the portable format string macros, not sure if necessary

c/src/core/logger.c
```diff
@@ -201,7 +201,7 @@ void pni_logger_log_raw(pn_logger_t *logger, pn_log_subsystem_t subsystem, pn_lo
   const char *start = &bytes.start[bytes.size-size];
   for (unsigned i = 0; i < size; i+=16) {
     pn_fixed_string_t out = pn_fixed_string(buf, sizeof(buf));
-    pn_fixed_string_addf(&out, "%04x/%04x: ", i, size);
+    pn_fixed_string_addf(&out, "%04x/%04zx: ", i, size);
```
jiridanek, line 204: this may rather need the portable format string macros, not sure if necessary

jiridanek:
It looks that msvc wants its annotation on both declaration and definition,

```
C:\Users\jdanek\Source\Repos\jiridanek\qpid-proton\c\src\core\object\string.c(197) : warning C28251: Inconsistent annotation for 'pn_string_format': this instance has no annotations. See c:\users\jdanek\source\repos\jiridanek\qpid-proton\c\src\core\object_private.h(187). 
```

jiridanek:
But it seems to be good enough to produce reports

image

astitcher:
I feel that we had these annotations in the original proton code base - I wonder where they went?
In any case I think this is a good direction - but after the 0.39 release!

jiridanek:
> I feel that we had these annotations in the original proton code base - I wonder where they went?

They are still there, but only applied to two of the functions. I'm rewriting those here to use the macros from Facebook.

> In any case I think this is a good direction - but after the 0.39 release!

Sure, I only decided to raise the PR now because otherwise I might forget to do it after the release. It found two broken format strings on Linux and quite a lot on Windows, but I guess this is not a release blocker. All about size_t printing; did not investigate, but it looks like somebody was switching to larger type and missed formatting strings while making the change?

c/include/proton/annotations.h
```diff
@@ -0,0 +1,93 @@
...
+#ifdef _MSC_VER
+
+    #include <stddef.h>
+
+    #ifdef _USE_ATTRIBUTES_FOR_SAL
+        #undef _USE_ATTRIBUTES_FOR_SAL
+    #endif
+
+    #define _USE_ATTRIBUTES_FOR_SAL 1
+    #include <sal.h>
```
jiridanek, line 58: no idea what this is there for, but facebook thought it worthy of including, so I copied it over

c/include/proton/annotations.h
```diff
@@ -0,0 +1,93 @@
...
+ *   Mutex& mutex, Args&&... args) { ... }
+ *
+ */
+
+// compiler specific attribute translation
+// msvc should come first, so if clang is in msvc mode it gets the right defines
+
+// warn format placeholders
+
+// NOTE: this will only do checking in msvc with versions that support /analyze
```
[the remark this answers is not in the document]
jiridanek: and the checks get lost in the noise very easily

jiridanek:
Sorry for creating the duplicate, I did not remember I already created a PR for this. It's actually good, since I already have it (mostly) reviewed, and the time has come preparing for next release, so it is a good time to do this.

PR #398, opened 2023-05-26
Fix build for jaeger to otlp tracing changes

DreamPearl:
Fix build for https://github.com/apache/qpid-proton/commit/1ad098c0703276f7a013a16b208f2af4fd86f993.

With the recent tracing changes with OT 1.9.1, the opentelemetry build succeed but the proton build failed at cmake configuration, here [build#763](https://github.com/apache/qpid-proton/actions/runs/5089577302/jobs/9147379969#step:11:60)

### DIFF
--- a/ci/otel.sh
+++ b/ci/otel.sh
@@ -38,6 +38,6 @@ mkdir build
 cd build
 
 cmake .. -DCMAKE_INSTALL_PREFIX=/usr -DBUILD_SHARED_LIBS=ON -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DBUILD_TESTING=OFF -DWITH_OTLP=ON -DWITH_OTLP_HTTP=ON -DWITH_OTLP_GRPC=OFF
-cmake --build . --target all --config RelWithDebInfo
-sudo cmake --install . --config RelWithDebInfo
+cmake --build . --target all
+sudo cmake --install .
 cd ../..


ci/otel.sh
```diff
@@ -38,6 +38,6 @@ mkdir build
 cd build
 
 cmake .. -DCMAKE_INSTALL_PREFIX=/usr -DBUILD_SHARED_LIBS=ON -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DBUILD_TESTING=OFF -DWITH_OTLP=ON -DWITH_OTLP_HTTP=ON -DWITH_OTLP_GRPC=OFF
```
jiridanek, line 40: The compilation prints deprecation warning

```
CMake Warning at CMakeLists.txt:165 (message):
  WITH_OTLP is deprecated and will be removed in future.  Please set either
  WITH_OTLP_GRPC or WITH_OTLP_HTTP, or even both.
```

https://github.com/apache/qpid-proton/actions/runs/5089577302/jobs/9147379969#step:10:205

jiridanek: @kpvdr IIRC you had trouble packaging the GRPC part, yes? Maybe it's an option to ignore the gRPC functionality, like it is done here? (Just thinking aloud, but it would make things simpler if it was the case.)

jiridanek (approved):
I'm honestly confused about why using `--config` causes these problems, but the PR does resolve them, there is one person on StackOverflow.com who used the same solution (https://stackoverflow.com/questions/74587365/imported-location-not-set-for-imported-target) and this is for the CI with no impact on Proton users, so even if it was not ideal solution, it is good enough for the CI.

ci/otel.sh
gemmellr: Its setting WITH_OTLP,  and also setting both WITH_OTLP_HTTP + WITH_OTLP_GRPC...the warnings seems like the latter two are replacement options, so can we just remove the first one now?

DreamPearl: Just to follow up here: Looks like we can't remove any of the flags for now, as it is leading to otel build failure.

PR #399, opened 2023-05-30
Remove 'continue' workaround for otel buiild

DreamPearl:
Fix deprecation warning.
```
CMake Warning at CMakeLists.txt:165 (message):
  WITH_OTLP is deprecated and will be removed in future.  Please set either
  WITH_OTLP_GRPC or WITH_OTLP_HTTP, or even both.
  ```
  
https://github.com/apache/qpid-proton/actions/runs/5121286125/jobs/9208853897#step:10:205

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,7 +64,7 @@ jobs:
     - name: OTel build/install
       if: runner.os == 'Linux'
       working-directory: ${{github.workspace}}
-      run: sudo sh ./ci/otel.sh || echo "Couldn't build Opentelemetry, continuing..."
+      run: sudo sh ./ci/otel.sh
       shell: bash
     - name: cmake configure
       working-directory: ${{env.BuildDir}}


gemmellr:
Unexpectedly, this seemingly causes an error during the otel build, which passed before but fails now.

It didnt fail the overall job due to Andrews prior 'commit before otel was released' workaround escaping the error from the build script. We should probably remove that so failures there cause the job to fail.


> /usr/bin/ld: cannot find -lopentelemetry_exporter_otlp_http: No such file or directory
collect2: error: ld returned 1 exit status
gmake[2]: *** [examples/otlp/CMakeFiles/example_otlp_http.dir/build.make:101: examples/otlp/example_otlp_http] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:892: examples/otlp/CMakeFiles/example_otlp_http.dir/all] Error 2
gmake: *** [Makefile:136: all] Error 2
Couldn't build Opentelemetry, continuing...

https://github.com/apache/qpid-proton/actions/runs/5121570439/jobs/9211136276?pr=399#step:10:292

DreamPearl:
Ah yes.
Then we need all three flags to make it work (I have also tried removing only -WITH_OTLP and keeping others, otel build is still failing). I guess, we have to ignore the warning for now and come back to it later.

And sure, I will remove the Andrew's earlier workaround so that we can see the otel build failure if it's there.

PR #401, opened 2023-08-01
NO-JIRA: [Python] TransactionalClientHandler doc correction

DreamPearl:


gemmellr:
Seems good (there is a test issue, but seems to be complaining about other areas, so maybe the parser itself was updated), but actually I think it deserves a JIRA given its changing the doc for a fundamental behaviour of using these bits, be good to note it was wrong and updated.

gemmellr:
The test issue was unrelated and fixed in 2822b9f18af1ac389c6d1a347d267764ae15a192.

PR #407, opened 2023-10-14
PROTON-2772: add format arg annotations for schannel functions

jiridanek:
What's missing is a compilation failure when any warnings are emitted.

MSVC needs /analyze to report the problems, there is actually lots of diagnostics printed then, and the mismatched format args are not causing compilation to fail so the warnings are easy to miss.


jiridanek:
* Superseded by https://github.com/apache/qpid-proton/pull/409

PR #408, opened 2023-10-14
PROTON-2773: add `PN_FALLTHROUGH` macro ready for c++17/c21

jiridanek:


jiridanek:
The original motivation for this is to be able to have composite CMake build with skupper-router. With skupper-router adding proton as cmake subproject. To do this, both projects need to compile with the same flags, and skupper-router defines `-Wimplicit-fallthrough`. Therefore, proton should be clean for this warning too.

jiridanek:
> See also https://lore.kernel.org/lkml/a588204afbfe4c8dd56d0cb00d8e6e14dc561a62.camel@perches.com/
>
> where Joe Perches who maintains the Linux kernel's patch validator (scripts/checkpatch.pl) wrote a tool to convert all of the Linux kernel's comments into attribute((fallthrough)), but also provided statistics of every possible spelling of fallthrough seen in the kernel, which is useful data for those looking to fix every possible case.

* https://github.com/llvm/llvm-project/issues/42810#issuecomment-981019196

c/include/proton/annotations.h
```diff
@@ -90,4 +90,24 @@
     #endif
 #endif
 
+// fallthrough
+#if defined __cplusplus && defined __has_cpp_attribute
+    #if __has_cpp_attribute(fallthrough) && __cplusplus >= __has_cpp_attribute(fallthrough)
+        #define PN_FALLTHROUGH [[fallthrough]]
+    #endif
+#endif
+#if !defined PN_FALLTHROUGH && defined __STDC_VERSION__ && defined __has_c_attribute
```
astitcher, line 99: Could be `#elif`

c/include/proton/annotations.h
```diff
@@ -90,4 +90,24 @@
...
+    #if __has_c_attribute(fallthrough) && __STDC_VERSION__ >= __has_c_attribute(fallthrough)
+        #define PN_FALLTHROUGH [[fallthrough]]
+    #endif
+#endif
+#if !defined PN_FALLTHROUGH && defined __has_attribute
```
astitcher, line 104: again could be `#elif`

c/include/proton/annotations.h
```diff
@@ -90,4 +90,24 @@
...
+    #if __has_attribute(__fallthrough__)
+        #define PN_FALLTHROUGH __attribute__((__fallthrough__))
+    #endif
+#endif
+#if !defined PN_FALLTHROUGH
```
astitcher, line 109: this could be `#else`

c/include/proton/annotations.h
```diff
@@ -90,4 +90,24 @@
...
+    #define PN_FALLTHROUGH (void)0
```
astitcher, line 110: Although does this actually tell a linter that there is an expected fallthrough? - That is what the `//Fallthru` comments are for, but I don't think you can preprocess into comments.

astitcher (commented):
Basically lgtm.
My comments are basically stylistic, just my preference for if...else if.. chains in this case.

c/include/proton/annotations.h
jiridanek: True, this depends on all relevant linters knowing one of the other ways to annotate the fallthrough.

The default is there for the case of a compiler that does not know how to warn about missed fallthrough but knows to warn about extraneous semicolon. Not that I'd expect to ever encounter something like this.

c/include/proton/annotations.h
jiridanek: For this c/c++ check, yes, this is exclusive, but I can't replace all the #endif #if with #elif here.

I changed this one line.

c/include/proton/annotations.h
jiridanek: no, because we might be on older c that has a __has_c_attribute but not the [[fallthrough]];, so we would go into

```
#if !defined PN_FALLTHROUGH && defined __STDC_VERSION__ && defined __has_c_attribute
````

but not define PN_FALLTHROUGH there

PR #409, opened 2023-10-14
PROTON-2772: make wrong printf format strings a MSVC compile error

jiridanek:


jiridanek:
This PR currently only reveals the issues, it does not try to fix them. There is few tens of warnings generated, for C6328 and C6340 warnings.

https://github.com/jiridanek/qpid-proton/actions/runs/6518361158/job/17703711656#step:12:810

The output of MSVC becomes quite messy when /analyze is enabled. Any suggestions what should be done about it would be most welcome. I _could_ just disable them one by one... and leave them for consideration at some time in the future.

jiridanek:
My plan here is to boot up Windows, add all the new warnings on MSVC to the CMakeLists.txt as suppressed, and merge that. I will add comments what these numbers mean. I'll consider putting this into a separate cmake file if the list of suppressions becomes very long.

I imagine this might be hard to maintain if we ever wanted to do warnings-as-errors on msvc, or if msvc then fails when compiling on an older version which does not know some particular warning flag (gcc does this). I will not try to do warnings-as-errors on MSVC to be sure.

jiridanek:
This has been here for a long time with no objections raised. I'm going to squash the commits and merge this first part (annotations and CMakeLists.txt flags).

I'll create a new PR for actually fixing the warnings and I'll leave it here for review next.

PR #410, opened 2023-10-14
PROTON-2774: replace `CMAKE_SOURCE_DIR` variable with `PROJECT_SOURCE_DIR`

jiridanek:
This change allows the CMake build file work as an embedded subproject in an composite CMake build.


astitcher (approved):
Assuming this builds and tests ok. lgtm.

jiridanek:
Even Appveyor passed, it had some infra problems last week and was constantly red then. Merging!

PR #411, opened 2023-10-16
NO-JIRA: Sort input file list

bmwiedemann:
Sort input file list so that `_cproton.cpython-310-x86_64-linux-gnu.so` builds in a reproducible way in spite of non-deterministic filesystem readdir order

See https://reproducible-builds.org/ for why this is good.

This patch was done while working on reproducible builds for openSUSE.

### DIFF
--- a/python/ext_build.py
+++ b/python/ext_build.py
@@ -40,7 +40,9 @@ proton_c_include = os.path.join(proton_base, 'include')
 sources = []
 extra = []
 libraries = []
-for root, _, files in os.walk(proton_core_src):
+for root, dirs, files in os.walk(proton_core_src):
+    dirs.sort() # needed for os.walk to process directories in deterministic order
+    files.sort()
     for file_ in files:
         if file_.endswith(('.c', '.cpp')):
             sources.append(os.path.join(root, file_))


python/ext_build.py
```diff
@@ -40,7 +40,9 @@
 sources = []
 extra = []
 libraries = []
-for root, _, files in os.walk(proton_core_src):
+for root, dirs, files in os.walk(proton_core_src):
+    dirs.sort()
```
astitcher, line 44: Since this code doesn't use dirs I can't see why you added it into the for loop and sort it.
Can you fix this PR to just be the single line that sorts files?
Afaict this is sufficient to fix your issue.

astitcher (changes requested):
Having a reproducible build is a very good idea. There seems to be some unnecessary changes here though. Can you fix that please.

python/ext_build.py
bmwiedemann: If there were no dirs, you wouldn't need to use `os.walk` in the first place but could use `sorted(os.listdir(dir))` or `sorted(glob.glob('path/to/*.cpp'))`. But there is a [`object` subdir](https://github.com/apache/qpid-proton/tree/main/c/src/core/object) involved.
When you omit this part, it can still traverse dirs in random order, so I added it for future-proofness in case a 2nd dir gets added there.

bmwiedemann: For a test, you can try with
```bash
#!/bin/sh
for i in $(seq 5) ; do
 mkdir -p t/$i
 for j in $(seq 5) ; do
  touch t/$i/$i$j
 done
done
```
and then
```python
#!/usr/bin/python3
import os;

for root, dirs, files in os.walk("t"):
    files.sort();
    #dirs.sort();
    print(files);
```
which produced here on a ext4 filesystem:
```
[]
['11', '12', '13', '14', '15']
['31', '32', '33', '34', '35']
['21', '22', '23', '24', '25']
['51', '52', '53', '54', '55']
['41', '42', '43', '44', '45']
```

So when you wrote "this code doesn't use dirs" you forgot the `os.walk` call itself that does use `dirs`.

astitcher: No, I literally mean that the variable `dirs` is never used and so sorting it is pointless. Indeed in the code you changed the variable `dirs` is not even there instead there is a placeholder `_`.
To expand a little bit -  the `os.walk` you mention generates a list of tuples` (root, dirs, files)` but the code in the loop never uses the `dirs` part of that tuple which is why in the original code there is only a placeholder `_`.
I this clearer now?

astitcher: I'd be very happy for you to add a comment stating that sorting the `files` list is necessary for reproducible builds. That would strongly suggest that if this code ever needs to use the `dirs` list in the future it should be sorted too.

bmwiedemann: In my test, I found that `_.sort()` also does the trick :-)
So that variable is used, even if you do not give it a proper name.

astitcher: Huh, Are you saying that the output directory list is the same as the internal directory list of `os.walk` so that modifying it changes the order of the results from the `for` loop? 
Wow if so then leave the original code, but there has to be a comment to explain why the code is there as it looks like `dirs` is never used so the code to sort it is dead code.

astitcher: Hmm, so I see that this is the documented behaviour for `os.walk` in the python library docs, but it seems very unintuitive when reading the code :cry:

bmwiedemann: I added a comment to explain it.

astitcher (approved):

PR #415, opened 2023-12-22
PROTON-2786: Don't reuse possibly incorrect `CMAKE_CXX_COMPILER` value determined by `check_language(CXX)`

jiridanek:


jiridanek:
Inspiration for the fix

* https://github.com/root-project/root/blob/0199f080f16d6168660e5dbcc15362b1b20979e0/cmake/modules/CheckCompiler.cmake#L37-L49

PR #419, opened 2024-01-07
PROTON-2772: add printf format arg annotations to satisfy MSVC

jiridanek:


c/src/core/engine.c
```diff
@@ -1645,9 +1645,9 @@ void pn_delivery_dump(pn_delivery_t *d)
   printf("{tag=%s, local.type=%" PRIu64 ", remote.type=%" PRIu64 ", local.settled=%u, "
          "remote.settled=%u, updated=%u, current=%u, writable=%u, readable=%u, "
          "work=%u}",
-         tag, d->local.type, d->remote.type, d->local.settled,
-         d->remote.settled, d->updated, pn_delivery_current(d),
-         pn_delivery_writable(d), pn_delivery_readable(d), d->work);
+         tag, d->local.type, d->remote.type, (unsigned)d->local.settled,
+         (unsigned)d->remote.settled, (unsigned)d->updated, (unsigned)pn_delivery_current(d),
+         (unsigned)pn_delivery_writable(d), (unsigned)pn_delivery_readable(d), (unsigned)d->work);
```
jiridanek, line 1650: MSVC would be happy to print `bool`s with `%i`, but `%u` it does not like. Maybe just go with `%i` or `%d` instead?

c/src/core/encoder.c
```diff
@@ -78,7 +78,7 @@ static uint8_t pn_type2code(pn_encoder_t *encoder, pn_type_t type)
   case PN_MAP: return PNE_MAP32;
   case PN_DESCRIBED: return PNE_DESCRIPTOR;
   default:
-    return pn_error_format(pni_encoder_error(encoder), PN_ERR, "not a value type: %u\n", type);
+    return pn_error_format(pni_encoder_error(encoder), PN_ERR, "not a value type: %u\n", (unsigned int)type);
```
jiridanek, line 81: This is about printing an enum, the compiler is unhappy with `%u`. Not really sure what's the nicest way to resolve this. In C23, it will be possible to specify a backing type for enum.

Nowadays, if I read correctly, iti is either a char, int, or unsigned int. Wouldn't maybe treating it as an int work better here? So %d and (int) cast?

c/src/core/engine.c
jiridanek: I'm changing this to %d

c/src/core/encoder.c
jiridanek: Let;s go with %d as well

PR #422, opened 2024-03-08
PROTON-2792: [C++] check that scheduled tasks are active under lock

astitcher:
Previously we checked whether the tasks were active without locking which was bad.

This is is an alternative fix (which has the correct semantics) to #421.

I think it probably still needs a test for the fixed behaviour and the correct semantics before merging. This would be new tests in ```container_test.cpp```


zlomek:
Yes, this fixes the issue while preserving the semantics, thank you.

cliffjansen:
+1
I will merge this.

PR #423, opened 2024-04-12
PROTON-2322/PROTON-2813: [Python] Finish PEP8 sanitization

astitcher:
This is a broad but shallow change, that changes a lot of files in fairly minor ways.


astitcher:
I want to be sure this didn't break any build, and especially any version of python from 3.9 onwards.
But assuming is doesn't then I'll commit it to main.

python/proton/_events.py
```diff
@@ -57,7 +57,6 @@ def more(self) -> bool:
         return pn_collector_more(self._impl)
 
     def pop(self) -> None:
-        ev = self.peek()
```
jiridanek, line 60: This is syntax error, method without a body

python/proton/_handlers.py
```diff
@@ -1305,15 +1305,15 @@ def update(transport: Transport, selectable: Selectable, now: float) -> None:
         try:
             capacity = transport.capacity()
             selectable.reading = capacity > 0
-        except:
+        except ProtonException:
```
jiridanek, line 1308: Note to self, check this again. This exception change 
thing is risky.

python/proton/_handlers.py
```diff
@@ -1305,15 +1305,15 @@ def update(transport: Transport, selectable: Selectable, now: float) -> None:
...
             if transport.closed:
                 selectable.terminate()
                 selectable._transport = None
                 transport._selectable = None
         try:
             pending = transport.pending()
             selectable.writing = pending > 0
-        except:
+        except ProtonException:
```
jiridanek, line 1316: Ditto

python/setup.cfg
```diff
@@ -28,26 +28,4 @@ install_requires = cffi>=1.0.0
 opentracing = opentracing; jaeger_client
 
 [flake8]
-# TODO(PROTON-2322) decrease the limit
-max-line-length = 150
```
jiridanek, line 32: I think I would've kept larger limit. Maybe not 150, but something like 120 sounds ok nowadays.

jiridanek: Nvm, that's done below

python/setup.cfg
```diff
@@ -28,26 +28,4 @@ install_requires = cffi>=1.0.0
...
-    # imported but unused
-    F401,
-    # ambiguous variable name 'l'
-    E741,
-    # local variable 'int_nodes' is assigned to but never used
-    F841,
-    # TODO(PROTON-2322) decide which of these two warnings we want disabled
-    #  _transport.py:908:21: W503 line break before binary operator
-    #  _transport.py:907:56: W504 line break after binary operator
-    W504,
```
jiridanek, line 47: One of the w503, w504 should remain disabled, imo, otherwise it's not possible to have line breaks in expressions, which can be useful. But if it works out for you without this, then I guess it's fine.

python/proton/_events.py
jiridanek: ah, never mind, there is one more statement below; so this was removed because `ev` was unused and `self.peek()` only returns value without doing any side-effects

python/proton/_handlers.py
jiridanek: looks good

python/proton/_handlers.py
jiridanek: lgtm

jiridanek:
Tried a Python compatibility check in IntelliJ, and got

image

https://github.com/apache/qpid-proton/blob/df1e9b684c364338faa88abf22921a05804e3b13/python/tests/proton_tests/main.py#L29

python 3.13 will be removing the `cgi` module from the standard library, so it needs to be removed from proton examples eventually

https://docs.python.org/3.13/whatsnew/3.13.html#summary-release-highlights

python/setup.cfg
```diff
@@ -28,26 +28,4 @@ install_requires = cffi>=1.0.0
...
-max-line-length = 150
-
-# TODO(PROTON-2322) re-enable all these warnings
-ignore =
-    # do not use bare 'except'
-    E722,
-    # imported but unused
-    F401,
-    # ambiguous variable name 'l'
-    E741,
```
jiridanek, line 41: don't like this warning, would've kept it disabled

jiridanek (approved):

python/setup.cfg
jiridanek: https://peps.python.org/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

astitcher: I rewrote the code to avoid using more than one line (and I think made it a little easier to understand)

And in general it's avoidable if you're checking before committing the change.

python/setup.cfg
astitcher: I agree that I wouldn't put it in a coding standard myself - but forbidding the letter 'l' to avoid confusion eith the number '1' has some sense IMO and it is in PEP8.

Again it's easy enough to avoid if you're checking before commit.

astitcher:
> python 3.13 will be removing the `cgi` module from the standard library, so it needs to be removed from proton examples eventually
> 
> https://docs.python.org/3.13/whatsnew/3.13.html#summary-release-highlights

That's interesting - I'll look at that I wonder what we use it for.

PR #429, opened 2024-07-10
NO-JIRA: c/cpp: fix examples installation

a3f:
The c/cpp examples so far installed the source files into ${PROTON_SHARE}. This is not what most users would expect, especially if we are cross-compiling and the target lacks a toolchain.

Fix this by installing the binary artifacts instead. For them to work out-of-the-box, we additionally copy the ssl-certs to the expected location as well.

### DIFF
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -715,7 +715,9 @@ if (BUILD_TOOLS)
   add_subdirectory(tools)
 endif (BUILD_TOOLS)
 
-install (DIRECTORY examples/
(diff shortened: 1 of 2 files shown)


astitcher:
This change misses the entire point that the examples are _source_ examples to show how the API user how to sue the API.
There may well be an issue that the examples should not be installed in the target for a cross compilation. But the example binaries are not intended to be installed as they aren't general utilities.
So there probably is an underlying issue here that cross compilation is not considered in the build - I would be grateful if you'd raise a JIRA to address this problem with as much detail as you understand.

PR #430, opened 2024-07-10
PROTON-2594: Add OpenSSL PKCS#11 PROVIDER support to enable HSM use 

a3f:
Hardware Security Modules (HSMs) are physical computing devices that
safeguard secrets and allow performing cryptographic operations like
encryption and signing without necessarily divulging the private key
material.

PKCS#11 is a platform-independent API for cryptographic tokens like
HSMs. It defines a scheme for pkcs11: URIs that describe objects on the
token as well as an API to interact with them.

This commit adds support for transparent use of PKCS#11 URIs: Whenever a
certificate or private key path is prefixed with pkcs11: it will be
interpreted as PKCS#11 URI instead of a file path and OpenSSL will do
all necessary communication with the HSM behind the scenes.

For programs like proton that use OpenSSL, this could have been
realized in two ways:

  - OpenSSL ENGINE:   Introduced in OpenSSL 0.9.6 and deprecated in 3.0
  - OpenSSL PROVIDER: Introduced in OpenSSL 3.0 to replace ENGINE

While both are supported in recent OpenSSL versions, it's more future
proof to use the PROVIDER API, even if this comes at the cost of having
to add an #ifdef to the code.

This has been tested on Linux/x86_64 with softhsm and Linux/arm32 with
OP-TEE, both with v0.5 of https://github.com/latchset/pkcs11-provider.

As everything is loaded dynamically, we do not link against any
PKCS#11-related shared libraries. It's expected that PKCS#11 provider
will already be loaded via OPENSSL_CONF if it's to be used.

### DIFF
--- a/c/examples/broker.c
+++ b/c/examples/broker.c
@@ -35,7 +35,13 @@
 /* The ssl-certs subdir must be in the current directory for an ssl-enabled broker */
 #define SSL_PW "tserverpw"
 /* Windows vs. OpenSSL certificates */
-#if defined(_WIN32)
+#if defined(PKCS11_URI)
+#  define SSL_FILE(NAME) PKCS11_URI ";object=" NAME
+#  define PRIVATEKEY(NAME) SSL_FILE(NAME ";type=private")
+#  define CERTIFICATE(NAME) SSL_FILE(NAME ";type=cert")
+#  define SET_CREDENTIALS(DOMAIN, NAME)                                 \
+  pn_ssl_domain_set_credentials(DOMAIN, CERTIFICATE(NAME), PRIVATEKEY(NAME), SSL_PW)
+#elif defined(_WIN32)
 #  define SSL_FILE(NAME) "ssl-certs/" NAME
 #  define PRIVATEKEY(NAME) SSL_FILE(NAME "-full.p12")
 #  define CERTIFICATE(NAME) SSL_FILE(NAME "-certificate.p12")
--- a/c/examples/broker.c
+++ b/c/examples/broker.c
@@ -33,17 +33,20 @@
 #include <string.h>
 
 /* The ssl-certs subdir must be in the current directory for an ssl-enabled broker */
-#define SSL_FILE(NAME) "ssl-certs/" NAME
 #define SSL_PW "tserverpw"
 /* Windows vs. OpenSSL certificates */
 #if defined(_WIN32)
+#  define SSL_FILE(NAME) "ssl-certs/" NAME
+#  define PRIVATEKEY(NAME) SSL_FILE(NAME "-full.p12")
 #  define CERTIFICATE(NAME) SSL_FILE(NAME "-certificate.p12")
 #  define SET_CREDENTIALS(DOMAIN, NAME)                                 \
-  pn_ssl_domain_set_credentials(DOMAIN, SSL_FILE(NAME "-full.p12"), "", SSL_PW)
+  pn_ssl_domain_set_credentials(DOMAIN, PRIVATEKEY(NAME), "", SSL_PW)
 #else
+#  define SSL_FILE(NAME) "ssl-certs/" NAME
+#  define PRIVATEKEY(NAME) SSL_FILE(NAME "-private-key.pem")
 #  define CERTIFICATE(NAME) SSL_FILE(NAME "-certificate.pem")
 #  define SET_CREDENTIALS(DOMAIN, NAME)                                 \
-  pn_ssl_domain_set_credentials(DOMAIN, CERTIFICATE(NAME), SSL_FILE(NAME "-private-key.pem"), SSL_PW)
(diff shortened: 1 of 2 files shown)


astitcher:
This looks good to me - is there any way of testing this functionality in the CI job? (Rather than just testing that nothing regressed!)

a3f:
> is there any way of testing this functionality in the CI job? (Rather than just testing that nothing regressed!)

How about we apply [this patch](https://github.com/a3f/pkcs11-provider-qpid-proton-bug-reproduction/blob/master/0001-HACK-PROTON-2594-cpp-connect_config_test-adapt-for-t.patch) and then run the test twice once with PEM files and once with PKCS#11?

a3f:
I added a `pkc11_test` that receives `pkcs11:` URIs for both server and client certificates and keys via environment variables. A script executed by CI adds the PEM files available in tree to a SoftHSM and exports said environment variables.

The test is skipped without being marked as failure whenever the environment variables are missing as not break other users running the test suite, but lacking the prerequisites.

This works fine locally, but unfortunately fails in CI and I am not sure why.

@astitcher How would you go about debugging issues that only fail in CI? I tried https://github.com/nektos/act, but I don't have a docker image that looks sufficiently enough like the Github runner VM images, so it's able to execute the tests as-is.

astitcher (commented):
In general  the tests in the cpp subdirectory are used to test the C++ API binding rather than the underlying proton-c functionality. This is because building the C++ binding is optional and so putting underlying proton-c testing here risks not testing fully.
It is true that the unit testing of proton-c does actually also use a C++ test harness but it is in the c subdirectory. If at all possible I'd really like to maintain this distinction for these tests too.
There are ssl tests there that I think you copy in the same way that you've copied a previous test for this test. It probably won't be as straightforward as this though - really sorry about that - we should probably add some more testing infrastructure to the C side to make it simpler.

astitcher:
> I added a `pkc11_test` that receives `pkcs11:` URIs for both server and client certificates and keys via environment variables. A script executed by CI adds the PEM files available in tree to a SoftHSM and exports said environment variables.
> 
> The test is skipped without being marked as failure whenever the environment variables are missing as not break other users running the test suite, but lacking the prerequisites.
> 
> This works fine locally, but unfortunately fails in CI and I am not sure why.
> 
> @astitcher How would you go about debugging issues that only fail in CI? I tried https://github.com/nektos/act, but I don't have a docker image that looks sufficiently enough like the Github runner VM images, so it's able to execute the tests as-is.

I'm not aware of anything specific. I guess adding extra debugging output until you can figure out enough to make it fail in your local environment too. I know that's not very helpful - sorry :-(

astitcher:
Hmm, have you made sure that the CI image actually has the relevant openssl modules for the softhsm? Maybe you need to apt install something to make the test work in the CI envioronment.

a3f:
> Hmm, have you made sure that the CI image actually has the relevant openssl modules for the softhsm? Maybe you need to apt install something to make the test work in the CI envioronment.

Yes, that's what the `apt install pkcs11-provider` was for. I managed to reproduce in a ubuntu-24.04 container. The pkcs11-provider shipped there was the culprit. I now build the pkcs11-provider manually and the tests pass finally: https://github.com/a3f/qpid-proton/actions/runs/11591599162.

> If at all possible I'd really like to maintain this distinction for these tests too.

I can understand that, but most users will likely not run the PKCS#11 tests anyway, because there are more prerequisites than the C++ compiler (Linux, softhsm2, pkcs11-provider, OpenSSL > 3.0).
Unfortunately, this side quest with debugging the CI has more than depleted the time I have alotted for this, so I won't be able to rewrite the test myself anytime soon.

astitcher:
Thanks for the work of adding the test.
Can you squash together the last 4 commits into 1 commit leaving just 3 commits overall please. I don't think those commits standalone really - they are just part of adding the test.

astitcher:
I've asked @cliffjansen to take a look at this. He has more experience with the openssl code.

a3f:
> This PR has some problems for which I have some proposed fixes. The main ones being compilation failures on older versions of OpenSSL and a memory leak.

Thanks and sorry for missing those. I now squashed your fixes directly into the commits.

> If you can confirm my changes do not break anything in your environment with a real hardware security module, I will check this in and add a documentation update. My proposed changes are here:

Your changes look fine to me. Feel free to push any commits you want to add on top of this PR.

> I will revert the github action back to ubuntu-latest once that advances to ubuntu-24.04 in a few weeks.

Sounds good.

> It is unfortunate the software emulation bits of PKCS11 support are sufficiently buggy/immature that you had to take extraordinary steps to build a custom package and configuration just to get the test to run. Hopefully the tests can be altered to use regular distro packages in time.

This is mostly because of the OpenSSL switch from ENGINE to PROVIDER. The PKCS#11 Engine is shipped universally, but is now deprecated. The provider is getting there.

Thanks again,

cliffjansen:
Now on main.  Rebased to af0124ef

PR #433, opened 2024-09-05
PROTON-2848: [Python] Add py.typed marker to package for PEP 561

PatrickTaibel:
As far as I can tell, the python binding is properly typed. In order to allow using tools like mypy, it is required to have a py.typed file in the package (see PEP 561).


astitcher:
I'm not as confident as you that all the type annotations are actually correct, but they are intended to be! And we should fix the ones that aren't so I will merge this change and we can deal with any fallout.

PR #437, opened 2024-11-20
PROTON-1442: [Cpp] Support for local transactions

DreamPearl:
[PROTON-1442](https://issues.apache.org/jira/projects/PROTON/issues/PROTON-1442)

AMQP Transaction Sequence:

1. Declare transaction:

- Client establishes link to Broker (Transaction resource) to target with transaction coordinator type (usual types are sender/receiver) (ATTACH frame)

- Client (Transaction Controller) sends a special message to that link to create transaction (TRANSFER frame)

- Broker returns a disposition with the transaction id (DISPOSITION frame)


cpp/include/proton/transaction.hpp
```diff
@@ -0,0 +1,124 @@
...
+    // PN_CPP_EXTERN transaction(transaction &o);
+    PN_CPP_EXTERN transaction();
+    PN_CPP_EXTERN ~transaction();
+    PN_CPP_EXTERN bool is_empty();
+    PN_CPP_EXTERN void commit();
+    PN_CPP_EXTERN void abort();
+    PN_CPP_EXTERN void declare();
+    PN_CPP_EXTERN void handle_outcome(proton::tracker);
+    PN_CPP_EXTERN proton::tracker send(proton::sender s, proton::message msg);
+    PN_CPP_EXTERN void accept(delivery &t);
```
astitcher, line 95: I think that you should be able to reject, release & modify in  transaction too

cpp/include/proton/transaction.hpp
```diff
@@ -0,0 +1,124 @@
...
+/// @copybrief proton::transaction
+
+namespace proton {
+
+class transaction_handler;
+
+// TODO: This should not be accessible to users.
+class transaction_impl {
+  public:
+    proton::sender txn_ctrl;
```
astitcher, line 43: I think this is one per connection not per transaction. The transaction needs a way to find its conection (or session if it's per session) anyway so you should always be able to find the transaction controller

cpp/include/proton/transaction.hpp
```diff
@@ -0,0 +1,124 @@
...
+    proton::transaction_handler *handler = nullptr;
+    proton::binary id;
+    proton::tracker _declare;
+    proton::tracker _discharge;
```
astitcher, line 47: These must be exclusive as you cannot be delivering a declare and a discharge at the same time so I think you only need to keep track of a single outgoing delivery.

cpp/include/proton/transaction.hpp
```diff
@@ -0,0 +1,124 @@
...
+    PN_CPP_EXTERN bool is_empty();
+    PN_CPP_EXTERN void commit();
+    PN_CPP_EXTERN void abort();
+    PN_CPP_EXTERN void declare();
+    PN_CPP_EXTERN void handle_outcome(proton::tracker);
+    PN_CPP_EXTERN proton::tracker send(proton::sender s, proton::message msg);
+    PN_CPP_EXTERN void accept(delivery &t);
+
+  friend class transaction_impl;
+  friend class container::impl;
```
astitcher, line 98: Pretty sure this can't be needed as only the implementation could need to use other implementatino classes.

cpp/examples/tx_recv.cpp
```diff
@@ -0,0 +1,130 @@
...
+        std::cout << "    [on_container_start] declare_txn started..." << std::endl;
+        c.declare_transaction(connection, *this);
+        std::cout << "    [on_container_start] completed!!" << &transaction
+                  << std::endl;
+    }
+
+    void on_transaction_declare_failed(proton::transaction) {}
+    void on_transaction_commit_failed(proton::transaction) {
+        std::cout << "Transaction Commit Failed" << std::endl;
+        connection.close();
```
astitcher, line 68: we need to bwe able to get from the transaction to the session/connection it's associated with so we don't need to store the connection away we should just get it from the transaction in the handler.

cpp/examples/tx_recv.cpp
```diff
@@ -0,0 +1,130 @@
...
+    proton::receiver receiver; 
+    std::string url;
+    int expected;
+    int batch_size;
+    int current_batch = 0;
+    int committed = 0;
+
+    proton::container *container;
+    proton::transaction transaction;
+    proton::connection connection;
```
astitcher, line 50: We shouldn't need to hold the connection as we should always be able to find the connection in a handler. This may be true for the transaction too - either we restrict the API to be one transaction per session or one transaction per connection. I think per session is porbably better as it matches the cms/jms api.

astitcher (changes requested):
I've taken a fairly thorough look. If you have any questions about these review items we should discuss further.

cpp/include/proton/container.hpp
```diff
@@ -312,6 +312,7 @@ class PN_CPP_CLASS_EXTERN container {
     /// Cancel task for the given work_handle.
     PN_CPP_EXTERN void cancel(work_handle);
 
+    PN_CPP_EXTERN transaction declare_transaction(proton::connection conn, proton::transaction_handler &handler, bool settle_before_discharge = false);
```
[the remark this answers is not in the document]
DreamPearl: Moved the entire  declare_transaction method to proton::session class.

cpp/include/proton/tracker.hpp
```diff
@@ -25,6 +25,7 @@
 #include "./binary.hpp"
 #include "./internal/export.hpp"
 #include "./transfer.hpp"
+#include "./messaging_handler.hpp"
```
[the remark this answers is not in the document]
DreamPearl: Fixed.

cpp/include/proton/transfer.hpp
```diff
@@ -33,9 +33,26 @@
...
+    /// Create an empty disposition.
+    disposition() : internal::object<pn_disposition_t>(0) {}
+
+    proton::value data() const;
+
+    /// @cond INTERNAL
+    friend class internal::factory<disposition>;
+    /// @endcond
+};
+
```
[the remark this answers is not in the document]
DreamPearl: Removed as we don't need it rn.

cpp/src/container.cpp
```diff
@@ -28,6 +28,7 @@
 #include "proton/uuid.hpp"
 
 #include "proactor_container_impl.hpp"
+#include <vector>
```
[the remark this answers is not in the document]
DreamPearl: Removed!

cpp/src/container.cpp
```diff
@@ -45,6 +46,10 @@ returned<connection> container::connect(const std::string &url) {
     return connect(url, connection_options());
 }
 
+transaction container::declare_transaction(proton::connection conn, proton::transaction_handler &handler, bool settle_before_discharge) {
+    return impl_->declare_transaction(conn, handler, settle_before_discharge);
+}
+
```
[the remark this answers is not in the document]
DreamPearl: Fixed!

cpp/src/handler.cpp
```diff
@@ -34,6 +34,7 @@
 
 #include "proton/connection.h"
 #include "proton/session.h"
+#include "proton/tracker.hpp"
```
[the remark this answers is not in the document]
DreamPearl: Removed.

cpp/include/proton/transaction.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+    static transaction mk_transaction_impl(sender &s, transaction_handler &h,
+                                           bool f);
+    PN_CPP_EXTERN transaction(transaction_impl *impl);
+    transaction_impl *_impl;
+
+  public:
+    // TODO:
+    // PN_CPP_EXTERN transaction(transaction &o);
+    PN_CPP_EXTERN transaction();
```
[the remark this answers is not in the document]
astitcher: Having said that I think a better overall API might be not to expose the  transaction class to the API user at all and to have all the methods either on session or session::commit(), session::abort(),  session::is_transactioned(). I think the way to do this without introducing a new transacted_session class would be to add a new session_option - something like bool transactioned() to make a transactioned session. Then in the transactioned case only call on_session_open when the transaction has been successfully declared. This would be instead of the transaction_handler::transaction_declared() callback.
Now any sender or receiver that is created from this session transparently will send or acknowledge in the transaction.

cpp/include/proton/transaction.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+    /// Called when a local transaction declare fails.
+    PN_CPP_EXTERN virtual void on_transaction_declare_failed(transaction);
+
+    /// Called when the commit of a local transaction fails.
+    PN_CPP_EXTERN virtual void on_transaction_commit_failed(transaction);
+};
+
+} // namespace proton
+
```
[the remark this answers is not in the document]
astitcher: So if the API makes transactions hidden in a session then these callbacks either go away or instead become session callbacks: on_session_transaction_committed(session&), etc. I think transaction_declared() goes away entirely and it's purpose is now another use for on_session_open(session&). on_.._declare_failed should be handled by the on_session_error(session&).

cpp/include/proton/transfer.hpp
```diff
@@ -77,20 +94,28 @@ class transfer : public internal::object<pn_delivery_t> {
     /// Return true if the transfer has been settled.
     PN_CPP_EXTERN bool settled() const;
 
+   // Set transaction
+    PN_CPP_EXTERN void transaction(transaction t);
+
+    PN_CPP_EXTERN class transaction transaction() const;
```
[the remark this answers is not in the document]
astitcher: I think these are not use visible. They are only used internally in the implementation - especially if the transaction is hidden inside the session.

astitcher (changes requested):
I've not reviewed this comprehensively, but I think I'm now very much leaning towards not making the transaction class visible to the API user at all and having the transaction methods on the session. This is essentially the API in JMS and CMS.

cpp/examples/tx_send.cpp
```diff
@@ -0,0 +1,175 @@
...
+        send(sender);
+    }
+
+    void on_sendable(proton::sender &s) override {
+        std::cout << "    [OnSendable] transaction: " << &transaction
+                  << std::endl;
+        send(s);
+    }
+
+    void send(proton::sender &s) {
```
pematous, line 91: sender is defined as tx_recv class attribute, so I believe we don't need to pass the sender to send method.

cpp/examples/tx_recv.cpp
```diff
@@ -0,0 +1,129 @@
...
+        else {
+            session.declare_transaction(*this);
+        }
+    }
+
+};
+
+int main(int argc, char **argv) {
+    std::string address("127.0.0.1:5672/examples");
+    int message_count = 9;
```
pematous, line 107: I would expect the defaults to be the same for tx_send and tx_recv (currently tx_send defaults to 6 messages while tx_recv to 9)

cpp/examples/tx_send.cpp
```diff
@@ -0,0 +1,175 @@
...
+            msg.id(unique_id++);
+            msg.body(m);
+            std::cout << "##### [example] transaction send msg: " << msg
+                      << std::endl;
+            transaction.send(sender, msg);
+            current_batch += 1;
+            if(current_batch == batch_size)
+            {
+                std::cout << " >> Txn attempt commit" << std::endl;
+                if (batch_index % 2 == 0) {
```
pematous, line 108: I see that every second batch is aborted. I'm not sure if that's intended, though it works as expected (python example is not doing any aborts).

cpp/examples/tx_send.cpp
```diff
@@ -0,0 +1,175 @@
...
+                    transaction.abort();
+                }
+
+                transaction = proton::transaction();
+                batch_index++;
+            }
+        }
+    }
+
+    void on_tracker_accept(proton::tracker &t) override {
```
pematous, line 120: I believe on_tracker_accept method is extra and is never fired. If so, counting confirmed doesn't make much sense (confirmed variable is not used to anything btw).

cpp/examples/tx_recv.cpp
```diff
@@ -0,0 +1,129 @@
...
+                  << std::endl;
+        std::cout << "[on_transaction_declared] txn is_empty " << (t.is_empty())
+                  << "\t" << transaction.is_empty() << std::endl;
+        receiver.add_credit(batch_size); 
+        transaction = t;
+    }
+
+    void on_message(proton::delivery &d, proton::message &msg) override {
+        std::cout<<"# MESSAGE: " << msg.id() <<": "  << msg.body() << std::endl;
+        transaction.accept(d);
```
pematous, line 83: btw. I'm thinking if transaction.accept() does actually have any effect? I mean, when I comment it out, the program behaves the same. btw. is this transaction accept any different from delivery accept? looks both settles the messages related to the delivery.

Does explicit delivery settling play role in transaction mode? or are they mutually exclusive and only commits and aborts applies? Can ie. single message in transaction be rejected while other messages accepted? 

I tried to accept / reject the delivery and it seems to have no effect in transaction mode (while in python, doing so makes the messages to be threat outside of the transaction, just like a normal messages, https://issues.redhat.com/browse/ENTMQCL-513).

cpp/examples/tx_recv.cpp
```diff
@@ -0,0 +1,129 @@
...
+        current_batch += 1;
+        if(current_batch == batch_size) {
+            transaction = proton::transaction(); // null
```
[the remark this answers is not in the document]
pematous: > I believe we should do rather commit here, that way it works as expected and the receiver is closed after expected number of messages received (with the current implementation it's not).

above mentioned change:
```
-            transaction = proton::transaction(); // null
+           transaction.commit();
```

works as expected (against Artemis broker) when using multicast. However, when a pre-defined anycast queue is used, the commit() call ends with segmentation fault.

cpp/src/node_options.cpp
```diff
@@ -162,6 +162,7 @@ class target_options::impl {
     option<enum target::expiry_policy> expiry_policy;
     option<std::vector<symbol> > capabilities;
     option<target::dynamic_property_map> dynamic_properties;
+    option<int> type;
```
astitcher, line 165: possibly replace this with `option<bool> is_coordinator;`?

cpp/src/node_options.cpp
```diff
@@ -175,6 +176,9 @@ class target_options::impl {
             get(dynamic_properties.value, target_map);
             value(pn_terminus_properties(unwrap(t))) = target_map;
         }
+        if (type.set) {
+            pn_terminus_set_type(unwrap(t), pn_terminus_type_t(type.value));
+        }
```
astitcher, line 181: use `is_coordinator`?

cpp/src/proton_bits.hpp
```diff
@@ -98,11 +102,15 @@ template <> struct wrapped<sender> { typedef pn_link_t type; };
 template <> struct wrapped<receiver> { typedef pn_link_t type; };
 template <> struct wrapped<transfer> { typedef pn_delivery_t type; };
 template <> struct wrapped<tracker> { typedef pn_delivery_t type; };
+template <> struct wrapped<disposition> {
+    typedef pn_disposition_t type;
+};
```
astitcher, line 107: Inconsistent formatting

cpp/src/proton_bits.hpp
```diff
@@ -111,6 +119,9 @@ template <> struct wrapper<pn_connection_t> { typedef connection type; };
 template <> struct wrapper<pn_session_t> { typedef session type; };
 template <> struct wrapper<pn_link_t> { typedef link type; };
 template <> struct wrapper<pn_delivery_t> { typedef transfer type; };
+template <> struct wrapper<pn_disposition_t> {
+    typedef disposition type;
+};
```
astitcher, line 124: Inconsistent formatting

cpp/src/sender_options.cpp
```diff
@@ -66,6 +66,7 @@ class sender_options::impl {
     option<bool> auto_settle;
     option<source_options> source;
     option<target_options> target;
+    option<coordinator_options> coordinator;
```
astitcher, line 69: Not needed if coordinator is a kind of target

cpp/src/sender_options.cpp
```diff
@@ -82,6 +83,10 @@ class sender_options::impl {
                 proton::target local_t(make_wrapper<proton::target>(pn_link_target(unwrap(s))));
                 target.value.apply(local_t);
             }
+            if (coordinator.set) {
+                proton::coordinator local_t(make_wrapper<proton::coordinator>(pn_link_target(unwrap(s))));
+                coordinator.value.apply(local_t);
+            }
```
astitcher, line 89: Ditto

cpp/src/sender_options.cpp
```diff
@@ -118,6 +123,7 @@ sender_options& sender_options::delivery_mode(proton::delivery_mode m) {impl_->d
 sender_options& sender_options::auto_settle(bool b) {impl_->auto_settle = b; return *this; }
 sender_options& sender_options::source(const source_options &s) {impl_->source = s; return *this; }
 sender_options& sender_options::target(const target_options &s) {impl_->target = s; return *this; }
+sender_options& sender_options::coordinator(const coordinator_options &s) {impl_->coordinator = s; return *this; }
```
astitcher, line 126: Ditto

cpp/src/transaction.cpp
```diff
@@ -0,0 +1,44 @@
+/*
```
astitcher, line 1: Call this file transaction_handler.cpp

astitcher (changes requested):
Lopts of detailed comments here, but the main change I'd like to see is either make coordinator a subclass of target or fold it into a simple boolean option of target.

cpp/examples/tx_send.cpp
DreamPearl: Removed.

cpp/include/proton/transfer.hpp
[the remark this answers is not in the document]
DreamPearl: Transfer is not holding transaction now. Removed.

cpp/src/contexts.hpp
```diff
@@ -161,6 +163,7 @@ class transfer_context : public context {
     transfer_context() : user_data_(nullptr) {}
     static transfer_context& get(pn_delivery_t* s);
 
+    std::unique_ptr<transaction> transaction_;
```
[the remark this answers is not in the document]
DreamPearl: Fixed.

cpp/src/transfer.cpp
```diff
@@ -50,6 +52,16 @@ enum transfer::state transfer::state() const { return static_cast<enum state>(pn
...
+void transfer::transaction(proton::transaction t) {
+    transfer_context &cc = transfer_context::get(pn_object());
+    cc.transaction_ = std::make_unique<proton::transaction>(t);
+}
+
+transaction transfer::transaction() const {
+    transfer_context& cc = transfer_context::get(pn_object());
+    return *cc.transaction_;
+}
+
```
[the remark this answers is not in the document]
DreamPearl: Removed. Transfer no longer holds the transaction now.

cpp/include/proton/session.hpp
```diff
@@ -29,13 +29,15 @@
...
 #include <string>
+#include <iostream>
 
 /// @file
 /// @copybrief proton::session
 
 struct pn_session_t;
 
 namespace proton {
+  class transaction_impl;
```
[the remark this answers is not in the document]
DreamPearl: We need this here; otherwise throwing an error `transaction_impl’ does not name a type`

cpp/include/proton/sender_options.hpp
```diff
@@ -94,6 +94,9 @@ class sender_options {
     /// Options for the receiver node of the receiver.
     PN_CPP_EXTERN sender_options& target(const target_options&);
 
+    /// Options for the coordinator node of the receiver.
+    PN_CPP_EXTERN sender_options& coordinator(const coordinator_options&);
```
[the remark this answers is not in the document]
DreamPearl: Done

cpp/include/proton/terminus.hpp
```diff
@@ -122,6 +122,7 @@ class terminus {
   friend class internal::factory<terminus>;
   friend class source;
   friend class target;
+  friend class coordinator;
```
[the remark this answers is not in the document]
DreamPearl: Removed. No longer needed.

cpp/src/node_options.cpp
DreamPearl: Done

cpp/src/node_options.cpp
DreamPearl: Done

cpp/include/proton/container.hpp
```diff
@@ -326,6 +326,7 @@ class PN_CPP_CLASS_EXTERN container {
   friend class receiver_options;
   friend class sender_options;
   friend class work_queue;
+  friend class transaction;
```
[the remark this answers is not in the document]
DreamPearl: Removed. No longer needed.

cpp/include/proton/fwd.hpp
```diff
@@ -52,9 +52,12 @@ class sender_options;
 class session;
 class session_options;
 class source_options;
+class coordinator_options;
 class ssl;
 class target_options;
 class tracker;
+class transaction;
```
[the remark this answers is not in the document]
DreamPearl: Removed.

cpp/include/proton/sender_options.hpp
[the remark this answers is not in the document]
DreamPearl: Removed.

cpp/src/proton_bits.hpp
DreamPearl: Not needed now. Removed.

cpp/src/proton_bits.hpp
DreamPearl: Same. Not needed, removed!

cpp/examples/tx_send.cpp
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    void on_sendable(proton::sender &s) override {
+        std::cout << "    [OnSendable] session: " << &session
+                  << std::endl;
+        send(s);
+    }
+
+    void send(proton::sender &s) {
+        static int unique_id = 10000;
```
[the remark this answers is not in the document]
DreamPearl: Done!

cpp/src/messaging_adapter.cpp
```diff
@@ -116,7 +119,29 @@ void on_delivery(messaging_handler& handler, pn_event_t* event) {
...
+                handler.on_tracker_release(t);
+                break;
+            }
+            if (t.settled()) {
+                handler.on_tracker_settle(t);
+                if (lctx.auto_settle)
+                    t.settle();
+            }
+        }
+    } else if (pn_link_is_receiver(lnk)) {
```
astitcher, line 144: I don't understand why the handling of messages on a transaction coordinator link should be different from any other link.

cpp/examples/tx_send.cpp
```diff
@@ -0,0 +1,164 @@
...
+        while (session.txn_is_declared() && sender.credit() &&
+               (committed + current_batch) < total) {
+            proton::message msg;
+            std::map<std::string, int> m;
+            m["sequence"] = committed + current_batch;
+
+            msg.id(std::atomic_fetch_add(&unique_id, 1));
+            msg.body(m);
+            std::cout << "Sending: " << msg << std::endl;
+            session.txn_send(sender, msg);
```
astitcher, line 97: I think the API we were aiming for would just use a sender object created from the transactioned session.

cpp/src/session.cpp
```diff
@@ -148,4 +158,231 @@ void* session::user_data() const {
...
+    proton::tracker tracker = s.send(msg);
+    update(tracker, 0x34);
+    return tracker;
+}
+
+void transaction_impl::accept(delivery &t) {
+    t.settle();
+}
+
+void transaction_impl::update(tracker &t, uint64_t state) {
```
astitcher, line 314: I think this should be delivery not tracker

cpp/examples/tx_send.cpp
DreamPearl: Done!

astitcher (commented):
First batch of comments

cpp/include/proton/session.hpp
```diff
@@ -105,14 +106,31 @@ PN_CPP_CLASS_EXTERN session : public internal::object<pn_session_t>, public endp
...
+
+    PN_CPP_EXTERN bool transaction_is_empty();
+    PN_CPP_EXTERN bool transaction_is_declared();
+    PN_CPP_EXTERN void transaction_commit();
+    PN_CPP_EXTERN void transaction_abort();
+    PN_CPP_EXTERN void transaction_declare();
+    PN_CPP_EXTERN void transaction_handle_outcome(proton::tracker);
+    PN_CPP_EXTERN void attach_txn_id(proton::tracker t);
+    PN_CPP_EXTERN void transaction_accept(delivery &t);
+    PN_CPP_EXTERN proton::connection transaction_connection() const;
```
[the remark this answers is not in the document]
DreamPearl: Is making them private enough  to make them internal.. or do you mean something else? I also marked them as "/// @cond INTERNAL". Will it help?

cpp/include/proton/transaction.hpp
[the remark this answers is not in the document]
DreamPearl: Updated. Now they are session callbacks, transaction_declared() becomes handler->on_session.open() and on_declare_failed is now handled by on_session_error().

cpp/include/proton/session.hpp
```diff
@@ -105,14 +105,22 @@ PN_CPP_CLASS_EXTERN session : public internal::object<pn_session_t>, public endp
     /// Get user data from this session.
     PN_CPP_EXTERN void* user_data() const;
 
+    PN_CPP_EXTERN void transaction_declare(proton::messaging_handler &handler, bool settle_before_discharge = false);
+    PN_CPP_EXTERN bool transaction_is_declared();
+    PN_CPP_EXTERN proton::binary transaction_id() const;
+    PN_CPP_EXTERN void transaction_commit();
+    PN_CPP_EXTERN void transaction_abort();
```
astitcher: Need documentation for these APIs

cpp/include/proton/target_options.hpp
```diff
@@ -60,6 +60,10 @@ class target_options {
     /// address is ignored if dynamic() is true.
     PN_CPP_EXTERN target_options& address(const std::string& addr);
 
+    /// Set the target be of type coordinator.
+    /// This immediately override the currently assigned type.
+    PN_CPP_EXTERN target_options& make_coordinator();
+
```
astitcher: I don't think this is needed to use the transaction API. If it's used in the implementation it needs to be purely internal. But I don't think it's actually used at all.

cpp/src/messaging_adapter.cpp
```diff
@@ -274,13 +277,6 @@ void on_link_local_open(messaging_handler& handler, pn_event_t* event) {
 
 void on_link_remote_open(messaging_handler& handler, pn_event_t* event) {
     auto lnk = pn_event_link(event);
-    // Currently don't implement (transaction) coordinator
-    if (pn_terminus_get_type(pn_link_remote_target(lnk))==PN_COORDINATOR) {
-      auto error = pn_link_condition(lnk);
-      pn_condition_set_name(error, "amqp:not-implemented");
-      pn_link_close(lnk);
-      return;
-    }
```
astitcher: I think this deletion isn't exactly correct: Whilst we do now need to cope with a remote coordinator this is only on the link the implementation itself has opened to control transactions not any other general link so we should still not allow incoming coordinator links in general.

cpp/src/sender.cpp
```diff
@@ -26,10 +26,12 @@
 #include "proton/source.hpp"
 #include "proton/target.hpp"
 #include "proton/tracker.hpp"
+#include "types_internal.hpp"
```
astitcher: This include is in the wrong block of includes - it is purely internal so should in the bottom block

cpp/src/session.cpp
```diff
@@ -148,4 +158,209 @@ void* session::user_data() const {
...
+        proton::target_options opts;
+        std::vector<symbol> cap = {proton::symbol("amqp:local-transactions")};
+        opts.capabilities(cap);
+        opts.make_coordinator();
+
+        proton::sender_options so;
+        so.name("txn-ctrl");
+        so.target(opts);
+
+        static InternalTransactionHandler internal_handler; // internal_handler going out of scope. Fix it
```
astitcher: This *Cannot* be a static, it must be part of the txn_impl somehow.

cpp/src/session.cpp
```diff
@@ -148,4 +158,209 @@ void* session::user_data() const {
...
+      DISCHARGING,
+    };
+    enum State state = State::FREE;
+    std::vector<proton::tracker> pending;
+
+    void commit();
+    void abort();
+    void declare();
+
+    void discharge(bool failed);
```
astitcher: For example I'm pretty sure the `discharge()` is only used internal to here so should be private.

cpp/src/session.cpp
```diff
@@ -148,4 +158,209 @@ void* session::user_data() const {
...
+    discharge(false);
+}
+
+void transaction_impl::abort() {
+    discharge(true);
+}
+
+void transaction_impl::declare() {
+    if (state != transaction_impl::State::FREE)
+        throw proton::error("This session has some associcated transaction already");
```
astitcher: Spelling error "associated". Grammar error "This session has an associated...". But change message to something briefer: "Session already has transaction" for example.

cpp/src/session.cpp
```diff
@@ -148,4 +158,209 @@ void* session::user_data() const {
...
+    state = State::DECLARING;
+
+    proton::symbol descriptor("amqp:declare:list");
+    std::list<proton::value> vd;
+    proton::value i_am_null;
+    vd.push_back(i_am_null);
+    proton::value _value = vd;
+    send_ctrl(descriptor, _value);
```
astitcher: Can't this be something shorter using inline construction? In any case get rid of `_` leading `_value`

cpp/src/session.cpp
```diff
@@ -148,4 +158,209 @@ void* session::user_data() const {
...
+    failed = _failed;
+    proton::symbol descriptor("amqp:discharge:list");
+    std::list<proton::value> vd;
+    vd.push_back(transaction_id);
+    vd.push_back(failed);
+    proton::value _value = vd;
+    send_ctrl(descriptor, _value);
+}
+
+proton::tracker transaction_impl::send_ctrl(proton::symbol descriptor, proton::value _value) {
```
astitcher: Replace `_value`

cpp/src/session.cpp
```diff
@@ -148,4 +158,209 @@ void* session::user_data() const {
...
+    proton::binary transaction_id;
+    bool failed = false;
+    enum State {
+      FREE,
+      DECLARING,
+      DECLARED,
+      DISCHARGING,
+    };
+    enum State state = State::FREE;
+    std::vector<proton::tracker> pending;
```
astitcher: Delete - pending is never used

cpp/src/session.cpp
```diff
@@ -148,4 +158,209 @@ void* session::user_data() const {
...
+            return;
+        }
+    } else if (state == State::DISCHARGING) {
+        // Attempting to commit/abort transaction
+        if (pn_disposition_is_failed(disposition)) {
+            if (!failed) {
+                state = State::FREE;
+                transaction_delete(session);
+                handler->on_session_transaction_commit_failed(session);
+                release_pending();
```
astitcher: Delete - pending is never used

cpp/src/session.cpp
```diff
@@ -148,4 +158,209 @@ void* session::user_data() const {
...
+                return;
+            } else {
+                // Transaction commit is successful
+                state = State::FREE;
+                transaction_delete(session);
+                handler->on_session_transaction_committed(session);
+                return;
+            }
+        }
+        pending.clear();
```
astitcher: Delete - pending is never used

cpp/src/session.cpp
```diff
@@ -148,4 +158,209 @@ void* session::user_data() const {
...
+    for (auto d : pending) {
+        delivery d2(make_wrapper<delivery>(unwrap(d)));
+        d2.release();
+    }
+    pending.clear();
+}
+
+void transaction_impl::handle_outcome(proton::tracker t) {
+    pn_disposition_t *disposition = pn_delivery_remote(unwrap(t));
+    pn_declared_disposition_t *declared_disp = pn_declared_disposition(disposition);
```
astitcher: The tests in here are strange and maybe wrong: `pn_declared_disposition` will return `nullptr` if the disposition is not a declared disposition, for `pn_disposition_is_failed` to ever return true the disposition must be a modified disposition. So the tests maybe should be reordered?

cpp/src/session.cpp
```diff
@@ -148,4 +158,209 @@ void* session::user_data() const {
...
+    proton::session session = t.session();
+    if (state == State::DECLARING) {
+        // Attempting to declare transaction
+        if (pn_disposition_is_failed(disposition)) {
+            state = State::FREE;
+            transaction_delete(session);
+            // on_transaction_declare_failed
+            handler->on_session_error(session);
```
astitcher: Use after free: handler is used after the transaction_impl is deleted! Also no need to set the state as the impl is just about to be deleted anyway.

astitcher (changes requested):
I've made a bunch of suggested changes to this branch which address some of the comments here

astitcher:
My suggested changes are in branch https://github.com/astitcher/qpid-proton/commits/local-transactions
The last commit in that branch is the relevant commit.
I will carry on working on that tomorrow to see how far I can progress it before the weekend.

cpp/src/contexts.hpp
```diff
@@ -25,6 +25,7 @@
 #include "reconnect_options_impl.hpp"
 
 #include "proton/work_queue.hpp"
+#include "proton/session.hpp"
```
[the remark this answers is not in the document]
DreamPearl: Removing this header breaking transitive includes.

Fixed. By adding `#include <proton/condition.h>`  here  and including `session.hpp`directly in delivery.cpp.

astitcher:
Closed in favour of #442

PR #438, opened 2025-01-29
PROTON-2531: Use surrogateescape when reading delivery tag

pjfawcett:
Delivery tags are bytes 'on the wire' but strings in the Python interface.
Use Latin-1 codec when converting, so that all valid bytes values (0 - 255) are accepted and converted without changing the values.

### DIFF
--- a/python/cproton.py
+++ b/python/cproton.py
@@ -200,16 +200,16 @@ def bytes2pybytes(b):
     return bytes(ffi.buffer(b.start, b.size))
 
 
-def bytes2string(b, encoding='utf8'):
-    return ffi.unpack(b.start, b.size).decode(encoding)
+def bytes2string(b, encoding='utf8', errors='strict'):
+    return ffi.unpack(b.start, b.size).decode(encoding, errors)
 
 
-def py2bytes(py, encoding='utf8'):
+def py2bytes(py):
     if isinstance(py, (bytes, bytearray, memoryview)):
         s = ffi.from_buffer(py)
         return len(s), s
     elif isinstance(py, str):
-        s = ffi.from_buffer(py.encode(encoding))
+        s = ffi.from_buffer(py.encode('utf8'))
         return len(s), s
 
 
@@ -487,7 +487,7 @@ def pn_data_get_symbol(data):
 
 
 def pn_delivery_tag(delivery):
-    return bytes2string(lib.pn_delivery_tag(delivery), 'latin-1')
+    return bytes2string(lib.pn_delivery_tag(delivery), errors='surrogateescape')
 
 
 def pn_connection_get_container(connection):
@@ -543,7 +543,7 @@ def pn_receiver(session, name):
 
 
 def pn_delivery(link, tag):
-    return lib.pn_delivery(link, py2bytes(tag, 'latin-1'))
+    return lib.pn_delivery(link, py2bytes(tag))
 
 
 def pn_link_name(link):
--- a/python/cproton.py
+++ b/python/cproton.py
@@ -204,12 +204,12 @@ def bytes2string(b, encoding='utf8'):
     return ffi.unpack(b.start, b.size).decode(encoding)
 
 
-def py2bytes(py):
+def py2bytes(py, encoding='utf8'):
     if isinstance(py, (bytes, bytearray, memoryview)):
         s = ffi.from_buffer(py)
         return len(s), s
     elif isinstance(py, str):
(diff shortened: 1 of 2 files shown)


python/cproton.py
```diff
@@ -204,12 +204,12 @@ def bytes2string(b, encoding='utf8'):
...
 
-def py2bytes(py):
+def py2bytes(py, encoding='utf8'):
     if isinstance(py, (bytes, bytearray, memoryview)):
         s = ffi.from_buffer(py)
         return len(s), s
     elif isinstance(py, str):
-        s = ffi.from_buffer(py.encode('utf8'))
+        s = ffi.from_buffer(py.encode(encoding))
         return len(s), s
```
astitcher, line 213: Not entirely sure about this change - as it's only necessary because of the change below for round tripping.

python/cproton.py
```diff
@@ -543,7 +543,7 @@ def pn_receiver(session, name):
 
 
 def pn_delivery(link, tag):
-    return lib.pn_delivery(link, py2bytes(tag))
+    return lib.pn_delivery(link, py2bytes(tag, 'latin-1'))
```
astitcher, line 546: So this change makes sense from the point of view of round tripping the tags. But it is a change of behavior which will raise an exception for some input strings which would previously have been accepted (but with a different encoding).

astitcher (commented):
I think this basically looks good.
However I'm now in two minds about what is the correct direction to fix this problem, because it's apparent now that every consistent round trip fix changes the behavior for some current case:
1. If we use this change, then if tags are specified using a string in python some unicode codepoints which were previously allowed will cause an exception.
2. If we change the binding to produce bytes from tags instead of strings then we've changed the python type of received tags. Of course this is the behavior that we should have been using and probably won't cause problems for many real life programs. It is a larger change of course because we need to change the existing tests.
So I was considering a change that doesn't round trip properly, but doesn't raise an error where there currently wouldn't be an error. This doesn't really seem better though, so I'm a bit torn between the options here.
I think if I could be sure that making the change to produce bytes wouldn't affect very many users of the binding then I'd prefer that. But that's not at all obvious to me!
In conclusion, this approach is probably the best one, although I think I would also consider using the raw-unicode-escape codec for the python string to tag bytes direction, whilst still keeping the latin1 codec for the bytes tag to string direction. This should correctly round trip any string/binary tag that this change accepts, whilst not raising an exception for any valid unicode if it is supplied as the string for a tag. If you do use a unicode codepoint greater than 0xff then this will not roundtrip but the current change doesn't accept it at all. For clarification I do want to carry on using latin1 in the direction from the wire as that does seem to be correct and special casing the unicode escapes in that direction seems wrong.

pjfawcett:
I share your indecision.  The current usage of this binding by user code is unknown so we can't be certain what a single "right" solution would be.  I put in the 'expected exception' in the unit test to highlight the issue with encoding with Latin-1.

I decided to go back and dig into the behaviour of version 0.38.0, which uses SWIG instead of CFFI.

After picking through the code, including the SWIG generated C wrapper code, I ascertained that:

1. When setting the tag bytes from a Python string, [PyUnicode_AsUTF8AndSize](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_AsUTF8AndSize) is called, which has the same behaviour as `encode("utf-8", errors="strict")`
2. When reading the tag bytes into a Python string, [PyUnicode_DecodeUTF8](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_DecodeUTF8) is called with `errors="surrogateescape"`, which has the same behaviour as `decode("utf-8", errors="surrogateescape")`

This combination allows for round tripping of Unicode strings when going string -> tag -> string and also accepts non-utf8 byte sequences in incoming tags.  (Although the resultant strings from such tags would cause exceptions if used as a new tag, since the  
 encoding "fails if the string contains surrogate code points")

So, in order to maintain backwards compatibility, and to fix the problem I encountered with some tags causing exceptions when read, I propose to implement the encoding/decoding using the parameters outlined above.

Having said that, it is annoying that one can't easily get to the raw byte values for the tags.  The current, 0.40.0, implementation allows the setting of the tag form a `bytes` value as so bypasses codecs.  I think it would be good to have more direct access to the incoming tag bytes.

Could I suggest that a `btag` property is added to the `Delivery` class, alongside the existing `tag` property, to get the tag a `bytes`.

I'm happy to do this, and to raise a separate ticket / PR if required.

astitcher:
Committed this with small changes.

PR #439, opened 2025-02-20
Fix Qpid::Proton::DEFAULT_URI_PARSER for Ruby 3.4

eduresende:
On Ruby 3.4 the DEFAULT_URI_PARSER is throwing an error:

```ruby
❯ DEFAULT_URI_PARSER = URI::Parser.new(:HOSTNAME => /(?:#{URI::PATTERN::HOSTNAME})|/)
'Class#new': wrong number of arguments (given 1, expected 0) (ArgumentError)
        from (irb):438:in '<main>'
```

This PR aims to replace the parser with a custom version that works well with Ruby 3.4

### DIFF
--- a/ruby/lib/core/uri.rb
+++ b/ruby/lib/core/uri.rb
@@ -46,7 +46,12 @@ end
 module Qpid::Proton
   private
   # Make sure to allow empty hostnames, Ruby 2.0.0 does not.
-  DEFAULT_URI_PARSER = URI::RFC2396_Parser.new(:HOSTNAME => /(?:#{URI::PATTERN::HOSTNAME})|/)
+  DEFAULT_URI_PARSER =
+    if RUBY_VERSION >= '3.4'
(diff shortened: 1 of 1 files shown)


astitcher:
@eduresende Is there an initial version of rubby that this change works for? In other words how backwards compatible is this change?

eduresende:
> @eduresende Is there an initial version of rubby that this change works for? In other words how backwards compatible is this change?

@astitcher Thank you for the answer. `URI::RFC2396_Parser` has existed since Ruby 2.2, but in Ruby 3.4 the default parser changed: https://www.ruby-lang.org/en/news/2024/12/25/ruby-3-4-0-released/

2025-02-26 at 14 00

as the problem only happens after Ruby 3.4, I changed the PR to apply the fix only under this condition. This way, we won't have any risks of incompatibility with previous Ruby versions.

What do you think?

astitcher:
Yeah that seems like the right approach.

astitcher:
Merged

PR #442, opened 2025-12-05
C++ Local transactions

astitcher:
This is the work in #437 just about ready to be merged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # General
 
 /build*/
+/build*
 /bld*/
 /BLD*/
 /out
--- a/cpp/examples/CMakeLists.txt
+++ b/cpp/examples/CMakeLists.txt
@@ -60,7 +60,9 @@ foreach(example
     scheduled_send
     service_bus
     multithreaded_client
-    multithreaded_client_flow_control)
+    multithreaded_client_flow_control
+    tx_send
+    tx_recv)
   add_executable(${example} ${example}.cpp)
   target_link_libraries(${example} Proton::cpp Threads::Threads)
 endforeach()
--- a/cpp/examples/CMakeLists.txt
+++ b/cpp/examples/CMakeLists.txt
@@ -61,6 +61,7 @@ foreach(example
     service_bus
     multithreaded_client
     multithreaded_client_flow_control
+    tx_interactive
     tx_send
     tx_recv)
   add_executable(${example} ${example}.cpp)
--- a/cpp/examples/options.hpp
+++ b/cpp/examples/options.hpp
@@ -53,6 +53,11 @@ class options {
         opts_.push_back(new option_value<T>(value, short_name, long_name, description, var));
     }
 
+    /** Overload for std::string: takes the full argument (including spaces). */
+    void add_value(std::string& value, char short_name, const std::string& long_name, const std::string& description, const std::string var) {
+        opts_.push_back(new option_string(value, short_name, long_name, description, var));
+    }
+
     /** Sets flag when parse() is called if option is present. */
     void add_flag(bool& flag, char short_name, const std::string& long_name, const std::string& description) {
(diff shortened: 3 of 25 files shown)


astitcher:
I've done a lot more work on this now, and I'm getting a lot happier about the correctness of its behaviour for outgoing messsages. I think the settlement behaviour for these is now correct under successful commit and also rollback.

I've enhanced the python broker some more and I think it now actually represents most of the transactional behaviour of a spec correct broker. It would be good to add a way to make it fail to commit - perhaps adding a transaction timeout timer would be a realistic way to do this.

astitcher:
I think this work is now good enough for a tech preview style release. There are still areas that I would like to see some further work though, and I will raise extra issues for these:
* There should be better testing for the various transaction scenarios:
   - Ideally this would include an in process broker for ease of use
* It would be more convenient if we could configure the settlement and release behaviour for messages:
  -  It would probably be more convenient for most users to automatically release any messages acquired under a transaction if the transaction is aborted or the commit fails.
  - There are some (probably rare) cases where we might want to automatically settle messages sent or acknowledged under a transaction
  - Probably the way to do this would be to add a transaction_option object to the the session - this is consistent with the current API and would be extensible.

gemmellr:
Some feedback after looking over the changes.

Examples:
- If explicitly releasing consumed messages after rollback, using modified(delivery-failed=true) would be the more typical than 'released' like the example does currently, in order to also increment delivery count.
- That said, not needing to explicitly release/other the rolled back messages at all would be nicer and simpler, see settlement discussion below.
- There is no example of doing both receiving and sending, which is one of the main transaction use cases for reliable processing of request and sending response.


The use of settlement in the client:
- For the non-tx case, when consuming messages the client also settles when accepting/other (either by application accepting/other, or by 'auto-accept' occurring). The application doesnt need to worry about settlement at all.
- For the tx case, the transactional acceptance is being done unsettled, which means having to deal with releasing[/modifying(delivery-failed=true)] the messages after a rollback, and [auto?] settling after commit.
- Alternatively, sending settled transactional acceptance/other on a link with default-outcome of modified(delivery-failed=true) would mean the transactional application similarly wouldn't need to deal with settlement after discharge, or explicitly releasing (if using batch-sized credit grants, as the example does). The transaction commit would either succeed, or fail if the prior acceptances couldnt be applied and the deliveries would disappear with delivery count bumped and the messages be resent. Similarly with explicit rollback. No need for applications or client to handle releasing or settlement as it would already be done. Less bandwidth used. This settled-transactional-acceptance is what the other clients do. Matching them would seem nice, its certainly the exercised path for brokers also.
- https://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transactions-v1.0-os.html#doc-idp145616

Tests:
No actual-tests in tree. Seems likely to cause issues.

API:
The API essentially bakes-in lower peak tx throughput. I dont really consider this an issue needing changed, just noting it as something to consider now before the API is in the wild, in case it wasnt considered already. I think it actually makes sense as is for the simplicity.

The separate explicit declare and discharge APIs, and subsequent need to wait for discharge completion before declaring again, means the use of transactions currently mandates waiting 2 full round trips in order to begin a subsequent transaction, which could up to half the peak message rate possible for cases with prefetched messages.
This is something users ran into in qpid-jms about 10 years ago. For transacted sessions JMS only has 'commit' and 'rollback', i.e there is always a transaction from the API perspective, so we switched to pipelining the discharges and declares to effectively remove a round trip and ~double peak throughput, returning to levels like what [our] other JMS clients could do. Of course, commit and rollback are synchronous APIs in JMS which does make that easier to do than here (but to be clear, it was still a massive pain to do).

Notably though, ProtonJ2 made the same choice of separate declare and discharge operations, and forcing the 2 round trips, despite the operations being synchronous and despite the prior knowledge from qpid-jms, because it is far far simpler and transactional use cases are seemingly less typical with the non-JMS clients.

PR #443, opened 2026-04-17
Epoll proactor fixes.  AMQP connection hang and busy loop.

cliffjansen:
Both exposed by same reproducer scenario.


astitcher:
LGTM

PR #444, opened 2026-05-01
PROTON-2931: epoll proactor thread races using async c-ares name reso…

cliffjansen:
fixes router_locust zombie connections

### DIFF
--- a/c/src/proactor/epoll.c
+++ b/c/src/proactor/epoll.c
@@ -836,7 +836,7 @@ static const char *pconnection_setup(pconnection_t *pc, pn_proactor_t *p, pn_con
 // Call with lock held and closing == true (i.e. pn_connection_driver_finished() == true), no pending timer.
 // Return true when all possible outstanding epoll events associated with this pconnection have been processed.
 static inline bool pconnection_is_final(pconnection_t *pc) {
-  return !pc->current_arm && !pc->task.ready && !pc->tick_pending;
+  return !pc->current_arm && !pc->task.ready && !pc->tick_pending && !pc->name_lookup_pending;
 }
 
 static void pconnection_final_free(pconnection_t *pc) {
@@ -1400,6 +1400,7 @@ static void pconnection_start(pconnection_t *pc, int fd) {
 }
 
 /* Called on initial connect, and if connection fails to try another address */
+/* May be called within the pconnection task or from an external name_lookup task */
 static void pconnection_maybe_connect_lh(pconnection_t *pc) {
   errno = 0;
   if (!pc->connected) {         /* Not yet connected */
@@ -1447,7 +1448,11 @@ bool schedule_if_inactive(pn_proactor_t *p) {
 static void connection_lookup_done_lh(pconnection_t *pc, struct addrinfo *ai, int gai_error) {
   pn_proactor_t *p = pc->task.proactor;
   bool notify = false;
-  if (gai_error) {
+
+  if (pconnection_rclosed(pc) && pconnection_wclosed(pc)) {
+    // Closed by application. Skip pending connect(). gai_error no longer of interest.
+    freeaddrinfo(ai);
+  } else if (gai_error) {
     psocket_gai_error(&pc->psocket, gai_error, "connect to ");
   } else if (ai) {
     pc->addrinfo = ai;
@@ -1464,6 +1469,7 @@ static void connection_lookup_done_lh(pconnection_t *pc, struct addrinfo *ai, in
 static void connection_done_cb(void *user_data, struct addrinfo *ai, int gai_error) {
   pconnection_t *pc = (pconnection_t *)user_data;
   lock(&pc->task.mutex);
+  pc->name_lookup_pending = false;
   connection_lookup_done_lh(pc, ai, gai_error);
   unlock(&pc->task.mutex);
 }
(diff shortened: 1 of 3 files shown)


c/src/proactor/epoll.c
```diff
@@ -1464,6 +1469,7 @@ static void connection_lookup_done_lh(pconnection_t *pc, struct addrinfo *ai, in
 static void connection_done_cb(void *user_data, struct addrinfo *ai, int gai_error) {
   pconnection_t *pc = (pconnection_t *)user_data;
   lock(&pc->task.mutex);
+  pc->name_lookup_pending = false;
```
astitcher: If this is within the lock, maybe it should be in the _lh function not here? Especially as setting the flag is done in a _lh function.

c/src/proactor/epoll.c
```diff
@@ -1447,7 +1448,11 @@ bool schedule_if_inactive(pn_proactor_t *p) {
 static void connection_lookup_done_lh(pconnection_t *pc, struct addrinfo *ai, int gai_error) {
   pn_proactor_t *p = pc->task.proactor;
   bool notify = false;
-  if (gai_error) {
+
+  if (pconnection_rclosed(pc) && pconnection_wclosed(pc)) {
```
astitcher: Is it possible to only have one of these set in this condition?

c/src/proactor/epoll.c
```diff
@@ -1472,10 +1478,28 @@ static void connection_done_cb(void *user_data, struct addrinfo *ai, int gai_err
...
+      psocket_error(&pc->psocket, EAI_FAIL, "internal error on connect");
+    }
+    return false;
+  }
+  if (!pc->name_lookup_pending) {
+    // connection_done_cb already completed
+    if (pn_condition_is_set(pn_transport_condition(tp)))
+      return false;
+  }
+  return !pc->queued_disconnect && !pni_task_wake_pending(&pc->task);
```
astitcher: This is pretty obscure compared to the previous return value (which would essentially be true at his point I think if there wer no errors from starting the lookup), perhaps an explanation in a comment is in order.

c/src/proactor/epoll_raw_connection.c
```diff
@@ -110,8 +111,14 @@ static void praw_connection_start(praw_connection_t *prc, int fd) {
 }
 
 /* Called on initial connect, and if connection fails to try another address */
+/* May be called within the praw_connection task or from an external name_lookup task */
 static void praw_connection_maybe_connect_lh(praw_connection_t *prc) {
+  int err = 0;
```
astitcher: Move this to just before the while loop where it belongs, also no need to set it to 0 here - it is initialised first thing in the loop.

c/src/proactor/epoll_raw_connection.c
```diff
@@ -161,6 +173,7 @@ static void raw_connection_lookup_done_lh(praw_connection_t *prc, struct addrinf
 static void raw_connection_done_cb(void *user_data, struct addrinfo *ai, int gai_error) {
   praw_connection_t *prc = (praw_connection_t *)user_data;
   lock(&prc->task.mutex);
+  prc->name_lookup_pending = false;
```
astitcher: As above shouldn't this be inside the _lh function? Even though there is no _lh function anymore for lookup start here.

c/src/proactor/epoll_raw_connection.c
```diff
@@ -211,6 +224,9 @@ static void praw_initiate_cleanup(praw_connection_t *prc) {
     shutdown(prc->psocket.epoll_io.fd, SHUT_RDWR);
     return;
   }
+  if (prc->name_lookup_pending) {
```
astitcher: What about the protecting lock?

c/src/proactor/epoll_raw_connection.c
```diff
@@ -224,21 +240,30 @@ pn_raw_connection_t *pn_raw_connection(void) {
...
-// Call from pconnection_process with task lock held.
-// Return true if the socket is connecting and there are no Proton events to deliver.
-static bool praw_connection_first_connect_lh(praw_connection_t *prc) {
+// Call from pconnection_process with no locks.
+// Callback may complete before pni_name_lookup_start returns.
+static void praw_connection_first_connect(praw_connection_t *prc) {
   const char *host;
   const char *port;
   pn_proactor_t *p = prc->task.proactor;
+  bool notify = false;
```
astitcher: Move this down to where it's used - I think inside the if (!rc) block

c/src/proactor/epoll_raw_connection.c
```diff
@@ -224,21 +240,30 @@ pn_raw_connection_t *pn_raw_connection(void) {
...
 }
 
-// Call from pconnection_process with task lock held.
-// Return true if the socket is connecting and there are no Proton events to deliver.
-static bool praw_connection_first_connect_lh(praw_connection_t *prc) {
+// Call from pconnection_process with no locks.
+// Callback may complete before pni_name_lookup_start returns.
+static void praw_connection_first_connect(praw_connection_t *prc) {
   const char *host;
   const char *port;
```
astitcher: Actually these lines should also be just before they are used in (new) line 253

c/src/proactor/epoll_raw_connection.c
```diff
@@ -443,17 +468,30 @@ pn_event_batch_t *pni_raw_connection_process(task_t *t, uint32_t io_events, bool
...
+    bool wake_event = pni_task_wake_pending(&rc->task);
+
+    t->working = false;
+    rc->name_lookup_pending = true;
+    unlock(&rc->task.mutex);
+    praw_connection_first_connect(rc);
+    if (wake_event) {
+      lock(&rc->task.mutex);
+      t->working = true;
+      return &rc->batch;
```
astitcher, line 491: It looks wrong to return without unlocking here - if it actually correct then a comment explaining why is needed.

astitcher (changes requested):
Some small , but important issues to address, and some cosmetic, but imo significant issues to address.

c/src/proactor/epoll.c
cliffjansen: done

c/src/proactor/epoll.c
cliffjansen: seemed like a sensible optimization, yet causes hangs without other intrusive code changes.  Retracted.

c/src/proactor/epoll.c
cliffjansen: agreed.  reverted

c/src/proactor/epoll_raw_connection.c
cliffjansen: done but first init to 0 retained with added comment

c/src/proactor/epoll_raw_connection.c
cliffjansen: done

c/src/proactor/epoll_raw_connection.c
cliffjansen: doh!

c/src/proactor/epoll_raw_connection.c
cliffjansen: done

c/src/proactor/epoll_raw_connection.c
cliffjansen: done

c/src/proactor/epoll_raw_connection.c
cliffjansen: triple doh!

astitcher:
Merged

PR #446, opened 2026-05-22
NO JIRA: avoid logging AMQP credentials in python binding

kserin:
When connecting to an AMQP broker using a URI that contains credentials (e.g. amqps://user:password@host:5671), the _Connector._connect method [logged the full Url object via %r](https://github.com/apache/qpid-proton/blob/13275ca2710d4f659bda540e8358f22b0a10be24/python/proton/_reactor.py#L1042) which includes username:password in the output.

I replaced the log to only log the scheme, host and port.

### DIFF
--- a/python/proton/_url.py
+++ b/python/proton/_url.py
@@ -229,8 +229,7 @@ class Url:
         if self.username:
             userpart += quote(self.username)
         if self.password:
-            userpart += ':'
-            userpart += quote(self.password)
+            userpart += ':***'
         if self.username or self.password:
             userpart += '@'
         return userpart + hostport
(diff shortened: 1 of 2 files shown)


astitcher:
One issue with this fix is that it just omits printing the auth details as if they weren't even there - for proper debugability of authentication issues you really should be able to see that they were there but mask them from logging.
Is the username actually sensitive info that needs to be masked? If so, maybe only giving a few characters, but not all of them, might be better for being able to correlate logged authentication attempts.
I also wonder if the correct place for this is actually in the `__repr__` dunder for the Url object instead of here.

kserin:
Thanks for the feedback. I updated the PR.

I've moved the fix into `Url._netloc()` (and by extension `__str__`/`__repr__`) rather than at the log call site, as you suggested. This way any code that converts a `Url ` to a string automatically gets the masked version.

For the username, I agree it's not really sensitive information, so it's left in clear and this also helps correlate authentication attempts in logs as you mentioned.

astitcher:
@kserin the unit tests now fail - the stringified urls now don't match.

astitcher:
Merged

PR #447, opened 2026-06-17
Add OpenSSL 4.x compat

thomasgoirand:
Hi. I maintain the qpid-proton package in Debian.
This patch fixed this Debian bug for me:
https://bugs.debian.org/1138394

while retaining compatibility with older OpenSSL 3 (I tried building with both, and it built in both cases).

I hope this helps.


astitcher:
Thanks for this - merged

